PR #16, opened 2023-10-22
Feat add docker compose pg

eric-intuitem:
provide config and script to use docker-compose to launch Ciso Assistant using docker-compose, based on a Postgres database.


nas-tabchiche (approved):
Looks good to me

PR #25, opened 2024-01-06
Added Arabic flag & Language to utils

taibah:
modified:   core/utils.py


eric-intuitem:
This is obsolete with the major refactoring based on Svelte. Arabic language will be added very soon in the new version.

PR #40, opened 2024-02-12
Ca 141 prevent deleting a library that is in use in an assessment

nas-tabchiche:


eric-intuitem (approved):
with fixed regression on deletion, LGTM

PR #44, opened 2024-02-13
UI/keep line breaks

nas-tabchiche:


eric-intuitem (changes requested):
there is still an issue in the requirement assessment detailed view.

nas-tabchiche:
Should be good to go now @eric-intuitem

eric-intuitem (approved):
LGTM

PR #61, opened 2024-02-21
Make The website work offline by using font-awesome locally

monsieurswag:


monsieurswag:

monsieurswag:
recheck

eric-intuitem (changes requested):
There is an interference with package-lock.json.
Shouldn't we use https://www.npmjs.com/package/svelte-fa instead?

eric-intuitem (changes requested):
This breaks library import symbol in docker-compose

eric-intuitem (approved):
With last commit it works well

PR #63, opened 2024-02-21
Ca 155 french translation

Mohamed-Hacene:


eric-intuitem (approved):
ongoing work, but we can already merge that part.

PR #65, opened 2024-02-22
fix: handle backend crash on startup if mailer failed

Mohamed-Hacene:


eric-intuitem (approved):
With added commit LGTM.

PR #69, opened 2024-02-23
Update functional-tests.yml

nas-tabchiche:


eric-intuitem (approved):
This makes sure functional tests are launched. So even they are currently broken, this is a progress.

PR #67, opened 2024-02-22
Ca 210 fix access control model in api

nas-tabchiche:

PR #70, opened 2024-02-23
Ca 203 api tests for user permissions and roles

Alexdev8:
### Add api testing for 6 more user groups than the administrator account and now verify that the user has the correct permissions according to the apps.py file
A quarter of the tests are currently failing because of various reasons including:
- **status code of 400  instead of 403** when an action is forbiden
- **missing permissions in the apps.py file** and consequently in the test_vars.py file leading to users that can perform some actions without explicit permission

#### I can't fix any remaning issues for the moment since I don't know which behavior is _normal_ or not
**_NOTE:_** A fix for theses issues has been worked on here #67 but needs to be improved


nas-tabchiche (approved):
LGTM. Although some fail, the tests are correct. It is now up to us to make these pass.

PR #73, opened 2024-02-25
Improve/rationalize object classes

eric-intuitem:
After studying the relations between our classes, it appears that it is possible to rationalise them.
- RootFolderMixin and FolderMixin are merged, with root folder being the default folder when creating an object. This will make it simpler in the future to allow objects to be creating either in the root folder or in a domain.
- Redundant fields are removed.
- created_at is moved in abstractBaseModel.
- Classes are ordered in a more logical way.
- RoleAssignment is attached to a folder (this was the last class not to be related to a folder).

Migrations are squashed.


eric-intuitem:
Important PR for data model stability

PR #78, opened 2024-02-26
Fix functional tests

Alexdev8:
This fix all the remaining problems
**Missing calendar tab needs to be fixed before merging**


Mohamed-Hacene:
> This fix all the remaining problems **Missing calendar tab needs to be fixed before merging**

However, if you go manually and check calendar tab, _'accordion-item-calendar'_ id is visible:

Screenshot from 2024-02-26 16-47-36

PR #84, opened 2024-02-28
repair pytest

eric-intuitem:
Pytest success is mandatory for main


Alexdev8:
I don't think that removing tests like this is a good idea. The remaining failing tests are actual problems in the API and should be fixed rather than ignored

nas-tabchiche (approved):
LGTM

PR #90, opened 2024-02-28
Put the is_publised attribute in the AbstractBaseModel class

monsieurswag:


eric-intuitem (changes requested):
Please see the new DoD.

nas-tabchiche (changes requested):
Just a few things before we can merge this:

- In object importers, just set `is_published` to `True` since imported library objects are always published anyway https://github.com/intuitem/ciso-assistant-community/blob/fc86564e9f56511f21411d129a118d32236c81b8/backend/library/utils.py#L281
- Add `is_published=True` to `FrameworkImporter`, `RequirementNodeImporter` and `RiskMatrixImporter` in import functions

eric-intuitem (approved):
Now LGTM. I have desquashed the migrations, and avoid useless overriding of field in library.

nas-tabchiche (approved):
LGTM

PR #96, opened 2024-02-29
Fix visible edit button for builtin objects

monsieurswag:
This solution seems good as long as the "builtin" attribute keep the same meaning for all the models using it.


Mohamed-Hacene (approved):
Ok

PR #103, opened 2024-03-03
status command

eric-intuitem:
The status command for manage.py allows to have a quick overview of the usage of the instance:
- number of users and how many are still first_login
- number of main objects


Mohamed-Hacene:
@eric-intuitem What does `first_login=False` mean? The user has logged himself more than once or he never logged in?

Mohamed-Hacene (approved):
Ok

PR #105, opened 2024-03-04
Ca 265 replace security measure function by reference applied control

Mohamed-Hacene:
**WARNING**: After merging this branch, the database will need to be reset because all occurences are replaced.


eric-intuitem (approved):
I have fixed the migration, and renamed existing measures to existing controls.
Now LGTM.

PR #108, opened 2024-03-04
Set up django-silk profiler

nas-tabchiche:


Mohamed-Hacene:
It works fine until I click on a specific request and get that error: 
Screenshot from 2024-03-04 16-35-26

nas-tabchiche:
> It works fine until I click on a specific request and get that error: 
> 
> Screenshot from 2024-03-04 16-35-26
> 
> 

Which version of python are you running?

nas-tabchiche:
This might be helpful to you: https://stackoverflow.com/a/52951224

Mohamed-Hacene:
Fix is to install lib2to3 specifying the python version we use: `sudo apt install python3.11-lib2to3` for python3.11

Mohamed-Hacene (approved):
Ok

PR #109, opened 2024-03-04
Ca 192 test that emails are correctly sent to the user

Alexdev8:
test the reset password email as well as the set password email


Mohamed-Hacene (approved):
Ok

PR #110, opened 2024-03-04
Ca 223 translate builtin groups

Mohamed-Hacene:
**WARNING**: After merging this branch, the database will need to be reset because **approver role** and **user group** have been updated following the spec.


eric-intuitem (changes requested):
Approver and global approver are missing, and shown as "Object object".

Mohamed-Hacene:
> Approver and global approver are missing, and shown as "Object object".
> 
> 

Does your database have been reset before testing ?

eric-intuitem (approved):
LGTM

PR #121, opened 2024-03-06
When changing the reference control of the edit form of an applied control the category doesn't change automatically to the one of the choosen refenrence control

monsieurswag:


eric-intuitem (approved):
LGTM

PR #127, opened 2024-03-07
Performance/requirement assessment detail performance

nas-tabchiche:
**This is not the final fix.**
This makes the page load 40-50% faster\*, but foreign keys are still fetched even if no form is opened.
The permanent fix is to fetch on form field mount. This is being worked on, a PoC is available on branch performance/modelform-fks. I will submit a PR for review once every test passes. This will make the initial page load at least 70% faster\*.

**\*** according to informal benchmarks I ran _on my machine_


eric-intuitem (approved):
LGTM

PR #128, opened 2024-03-08
Show an error modal when trying to access the scoring assistant without an imported risk matrix

monsieurswag:
Things we can do in the future :

- Improve the appearance of the modal. (the 'displayJSONModal' modal type is not meant to dispay a single string).
- We could store a boolean value in the sessionStorage to track if the user has an imported risk matrix available or not in the future in order to make the load time of the scoring assistant faster (because there will be no need to fetch each time we click on the link of the scoring assistant)

### DIFF
--- a/frontend/messages/en.json
+++ b/frontend/messages/en.json
@@ -65,7 +65,7 @@
 	"roleAssignments": "Role assignments",
 	"xRays": "X-rays",
 	"scoringAssistant": "Scoring assistant",
-	"scoringAssistantNoMatrixError": "Error: Your must import a risk matrix in order to access this page !",
+	"scoringAssistantNoMatrixError": "Please import a risk matrix from the library to get access to this page",
 	"libraries": "Libraries",
 	"backupRestore": "Backup & restore",
 	"myProfile": "My profile",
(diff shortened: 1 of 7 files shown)


eric-intuitem (approved):
OK for me with a gentler message.

eric-intuitem (changes requested):
OK, but functional test broken, to be fixed.

Alexdev8:
> The functional test errors surely come from this behavior :
> 
> const clickEvent = new MouseEvent('click', { bubbles: true, cancelable: false }); event.target.dispatchEvent(clickEvent);
> 
> When the user clicks on the scoring-assistant link the click event is cancelled with event.preventDefault(), and if there is at least one imported risk matrix the code dispatch a non-cancelable click event on the scoring-assistant link which loads the page of the scoring assistant. The test seems to no detect the page loaded because this system makes the page loading a bit different from the other sidebar links.
> 
> So the problem is not from the code of this PR but rather from the tests that don't support this behavior for the moment.

@monsieurswag I just pushed a fix for the tests.
The problem was that the scoring assistant page behavior changed depending on either there is no risk matrix imported and therefore a modal is displayed or either there is one and the page is loaded like usual. The test now handle both cases.

eric-intuitem (approved):
GG.

PR #129, opened 2024-03-08
Add a test to get the folder assigned to the user's user group

Alexdev8:
### Add a missing test to check that a user have access to the folder of the domain he is assigned to
This test is failing for 4 non-global user groups because it needs to be fixed in the backend


eric-intuitem:
This is fixed in PR#135, so once it is merged, this PR should work.

Mohamed-Hacene (approved):
Ok

PR #130, opened 2024-03-08
Fix library upload not working due to unsupported content-type in DRF

monsieurswag:


eric-intuitem (changes requested):
The download occurs, but there is a red error message in the GUI.

PR #131, opened 2024-03-08
fix: add missing fields_to_check in Project model

Mohamed-Hacene:


eric-intuitem (approved):
LGTM. Other ticket opened for error display.

PR #136, opened 2024-03-11
Ca 296 when clicking on a risk we get an ugly screen

Mohamed-Hacene:
Feel free to suggest some improvements on the layout !


eric-intuitem (approved):
LGTM, nice improvement

PR #137, opened 2024-03-11
The user can now remove a reference control in a saved applied control

monsieurswag:


eric-intuitem (changes requested):
functional test broken

eric-intuitem (approved):
LGTM

PR #138, opened 2024-03-12
Ca 197 show an error message to the user when he tries to create an object with the same name as another object in the same scope

Mohamed-Hacene:


eric-intuitem (changes requested):
There are missing fields_to_check, which is not implicit anymore. I have added one for Threat, but there are others to consider.

eric-intuitem (approved):
LGTM

PR #140, opened 2024-03-12
Update settings.py

eric-intuitem:
Create a SQLITE_FILE env variable to be able to set the sqlite file somewhere else


Alexdev8 (approved):
LGTM

PR #141, opened 2024-03-12
Ca 301 improve the e2e tests.sh script

Alexdev8:
### Use a different name for the test db and change the backend port name to a less common one (8173) to reduce the risk of an already used port.


eric-intuitem (approved):
LGTM. Incredible work, kudos!

PR #142, opened 2024-03-12
The error message of the library upload form is not undefined anymore

monsieurswag:
The error is not undefined anymore but the really well detailed backend error given by the django API isnt display in the frontend for now, so i will have to implement a "dynamic translation" (a runtime translation mechanism) to translate them later on.


eric-intuitem (approved):
LGTM

PR #143, opened 2024-03-13
Fix the error when submitting a threat edit form with a null value for the provider

monsieurswag:
I just changed the provider field to a nullable field in the zod schema used by superform since the provider field is declared as nullable in the backend :
```
provider = models.CharField(
    max_length=200, blank=True, null=True, verbose_name=_("Provider")
)
```


eric-intuitem (approved):
LGTM

PR #144, opened 2024-03-13
Fix the requirement assessment blank status field

monsieurswag:
The usage of bind:this to set the "select" attribute to the <option> element representing the default value is normal, i couldn't find any way to set the "select" attribute in a normal basic conditional block.
The select attribute was always disappearing once the component was rendered in the browser.


eric-intuitem (changes requested):
The status can be null (represented on the GUI as --), so no warning of any kind is expected.

eric-intuitem:
This seems solved in current main. Closing.

PR #145, opened 2024-03-13
Give the possibility to select items with the same name but different domains in Multiselect inputs

monsieurswag:
This problem is still not fixed when there are 2 risk scenarios with the same name and the same domain.
Should it be possible for 2 risk scenarios with the same name to be under the same domain or not ?


eric-intuitem:
Risk scenarios are scoped by domain/project/risk assessment. So 2 of them can have the same name except in the same risk assessment (which is currently not enforced).

eric-intuitem (changes requested):
Broken functional tests.
See also comment on spec.

Alexdev8:
If the scope is displayed in the selection,
image
I think it should also be visible in the object details here
image

monsieurswag:
Risk scenarios with the same name and the same project name are now supported (as long as the projects with the same name do not share a domain with the same too)
The solution is to display both the domain name and project name next to the risk scenario name, but if the domain name or/and the project name are long it may bother the user because in this case the text of the options in the multiselect will be long.

Mohamed-Hacene (approved):
0k

eric-intuitem (approved):
LGTM

PR #146, opened 2024-03-13
Clicking on the keyboard don't close the creation modal

monsieurswag:
When the creation modal is opened it will automatically make the user focus on the first text-like input, it makes things a little faster for the user and the user will never have the pain to have a modal closed because he typed on his keyboard before clicking on the input.


eric-intuitem (approved):
Useful fix, LGTM.

PR #147, opened 2024-03-14
Filter folders' queryset to domains

nas-tabchiche:


eric-intuitem (changes requested):
We need to be able to put assets and threats either in a domain or in the root folder (seen as "Global").
For other objects created by a user, they should be in a domain.
I have updated the architecture accordingly.

nas-tabchiche:
Creating a proxy model `Domain`, in the same fashion as `Policy` could save us a bit of headache here. This will make the distinction between domain and folder clearer in the codebase, and leave no room for confusion. I will do this first thing in the morning tomorrow

eric-intuitem (approved):
LGTM

PR #154, opened 2024-03-17
update

eric-intuitem:
pur
npm upgrade


eric-intuitem:
upgrading drf to version 3.15.0 breaks the API tests on GitHub, but not locally. To be investigated.

PR #156, opened 2024-03-18
feat: temporary hide my profile edit button for non-admin

Mohamed-Hacene:
**Temporary feature**

Since a non-admin user hasn't `change_user` permission, he cannot edit its personal account and get a 403. I propose to hide this button in v1.0 as a user could only change his name (not too restrictive). We'll need to add an exception to be able to update the personal account without `change_user` permission.


eric-intuitem (approved):
LGTM

PR #158, opened 2024-03-19
improve requirement assessment display

eric-intuitem:
- frontend rendering and styling -


Mohamed-Hacene (approved):
Ok, we should reverse status and title/description position to keep them inline

PR #166, opened 2024-03-20
Fix 404 when viewing user group's domain

nas-tabchiche:


frontend/src/lib/utils/crud.ts
```diff
@@ -273,7 +273,7 @@ export const URL_MODEL_MAP: ModelMap = {
 		localFrGender: 'm',
 		verboseName: 'User group',
 		verboseNamePlural: 'User groups',
-		foreignKeyFields: [],
+		foreignKeyFields: [{ field: 'folder', urlModel: 'folders' }],
```
nas-tabchiche, line 276: Large diff due to formatting (again), only change is here

Mohamed-Hacene (approved):
0k

PR #168, opened 2024-03-21
add python venv usage in functional tests

Alexdev8:
### It should remain an error for windows users because of the venv activation command that changes between unix and windows OS

### DIFF
--- a/frontend/tests/e2e-tests.sh
+++ b/frontend/tests/e2e-tests.sh
@@ -2,6 +2,7 @@
 APP_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 DB_DIR=$APP_DIR/backend/db
 DB_NAME=test-database.sqlite3
+VENV_PATH=$APP_DIR
 
 SCRIPT_LONG_ARGS=()
 SCRIPT_SHORT_ARGS=()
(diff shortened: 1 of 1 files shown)


frontend/tests/e2e-tests.sh
```diff
@@ -63,6 +72,15 @@ if [[ " ${SCRIPT_SHORT_ARGS[@]} " =~ " -h " ]] || [[ " ${SCRIPT_LONG_ARGS[@]} "
     exit 0
 fi
 
+ENV_CFG=$(find $VENV_PATH -name "pyvenv.cfg" -print -quit)
+if [[ ! -z $ENV_CFG ]]; then
+    VENV_PATH=$(dirname $ENV_CFG)
+    source $VENV_PATH/bin/activate
```
nas-tabchiche: We should `deactivate` to avoid contaminating the shell running the script

Alexdev8: I read that because it is a different shell that runs the script there is no risk of "contamination" and the venv is automatically deactivated just like all the exported env variables are lost after the script exit. I thought it would be simpler like this. But i could still make a condition in the cleanup to deactivate it if you think it is best.
Also what do you think about the problem with venv activation on a Windows OS ?

PR #181, opened 2024-03-30
Allow adding evidences from applied control detail

nas-tabchiche:
This PR adds the ability to manage applied controls' evidence through detail view. A **big** UX improvement would be to be able to create evidence directly from the create/update form.


eric-intuitem (changes requested):
Great improvement! To make it even better, we should improve the layout of the view page to make the evidence part visible on any common screen.

eric-intuitem (approved):
LGTM

PR #189, opened 2024-04-01
add base before URL to allow easy hosting with a single domain name

eric-intuitem:


eric-intuitem:
We will not use it for the moment, but keep the branch

PR #191, opened 2024-04-02
Fix undefined risk-assessment status

monsieurswag:
The risk-assessment status field is nullable, but the null value can't be translated especially because paraglide refuse to store null in its message dictionary (Error: Cannot compile message with ID "null").
So my only option was to use a ternary operator to translate null into "--".
You can change it to a better text before merging this PR if you find one.


eric-intuitem (approved):
LGTM

PR #192, opened 2024-04-02
Return an error message when an admin try to delete the only admin account

monsieurswag:
I removed import { fail } from 'assert'; in "frontend/src/routes/(app)/users/[id=uuid]/edit/+page.server.ts", it seems like it was an error surely from someone who auto-imported the wrong fail function in a previous commit.

According to this Mixin definition the "name" field isn't unique which means there may multiple UserGroup object with the same name :

```
class NameDescriptionMixin(AbstractBaseModel):
    name = models.CharField(max_length=200, verbose_name=_("Name"), unique=False)
```

I am not sure having 2 user groups with the exact same name should be authorized by the application, is this normal ?
If some evil user could create a user group with the same name as the admin group and bypass some permission checks.
To address this potential issue in my code i always added builtin=True when fetching the admin group since builtin objects are meant to be immutable.

### DIFF
--- a/backend/core/views.py
+++ b/backend/core/views.py
@@ -57,7 +57,7 @@ from core.utils import RoleCodename, UserGroupCodename
 
 from django.db import models
 
-from iam.models import User, RoleAssignment, Folder
+from iam.models import User, UserGroup, RoleAssignment, Folder
 
 User = get_user_model()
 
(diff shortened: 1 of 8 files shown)


eric-intuitem:
Indeed, once groups will be user-defined, we'll need to add     fields_to_check = ["name"]

eric-intuitem (changes requested):
Missing translation

monsieurswag:
Done

backend/iam/models.py
```diff
@@ -2,7 +2,7 @@
     Inspired from Azure IAM model """
 
 from collections import defaultdict
-from typing import Any, List, Self, Tuple
+from typing import Any, List, Self, Tuple, Self
```
Mohamed-Hacene: Remove `Self` duplication

backend/core/views.py
```diff
@@ -835,6 +835,26 @@ def get_queryset(self):
         # TODO: Implement a proper filter for the queryset
         return User.objects.all()
 
+    def update(self, request: Request, *args, **kwargs) -> Response:
+        user = self.get_object()
+        if user.is_admin() :
+            number_of_admin_users = User.get_admin_users().count()
+            admin_group = UserGroup.get_admin_group()
```
Mohamed-Hacene: It would be better to get admin group directly without using a method for it

monsieurswag:
Remove builtin and user_groups__builtin

Mohamed-Hacene (approved):
0k

eric-intuitem (approved):
LGTM

PR #200, opened 2024-04-03
Reset the model table page when switching between model tables of different type of objects

monsieurswag:


monsieurswag:
I made this PR as draft because i may an even better solution

eric-intuitem (approved):
LGTM

PR #204, opened 2024-04-05
Ca 352 prevent modal opening when mandatory objects are missing

Mohamed-Hacene:
This PR has a lot of commits because there is a problem with Fink online editor. It now pushes modifications in the root repository instead of the folder base location in `/frontend/messages`. It's nonsense because it's never done that before so I opened an [issue](https://github.com/opral/inlang-fink/issues/8) to know more about it.


eric-intuitem (approved):
LGTM

PR #230, opened 2024-04-08
The libary field of the frameworks detail view now display its name instead of its uuid

monsieurswag:


eric-intuitem (changes requested):
The library field is not useless, it should be displayed properly, i.e. with its name, not its UUID.

monsieurswag:
My bad i thought the "name" field always displayed the same name as the "libary" field, the problem is fixed

eric-intuitem (approved):
LGTM

PR #233, opened 2024-04-09
Remove the user list from the navigation sidebar for non-admin users

monsieurswag:
A non-admin can still access the user list going to the /users page by manually modifying its URL in the browser, so this is not a security fix at all.
This is just a simple frontend fix so that non-admin users don't have access to a page that is useless for them.


monsieurswag:
This PR is in draft until this problem is also fixed on the backend side

monsieurswag:
A non-admin user still have the permission to fetch the list of user-groups in the Django API but this is not really critical

PR #234, opened 2024-04-09
Implement memory caching for deserialized libary content

monsieurswag:
The response time of /api/libraries was 2930ms it's now 18ms.
Only the first request to /api/libraries takes ~3000ms but all the following requests will take ~20ms with this caching mechanism.


eric-intuitem (approved):
LGTM. Nice quick-win!

PR #235, opened 2024-04-09
Fix root assessment display (CA-360)

eric-intuitem:


Mohamed-Hacene (changes requested):
The tree seems ok in compliance assessment view but is broken in the framework or library detailed view. Moreover, we loose threats and reference controls suggestions

PR #238, opened 2024-04-11
Improve user detail view and edit view breadcrumbs

monsieurswag:
Display the first name and last name of the user in the breadcrumb when defined instead of the email address.


Mohamed-Hacene (changes requested):
Why not, but I would just add the case when first_name **AND** last_name are configured together (here we could have only the first name)

Mohamed-Hacene (approved):
0k

PR #241, opened 2024-04-11
Hotfix/missing translation

Mohamed-Hacene:


eric-intuitem (changes requested):
Still missing a few translations when clicking on the password set link:
"You can set the new password here.
Careful, the user will be disconnected if he has a session running."
"New Password"
"Confirm new password"

eric-intuitem (approved):
LGTM

PR #259, opened 2024-04-15
Make data persist in create forms after the closing the form without losing content

monsieurswag:
…submitting it


nas-tabchiche:
This might be achieved in a simpler fashion using [snapshots](https://kit.svelte.dev/docs/snapshots).

PR #265, opened 2024-04-15
chore: remove cookie rate limiter

Mohamed-Hacene:


ab-smith (approved):
Dummy secret and this param was never used

PR #266, opened 2024-04-15
add tisax framework

eric-intuitem:
- add tisax with conversion script
- improve convert_framework error message


ab-smith (approved):
Ok approuved, but the rendering of frameworks with shall and should is not optimal. the level attribute in the data model will be useful to improve this

PR #267, opened 2024-04-16
Ca 178 add score notion to compliance assessment

Mohamed-Hacene:


backend/core/models.py
```diff
@@ -1501,6 +1519,13 @@ class Status(models.TextChoices):
         verbose_name=_("Applied controls"),
         related_name="requirement_assessments",
     )
+    
+    def get_score_choices(self):
+        return [(i, i) for i in range(self.compliance_assessment.framework.min_score, self.compliance_assessment.framework.max_score + 1)]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._meta.get_field('score').choices = self.get_score_choices()
```
nas-tabchiche, line 1528: What is the point of getting score choices if they are just a range of integers between min_score and max_score?

frontend/src/lib/components/Forms/Score.svelte
```diff
@@ -0,0 +1,78 @@
...
+	
+	$: scoringEnabled = $value === null ? false : true
+
+	function preventNull(value: number){
+		if(value === null){
+			return 0
+		}
+		return value * 100 / max_score
+	
+	}
```
nas-tabchiche: The name of this function implies it is meant to prevent null values while it computes a percentage. This is misleading

backend/core/models.py
Mohamed-Hacene: It permits to be consistent on the API, to be sure we can only add a score in the range of min_score and max_score.

nas-tabchiche: This could be done at the validation stage. This way you would only have to check the boundaries instead of iterating over every single integer contained between min_score and max_score.

PR #273, opened 2024-04-17
Experimental: support token authentication for API

ab-smith:


ab-smith:
DRF Knox is a better option to explore, at least on paper

ab-smith:
will be merged with the work on #283

PR #276, opened 2024-04-17
Fix pagination overlow when changing the number of rows per page by a bigger value

monsieurswag:
When the pagination overflows e.g: "Showing 101 to 25 of 25" it will automatically the current page number to the last (non-empty) page.


eric-intuitem (approved):
LGTM

PR #278, opened 2024-04-17
Rename audit

Mohamed-Hacene:


eric-intuitem:
LGTM, but we should first rename "auditor" to "reader" to avoid confusion.

ab-smith (approved):

PR #300, opened 2024-04-20
i18n: Set up arabic localization with Fink 🐦

nas-tabchiche:


Mohamed-Hacene:
You juste need to add Arabic in this file: frontend/src/lib/utils/locales.ts and it should be ok

PR #301, opened 2024-04-20
add fadp

eric-intuitem:


ab-smith:
I get that adding not assessable items is useful for context but can become confusing and overwhelming

nas-tabchiche:
How about we add a toggle to show/hide non-assessable nodes in the UI?

ab-smith (approved):

PR #306, opened 2024-04-22
Format code base and update linting workflows

nas-tabchiche:
Only include formatting checks for now.

TODO: re-enable eslint and svelte-check once codebase is cleaned up


Mohamed-Hacene (approved):
0k

PR #315, opened 2024-04-23
Remove login_required decorator

nas-tabchiche:
The login_required decorator was used before we switched to DRF to redirect to login page if there were no user authenticated. Since we were using session authentication up until recently, the decorator had no impact on the proper functioning of the view, even though it overlapped with the permission_classes.
It broke when we switched to token-based authentication, since it checks for session objects, which are no longer used.


eric-intuitem (approved):
LGTM

PR #317, opened 2024-04-23
Allow deletion of referential objects whose urn does not start with "…

nas-tabchiche:
…urn:intuitem"

This is a quick and dirty way to allow deleting referential objects that are supplied by the user.

TODO: Come up with a better way to differentiate external referential objects from those packaged by intuitem.


eric-intuitem:
Let's rather wait for library management refactoring.

PR #320, opened 2024-04-24
Fix error 400 when exporting backup

nas-tabchiche:
This fixes the error 400 when attempting to export a backup.

The issue was due to dump_db_view being a function-based view, and thus not inheriting `settings.DEFAULT_AUTHENTICATION_CLASSES`, therefore assuming the authentication is session-based while it is in fact token-based.


Mohamed-Hacene (approved):
0k

PR #321, opened 2024-04-24
Ca 294 refactor library management

monsieurswag:


monsieurswag:
There are a lot of remaining things to do.
Clean dead code, check that only users with the correct permission can create/read/delete the libraries they can access to, maybe improve the performances of some operations, add the scoring related library fields to the LibraryMixin, update the data-model.md file, and obviously check that nothing has been broken elsewhere in the app and fix some little bugs.
There is no translation for the moment in the frontend.
And there is some kind of bug when attempting to delete a custom loaded libraries.
Also when i upload a copy of an existing StoredLibrary the library is not stored (surely because the URN already exist in the database), but there is a "success" type of toast message in the frontend after the operation which is not normal.

eric-intuitem (changes requested):
Almost good! But the functional tests are broken.

eric-intuitem (approved):
LGTM. Huge step forward!

PR #328, opened 2024-04-25
Stop using MIME types to infer attachment file extension on download

nas-tabchiche:
We used to infer the extension of attachments by checking their MIME types. This was unnecessarily complex as you can simply pass the attachment's file name through the `Content-Disposition` header. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition


ab-smith:
maybe cover xlsx as well since you're at it

eric-intuitem (approved):
LGTM

PR #329, opened 2024-04-25
fix: risk acceptance actions

Mohamed-Hacene:
Risk acceptance confirm modal needs a form to be triggered since superform V2. I added a simple "id" form juste to be able to trigger it, since we don't need any information about the form.


eric-intuitem (approved):
LGTM

PR #343, opened 2024-04-30
add transpose flag in matrix

eric-intuitem:
when set to true, this will allow swapping probability and impact in matrix. This is not yet managed in the frontend.


eric-intuitem:
It is preferable to set the transposition flag in the database and keep a single format for the library. This will avoid duplication of matrices.

PR #350, opened 2024-05-01
Add ISO27001 FR translate

protocolpaladin:
Add French version and complete 6.3 core point in version FR (Not ok in EN version)


protocolpaladin:

eric-intuitem:
There is a known issue with tests for contributions, we need to fix that before merging.

PR #363, opened 2024-05-04
Matrix generator

eric-intuitem:
Add matrix conversion in convert_framework.py


eric-intuitem:
Add colours for probability and impact, even if not yet implemented.
Seek colour on "name" column, and check consistency with matrix.

ab-smith (approved):

PR #364, opened 2024-05-04
Support multi-level and implementation groups

nas-tabchiche:


ab-smith:
Score radial bar needs to consider the IG filter

ab-smith (approved):
ok let's go

PR #371, opened 2024-05-05
Update startup-tests.yml

eric-intuitem:
avoid using a secret for postres password:
- this has no security benefit, as the service is running on localhost for the test runner
- this makes external PRs fail, because they are not allowed to get the secret. To be reviewed carefully before merge.


ab-smith:
Think about the exception for gitGuardian

ab-smith (approved):

PR #383, opened 2024-05-06
Proposition d'ajout de la PGSSI-S

krismas:
Proposition d'ajout de quelques éléments en lien avec la PGSSI-S (https://esante.gouv.fr/produits-services/pgssi-s) :

- Canevas de PSSI
- Exemple de matrice de risques


eric-intuitem:
Due to a known limitation, we need either to change the urn or to rename the file so as to align them. The simplest solution is to remove the 'v' in the yaml file names.

ab-smith:
I'll merge it and rename the file to keep the credits to @krismas in the git history

ab-smith (approved):

PR #384, opened 2024-05-06
remove pg in tests, use sqlite instead

eric-intuitem:
This will really solve the issue of external contributors tests failing, and is simpler.


ab-smith:
Won't the DB be more flexible for persistency cross-stage testing?

eric-intuitem:
Keep PG, will try another way

PR #393, opened 2024-05-07
#392_Fix creation of objects when parent is in root folder

nas-tabchiche:


ab-smith (approved):
LGTM,
wonder if we need an extra fetch for this though

PR #401, opened 2024-05-09
Create iso-42001-2023.yaml

eric-intuitem:
outline of the ISO 42001:2023 standard


AndrzejRPiotrowski:
Please add

ab-smith:
The license is pretty clear and restrictive on this, and putting the outline this way doesn't bring much value.

PR #404, opened 2024-05-09
400 add SecNumCloud 3.2

eric-intuitem:
Add SecNumCloud 3.2


ab-smith:
I wonder if it's not better to combine the appendix and let the auditor mark the 13 requirements as Not applicable when it's the case

ab-smith (approved):
LGTM

PR #414, opened 2024-05-12
Add support for BSI IT-grundschutz-kompendium edition 2023

ab-smith:


ab-smith:
@eric-intuitem double check the performances please

ab-smith:
locale is not rendered in the library store (undefined) we need extra string for DE

ab-smith:
replaced by #455

PR #416, opened 2024-05-12
Switch to CLoader for yaml

ab-smith:
Got about 7x speed in yaml parsing with this tweak, if the yaml-cpp library is installed


eric-intuitem (approved):
LGTM

PR #447, opened 2024-05-17
URNs containing whitespaces will now be considered invalid

monsieurswag:
The current URN regex allow messy URN with whitespaces.
According to the RFC 8141 URN characters must be ASCII but we currently allow any character including UTF-8 characters at the end of the URN.

```py
import re

URN_REGEX = r"^urn:([a-zA-Z0-9_-]+):([a-zA-Z0-9_-]+):([a-zA-Z0-9_-]+)(?::([a-zA-Z0-9_-]+))?:(.+)$"
URN = 'urn:myself:one:two:tree: 89 ( ( )) <> r"àét: :r:"é r:"é ré"ré"          ez㘃㘄㘅㘆ara 1<>   10é:✅✆✇✈✉✊✋ré :<:'

print(re.match(URN_REGEX,URN) is not None)```


eric-intuitem (approved):
LGTM

PR #455, opened 2024-05-18
BSI framework: IT-grundschutz-kompendium

ab-smith:


eric-intuitem (approved):
LGTM. This is a big one!

PR #461, opened 2024-05-20
Improve performance for 5G SCM

eric-intuitem:
It is sufficient to show them in each requirement assessment.


backend/core/helpers.py
```diff
@@ -309,12 +309,6 @@ def get_sorted_requirement_nodes_rec(
                             "status_display": req_as.get_status_display(),
                             "status_i18n": camel_case(req_as.status),
                             "style": "leaf",
-                            "threats": ThreatReadSerializer(
-                                req.threats.all(), many=True
-                            ).data,
-                            "reference_controls": ReferenceControlReadSerializer(
-                                req.reference_controls.all(), many=True
-                            ).data,
```
nas-tabchiche, line 317: Doesn't this break the display of threats and reference controls in the requirement trees?

ab-smith: The idea is to take them out of the tree and keep them only on the requirement assessment to accelerate the tree building

eric-intuitem: It simply removes them, and it's way faster to get the display.

ab-smith (approved):

backend/core/helpers.py
nas-tabchiche: Nothing that can't be fixed with prefetching. I am aware this branch was merged, but are we really sure we don't want to keep suggested threats/applied controls in the tree?

PR #464, opened 2024-05-20
avoid store conflict

eric-intuitem:
Avoid library store error if hash changed but the triplet (urn, locale, version) is unchanged.


monsieurswag (approved):
LGTM

PR #478, opened 2024-05-22
Fix/enisa 5g scm urns

eric-intuitem:
This will avoid trouble in the future when we will check the consistency of libraries, in particular the format of URNs.


Mohamed-Hacene (approved):
0k

Issue #473, opened 2024-05-22
Erreur d'export d'audit / Error on audit export -- community edition;


devoyon:
**Describe the bug**
Runiing multiple audits on same fraworks but on different projects.
When exporting audit you have a zip downloaded on your pc. The zip contain the wrong audit (in fact always provide the first audit of the framework done on CISO)

**To Reproduce**
1. Create 2 projects.
2. On both projects run an audit (used ANSSI hygienne)
3. On both, complete or fill the dedicated audits
4. Goes on second project -> audit -> export results
5. You get the export of the audit on the first project...

**Expected behavior**
Just having the export of the audit we are on :)

**Environment (please complete the following information):**
Ciso 1.3.3 build ce6ffaf
Community edition under docker under WSL2 (W11)
Chrome Version 124.0.6367.208 (Build officiel) (64 bits)

Mohamed-Hacene:
Hi @devoyon, 

I tried to reproduce your error twice, but I always got the right export. I'm going to run more tests to be sure I'm not missing something. Could you try to reproduce that behavior on your side with another framework ?

Mohamed-Hacene:
Ok, I was finally able to reproduce that behavior. Thank you very much we will investigate, you can follow the fix on the linked branch or wait for the PR.

Mohamed-Hacene:
The PR #476 should fix this issue, @devoyon could try to reproduce your case on the fix branch ?

PR #484, opened 2024-05-25
Add optional include_non_assessable parameter to ComplianceAssessment.get_requirement_assessments

tovam:
I'm currently developing a feature to export an audit report that contains all the applied controls for each of the requirements of the framework, which necessitates retrieving all requirement assessments of an audit, whether they are assessable or not.
I thought I would propose this change in case you're interested in it.

### DIFF
--- a/backend/core/models.py
+++ b/backend/core/models.py
@@ -1450,24 +1450,32 @@ class ComplianceAssessment(Assessment):
             if group.get("ref_id") in self.selected_implementation_groups
         ]
 
-    def get_requirement_assessments(self):
+    def get_requirement_assessments(self, include_non_assessable=False):
         """
-        Returns sorted assessable requirement assessments based on the selected implementation groups
+        Returns sorted assessable requirement assessments based on the selected implementation groups.
+        If include_non_assessable is True, it returns all requirements regardless of their assessable status.
         """
         if not self.selected_implementation_groups:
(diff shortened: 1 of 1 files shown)


Mohamed-Hacene:
> I'm currently developing a feature to export an audit report that contains all the applied controls for each of the requirements of the framework, which necessitates retrieving all requirement assessments of an audit, whether they are assessable or not. I thought I would propose this change in case you're interested in it.

Actually, the audit report contains all the applied controls for each of the requirements and display all of the requirement assessments whether they are assessable or not. If a requirement assessment is not assessable, it's not possible to link applied controls to it. If you're talking about getting only applied controls, it's already possible by exporting the action plan. 

However, the change you suggest in the function is a good thing, the name of the function is `get_requirement_assessments` and  not `get_assessable_requirement_assessments` so it should be possible to retrieve non assessable requirements too.

tovam:
> Actually, the audit report contains all the applied controls for each of the requirements and display all of the requirement assessments whether they are assessable or not. If a requirement assessment is not assessable, it's not possible to link applied controls to it. If you're talking about getting only applied controls, it's already possible by exporting the action plan.

During my tests the audit reports didn't show the controls but after double-checking I guess it's because of #473 

Anyway thanks for your answer, I just pushed a fixed commit for `get_requirement_assessments`

ab-smith (approved):

PR #486, opened 2024-05-26
Compliance assessments scoring system functional test

Alexdev8:
### This test aims to ensure that the scoring feature of compliance assessments is working properly and that the resulting score computation is correct.
The test takes approximately **18.7s** to pass locally and has an autocleanup function.


eric-intuitem (approved):
LGTM

PR #488, opened 2024-05-26
Fix functional tests github action artifacts

Alexdev8:
Artifacts were not uploaded anymore in the functional tests action since the split by browser of the tests


eric-intuitem (approved):
LGTM

PR #492, opened 2024-05-28
Locale/backend

Mohamed-Hacene:
- [x] translate x-rays
- [x] synchronize language tag between client/server side
- [x] translate reports

Only Audit html report is not translated, a special PR will have to be made for this.


eric-intuitem:
Need merge with main to resolve conflicts

eric-intuitem (approved):
LGTM

PR #501, opened 2024-05-30
Correction MCAS

krismas:
Correction de quelques codes et libellés


eric-intuitem (changes requested):
To enable a smooth upgrade, the version shall be incremented (version=2).
Changing the urn will provoke potential loss of data for the corresponding requirements. It's up to you to decide if it's worth doing it, but we should warn users and provide the list of changed URNs.

eric-intuitem (approved):
LGTM

PR #502, opened 2024-05-31
Display annotations in the frontend

monsieurswag:


eric-intuitem (changes requested):
Nice! We should manage \n correctly for better layout.
Can be tested with CCB-CFF

eric-intuitem (approved):
LGTM

PR #503, opened 2024-06-01
Order requirement nodes properly

eric-intuitem:
order children at the proper moment.


ab-smith:
There's a default value set to created_at. Are you sure that the `sorted` method parses that (date) without reformatting it?

ab-smith (approved):

PR #504, opened 2024-06-01
438: Support levels in ASVS

### LINKED ISSUE #438: ASVS: include the levels using implementation groups
**Problem statement**
In some frameworks (for instance Cyber Fundamentals in Belgium, ASVS from OWASP), depending on the maturity of the company, it is possible to select one level of complexity of the framework. By default in actual implementation, the max level of complexity is considered, as it usually comprises every controls of the framework. 
It would be easier in some cases to allow the user to select the level of complexity of the framework to fit its needs as well as saving some time. 
There are several solutions for this. The easiest (from my perspective) is to integrate in the JSON definition of the framework an attribute 'complexity' that defines the different levels and the related non-applicable controls (therefore, the max level complexity does not define any of these controls). When selecting the framework for an audit for instance, the user selects the complexity needed and the N/A controls are automatically filled in.

**Expected behavior**

1. Go to 'Audit'
2. Select 'New audit'
3. Select the framework needed for the audit
4. If the JSON definition contains the attribute 'complexity', then display the defined levels
5. Depending on the chosen level, pre-fill the N/A controls as defined in the JSON

**Additional context**
ASVS contains three levels : L1, L2 & L3 (see https://owasp.org/www-project-application-security-verification-standard/). The following JSON would be used : 
```
[...]
"complexity": {
    "L1": {
        "V1.1", "V1.2", [...]
    }
    "L2" : {
            [... N/A controls for L2]
    }
    "L3" : {}
}
[...]
```

ab-smith:
This will fix #438


eric-intuitem (approved):
LGTM

ab-smith:
doc updated as well:
https://intuitem.gitbook.io/ciso-assistant/features-highlights/library-upgrade

PR #513, opened 2024-06-03
Activate wal for sqlite

eric-intuitem:
set WAL mode for sqlite3


eric-intuitem:
We'll decide later if this is useful.

PR #514, opened 2024-06-04
Fix probability and impact unused colors

monsieurswag:
Colors shall be defined for each "probability" and "impact" row in each matrix library YAML file before merging this PR.


monsieurswag:
Colors should be displayed in some ways in /risk-matrices/{uuid}

eric-intuitem (approved):
LGTM

PR #516, opened 2024-06-04
Handle invalid library storing exceptions

monsieurswag:
This PR fixes the backend silently crashing because of a raised exception not being caughed when trying to store a library with invalid content.


ab-smith:
will be better with a toast indicating that "the file is not supported or malformed" , right now it's a silent error management

monsieurswag:
> will be better with a toast indicating that "the file is not supported or malformed" , right now it's a silent error management

This is a backend error message during the execution of either the "migrate" or the "storelibraries" command.
The error is not silent since it's directly written in the backend console.
And there can't be a toast because a toast is frontend only.

eric-intuitem (approved):
LGTM

PR #524, opened 2024-06-06
Fix screenshot filename in FileInput.svelte

tovam:
Two small changes to fix screenshot filenames : 

- date.getDay() -> date.getDate()
- date.getMonth() -> date.getMonth() + 1


eric-intuitem (approved):
LGTM
We will use iSO8601 format as an improvement

PR #528, opened 2024-06-07
User can add evidences from applied controls

monsieurswag:
This is an incomplete PR, the goal is to give user the ability to add evidences to its applied control directly from the detail view of its applied control.

Current problems :

- The toast isn't displayed in the frontend (this is surely due to not using SuperForm with my buttons, but we could maybe just directly trigger the toast in the frontend).
- The toast messages are not yet localized.
- If the user has a lot of evidences he will have a hard time listing the evidences currently linked to its applied control
  - Solution 1: Adding a <input type="checkbox" ...> to filter the ModelTable to it only displays the evidences currently linked to the applied control
  - Solution 2 (which is better): Displaying a second ModelTable in a second tabgroup (the first ModelTable of the tabgroup being one displaying the evidences linked to the applied control and the second ModelTable displaying all existing evidences and allowing the user to link these evidences to its applied control).

### DIFF
--- a/backend/core/views.py
+++ b/backend/core/views.py
@@ -34,7 +34,12 @@ from rest_framework.decorators import (
 from rest_framework.parsers import FileUploadParser
 from rest_framework.request import Request
 from rest_framework.response import Response
-from rest_framework.status import HTTP_400_BAD_REQUEST, HTTP_403_FORBIDDEN
+from rest_framework.status import (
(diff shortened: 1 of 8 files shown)


AndrzejRPiotrowski:
is this working solution ?

ab-smith:
not yet

PR #540, opened 2024-06-11
Patch skeleton's Modal and Toast components

nas-tabchiche:
They use @html tags, making them vulnerable to XSS.


Mohamed-Hacene (approved):
0k

PR #543, opened 2024-06-11
tuning of gunicorn

ab-smith:


ab-smith:
closing this for now and keeping it for future reference

PR #582, opened 2024-06-26
Use a symbol for the special actions field of the field component map

monsieurswag:
As "actions" was a special key with a different purpose than the other keys that can be declared in the field component map, and because we might one day have a field named "actions" in some object.

We can't let the "actions" key be enumerable and let it forbid us from naming an object field "actions" just because we couldn't use a custom component to display it in the model table.

Using a Symbol for a special key like this one is cleaner.


Mohamed-Hacene (approved):
0k

PR #585, opened 2024-06-26
Adding RODO framework

AndrzejRPiotrowski:
Adding polish Framework for GDPR - RODO


eric-intuitem:
@AndrzejRPiotrowski  can you convert the excel to yaml and add that in backend/library/libraries with the other libraries? So when we merge the PR the framework is available?
Btw there is a problem with the locale, you have indicated "en", while the requirements are in Polish, so the locale should be "pl".

eric-intuitem:
@AndrzejRPiotrowski as discussed, please let us manage the readme, and remove references to rodo.pl.
Btw we are working on a translation feature for libraries, this PR would be a very good candidate. I suggest you wait a few days and use the new feature.

eric-intuitem:
translation feature is (almost) out, @AndrzejRPiotrowski if your are OK we will integrate your translation directly in our GDPR checklist library.

eric-intuitem:
As agreed with @AndrzejRPiotrowski this is replaced by #691.

PR #595, opened 2024-06-28
Add filters to modeltable

Mohamed-Hacene:


eric-intuitem (changes requested):
Nice, but a few improvements are necessary:
- reference&applied controls: add filter by category
- assets: add filter by type
- Framework, stored and loaded libraries: add filter by provider

eric-intuitem (approved):
LGTM

Issue #338, opened 2024-04-26
Individual Control Assignment


andersond952:
**Problem statement**
I want to be able to assign individual controls within a framework to different user within an organization.

**Expected behavior**
When a user logs in they could go to a view to see the controls they've been assigned rather than having to sift through the audit on their own. Person in charge of the audit would not have to keep up with who is responsible for tracking down each control.

ab-smith:
part of Q2 roadmap indeed

PR #620, opened 2024-07-06
Add risk scenario owner capability

ab-smith:


ab-smith:
It's not exactly #338, but it is related, and we can consider the same logic.

eric-intuitem (approved):
LGTM, good preliminary step.

PR #628, opened 2024-07-08
Feat/mapping tools

eric-intuitem:
tool to create a xls for mapping
enhance convert_library to support mapping
with iso27001 example
rename yaml variables when missing urn suffix


ab-smith (approved):
functional ok

PR #631, opened 2024-07-08
Fix/use autocomplete for filters

nas-tabchiche:


nas-tabchiche:
@eric-intuitem mergeable once CI passes

nas-tabchiche:
Functional tests fail on firefox because of the libraries import tests taking more than 150 seconds to run, and thus time out. Fixing this on this branch.

eric-intuitem (approved):

PR #648, opened 2024-07-11
Feat/add requirement assessment view page

Mohamed-Hacene:
Create a read-only page for requirement assessments, to allow readers to see the detailed page


eric-intuitem (approved):
LGTM

PR #667, opened 2024-07-18
fix: add score and observation on audit report

Mohamed-Hacene:


ab-smith (approved):
functional ok,
we need to do a UI refresh of this later this year as it's getting too stacked.

Issue #664, opened 2024-07-17
Library/compliance : Impossible to re-upload / framework after deletion.


whoo:
**Describe the bug**
In the Compliance > Frameworks section, if a framework is deleted, it cannot be imported again.

**To Reproduce**
1. Go to 'Compliance > Frameworks'
2. Click on 'Import Framework'
3. Import NIST-CSF( as exemple) | The library should be successfully uploaded
4. Go to 'Compliance > Frameworks'
5. Delete NIST-CST | The library should be successfully unload
6. Go to 'Compliance > Frameworks'
7. Click on 'Import Framework'
8. you cannot import NIST-CSF anymore.

**Expected behavior**
Framwork could be loaded do Compliance>Frameworks.

**Environment (please complete the following information):**
 - Docker environnement.
 

Tks.

ab-smith:
thank you @whoo for the feedback. This is a regression indeed due to the caching system and will be fixed.
In the meanwhile, if you take away the framework from the `Loaded libraries` it shall do the trick for you:
<img width="435" alt="image" src="https://github.com/user-attachments/assets/3c50cab2-63d7-4fa0-898b-1b140e781855">

ab-smith:
fixed in v1.6.3

PR #677, opened 2024-07-21
Prevent direct deletion of library objects

eric-intuitem:
Library objects should only be deleted by unloading the corresponding library, to keep consistence of loaded libraries.


eric-intuitem:
fix to avoid #664

ab-smith:
this just blocks all deletion, which is not exactly what we want

ab-smith (approved):
functional ok now and we have one less click to do which is good

PR #684, opened 2024-07-27
translate matrix 5x5 iso27005 in French

eric-intuitem:


ab-smith:
ok for translation but rendering is broken
image

ab-smith (approved):

PR #686, opened 2024-07-29
add matrix 4x4 with 5 levels

eric-intuitem:
add matrix 4x4 with 5 levels

todo: change text color on black background


ab-smith (approved):
LGTM

PR #690, opened 2024-07-29
add NIS1 rules FR properly

eric-intuitem:
correct URN (no link to ANSSI)
FR and EN


ab-smith:
the audit export is not following the current locale anymore

ab-smith (approved):
merging this but we need to fix localization inconsistencies

PR #698, opened 2024-08-01
create mapping for cjis update

eric-intuitem:
remove old lib
fix error message


ab-smith:
let's add the version on the display name of the old library

ab-smith (approved):

PR #700, opened 2024-08-02
Fix signup closed error on SSO login

nas-tabchiche:


backend/iam/adapter.py
```diff
@@ -1,32 +1,23 @@
...
-from knox.views import LoginView
-from django.utils.http import url_has_allowed_host_and_scheme
-from urllib.parse import urlparse
 
 User = get_user_model()
 
 
-class MyAccountAdapter(DefaultAccountAdapter):
-    def is_open_for_signup(self, request):
-        return False
```
nas-tabchiche, line 28: Important change is here.

This method is used by allauth to check if a user is already registered. If signup is disabled and this is set to `False`, the login flow is interrupted.

This check is redundant since we already do it in `FinishACSView`. Furthermore, in some cases, SSO for users that are already signed up fails because of this for some reason.

I took a few hours to try and debug it but it proved to be too much of a hassle to bother with it since all there is to it is to check whether a user object exists in the database.

eric-intuitem (approved):
LGTM

PR #703, opened 2024-08-05
Fix objects_meta always counting 1 object even when there is more than that

monsieurswag:
Before fix: overview_problem_libraries
After fix: overview_problem_libraries_fix


ab-smith:
I don't see the fix on the branch, is it still ongoing?

eric-intuitem:
This works well for a new installation. But it does not fix existing installations, where the wrong counters are already loaded in the library. A custom migration could fix that.

ab-smith (approved):
functional ok

nas-tabchiche (approved):
LGTM

PR #715, opened 2024-08-07
Hotfix/idp initiated sso

nas-tabchiche:
This essentially removes redundant or irrelevant (for our case, as we are headless, use the SAML provider, have a single user adapter) steps that happen after the token is issued, and before the redirect to the frontend.


ab-smith:
image

looks like we've lost a piece of the error management. 
If the user has not been added yet, and logged in to O365 (let's say on a different tab), we end up in a situation where nothing is happening and we don't understand why; see message above

ab-smith (changes requested):
comment above

ab-smith (approved):
error management can be improved later on progressively

PR #720, opened 2024-08-08
use django 5.1

eric-intuitem:
pur to use django 5.1


ab-smith:
I've merged #716 that covers the security warning; we can prepare this without any rush

eric-intuitem:
LGTM.

ab-smith:
make sure to install the new requirements and re-run migrate

PR #725, opened 2024-08-09
Make filters persistents through URL

monsieurswag:


monsieurswag:
This pull request also add a filter based on the library content type

PR #736, opened 2024-08-13
fix: allow empty link

Mohamed-Hacene:
Link fields were not allowed to be empty once they have been filled. This PR fix that.


eric-intuitem (approved):
LGTM

PR #737, opened 2024-08-14
Add impact qualification

monsieurswag:


backend/core/models.py
```diff
@@ -1680,6 +1680,23 @@ class RiskScenario(NameDescriptionMixin):
...
 
+    QUALIFICATIONS = [
+        ("Financial", _("Financial")),
+        ("Legal", _("Legal")),
+        ("Reputation", _("Reputation")),
+        ("Operational", _("Operational")),
+        ("Confidentiality", _("Confidentiality")),
+        ("Integrity", _("Integrity")),
+        ("Availability", _("Availability"))
+    ]
```
nas-tabchiche: We should allow impact qualifications to be specified in risk matrices, as it is done for strength of knowledge. Most likely some users will want to have their own

eric-intuitem: The default values can be what is proposed here.

eric-intuitem (changes requested):
1/ The qualification can be multivalued (eg financial + confidentiality).
2/ I The addition of the status is useless, and in fact the added status is dead code.

eric-intuitem (changes requested):
Almost good! Just need to translate the qualification values.

eric-intuitem (approved):
LGTM

PR #749, opened 2024-08-19
fix: handle no first/last name during sso login

Mohamed-Hacene:
During the SSO login, if the user has not configured his first name or last name on the IdP, an error will be triggered.
This PR prevents that.

### DIFF
--- a/backend/iam/sso/saml/views.py
+++ b/backend/iam/sso/saml/views.py
@@ -135,14 +135,14 @@ class FinishACSView(SAMLViewMixin, View):
             email = auth._nameid
             user = User.objects.get(email=email)
             idp_first_name = auth._attributes.get(
-                "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname"
+                "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname", [""]
             )[0]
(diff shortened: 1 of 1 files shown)


backend/iam/sso/saml/views.py
```diff
@@ -135,14 +135,14 @@ def dispatch(self, request, organization_slug):
             email = auth._nameid
             user = User.objects.get(email=email)
             idp_first_name = auth._attributes.get(
-                "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname"
+                "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/givenname", [""]
```
nas-tabchiche, line 138: As `idp_last_name` is a string, why fall back to `[""]` and not simply `""`?

Mohamed-Hacene: because `_attributes` returns an array containing the `givenname` or the `lastname`.
`idp_last_name` is a string because we get index 0 of this array directly.

nas-tabchiche: Kinda didn't read the following line, my bad. LGTM

nas-tabchiche (approved):

PR #751, opened 2024-08-21
hotfix: firefox pipeline deactivation

Mohamed-Hacene:
Functional tests on firefox browser are uncertain due to dependency problems with playwright on GitHub Actions.
This PR deactivate it to to prevent checks from failing randomly.


eric-intuitem (approved):
LGTM. We need to rework that later.

PR #756, opened 2024-08-22
Avoid initialisation error in case of welcome mail sending failure

eric-intuitem:
Make all tasks before sending mail, especially putting a superuser in the global admin group. Otherwise, the superuser can be created without any right, which makes the instance unusable.


nas-tabchiche (approved):
LGTM, merging

PR #762, opened 2024-08-26
Fix audit change detection

### LINKED ISSUE #758: Better manage the change detection during a requirement assessment
when in an edit page of a requirement assessment, the browser will always pop up the dialog to confirm that unsaved changes will be lost even when no change was made, this is becoming annoying when we have multiple items that we want to browse quickly

monsieurswag:
Simple code to avoid encountering the "Do you want to leave this page? Changes you made may not be saved." message when trying to leave the audit edit view without having modified anything.
will fix #758


eric-intuitem (approved):
LGTM. There are other similar situations, but this is a good step forward.

PR #767, opened 2024-08-27
Fix layout shifting

nas-tabchiche:
A squote character shifts the AppShell layout down. A quick git blame revealed that it was my mistake, sorry!


Mohamed-Hacene (approved):
0k

PR #777, opened 2024-08-29
Add Risk Acceptances workflow testing

Alexdev8:
## This test checks that risk acceptances can be accepted, refused and revoked and that a non approver user cannot be selected as an approver for a risk acceptance.
Duration: approx. 28s
> This test could maybe also include the check that a non approver user cannot effectively approve a risk acceptance he is not assigned to. 
**However, this addition would significantly extend the duration of the test and I didn't think it would be worth**


Alexdev8:
The `logedpage` parameter removed in the commit [6d418a3](https://github.com/intuitem/ciso-assistant-community/pull/777/commits/6d418a3efc6e29c103c8c496b3d30857ecccdd3d) is not in fact unused. It is a fixture called by Playwright at the start of the test and removing it inevitably cause it to fail.

I don't know how to pass the CodeFactor check with this parameter. Maybe just by calling `logedPage` in the test for no other reason than the GitHub check. Or we can just ignore the CodeFactor check.

PR #780, opened 2024-08-30
feat: allow sso configuration without metadata url

Mohamed-Hacene:
Allow to configure SSO with Identity providers which don't provide a metadata URL


eric-intuitem (approved):
LGTM

PR #807, opened 2024-09-08
Add cost field to applied controls

### LINKED ISSUE #806: Add cost on applied Controls
It could be useful for reporting, communication, and reporting to have the cost of controls.
We can have the currency as a global setting and just a float number for the cost.

eric-intuitem:
Currency not yet managed.
NumberField needs improvement for decimal management
will fix #806


eric-intuitem:
implements #806

nas-tabchiche (approved):
LGTM

PR #808, opened 2024-09-08
Improve Cyfun by highlighting key measures

eric-intuitem:
Add 3 implementation groups, and add a header in the description.


backend/library/libraries/ccb-cff-2023-03-01.yaml
```diff
@@ -69,6 +69,15 @@ objects:
     - ref_id: E
       name: essential
       description: null
+    - ref_id: BK
+      name: basic - key measure
```
ab-smith: key measures is probably a better title for the IG

ab-smith (approved):

PR #824, opened 2024-09-11
Add enclave paradigm for TPRM implementation

eric-intuitem:
add enclave principle
add conclusion field
add observation field for all assessments


nas-tabchiche (approved):
LGTM

PR #828, opened 2024-09-13
Fix potential information disclosure

nas-tabchiche:
serializers.ValidationError is passed to the ViewSet and returned to the client. Displaying the message of the exception WILL cause unwanted information disclosure.


eric-intuitem (approved):
LGTM

PR #831, opened 2024-09-15
Update NZISM to Version 3.8

Coffee-007:
Updating to the latest version of NZISM 3.8.
Moved from individual files per framework version to a major framework version (3) with minor versions as this is changed by the NZ Government.


eric-intuitem (approved):
LGTM

PR #832, opened 2024-09-17
Fix default folder value in migrations

nas-tabchiche:
Use Folder.get_root_folder_id instead of get_root_folder


ab-smith:
we can reopen it if we can reproduce the issue

PR #841, opened 2024-09-18
fix: add tags in applied control list inside scenarios page

Mohamed-Hacene:


nas-tabchiche (changes requested):
The tag does not show in risk scenario update view

image

nas-tabchiche (approved):
LGTM

PR #847, opened 2024-09-19
Front end Bug correction on calendar

melinoix:
**Front end bug correction on calendar**: removed the bug that showed shaded boxes on every september of every next years which is not suppose to happen.


melinoix:

melinoix:
recheck

melinoix:
recheck

PR #849, opened 2024-09-19
added functional front end tests - about section

melinoix:
added functional front end tests in the about section to see if there is a build and a version well printed on screen.


nas-tabchiche (approved):
LGTM

PR #851, opened 2024-09-19
Update iso27001-2022.yaml

eric-intuitem:
- add implementation groups to keep SoA or not.
- define annex A as reference security controls


ab-smith (approved):
lgtm, could be useful to have the reference controls as recommendations

PR #856, opened 2024-09-22
refactor/code duplication

nas-tabchiche:
- **Write defaultCreateFormAction function**
- **Write getSuccessMessage, getHTTPMethod functions**
- **Use defaultWriteFormAction for default model route**
- **Use defaultWriteFormAction for nested create forms in default model route**
- **Use defaultWriteFormAction for edit forms in default model route**


eric-intuitem (approved):
LGTM

PR #858, opened 2024-09-23
Changing domain manager permissions

melinoix:
Domain manager permissions are changed to allow to create reference control inside his domain/sub-domains

### DIFF
--- a/backend/core/startup.py
+++ b/backend/core/startup.py
@@ -176,6 +176,9 @@ DOMAIN_MANAGER_PERMISSIONS_LIST = [
     "view_user",
     "view_requirementmappingset",
     "view_requirementmapping",
+    "view_appliedcontrol",
+    "change_appliedcontrol",
+    "delete_appliedcontrol",
 ]
 
 ADMINISTRATOR_PERMISSIONS_LIST = [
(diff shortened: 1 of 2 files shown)


backend/core/startup.py
nas-tabchiche, line 1: The task's subject was to allow the creation, edition and deletion of a `ReferenceControl`, not `AppliedControl`. Furthermore, `view_appliedcontrol` is present twice in the list. Once on [line 132](https://github.com/intuitem/ciso-assistant-community/blob/a379d48dd178d6af9203da86143a917a72f2bae0/backend/core/startup.py#L132) and once on line 179.
For the first point, after reading **my own** comment under the ticket, I realized I wrote `_appliedcontrol`. My mistake. Still, a domain manager has no business managing applied controls 😉

nas-tabchiche (changes requested):
### To do

Change `{view,change,delete}_appliedcontrol` to `{view,change,delete}_referencecontrol`.

backend/core/startup.py
eric-intuitem: Let's keep it simple: a domain manager can do anything in its domain. The single exception is approval.
| Domain manager      | full access to selected domains (except approval), in particular managing rights for these domains. Read access to global objects |

nas-tabchiche:
Closing this in favor of #893

PR #859, opened 2024-09-23
Disabling score when Not applicable selected

melinoix:
Disabling score editing when the result is "not applicable".


nas-tabchiche (approved):
LGTM

PR #860, opened 2024-09-23
Added a message when no X-rays available

melinoix:
Added a message when the user should create a project to use the x-rays functionality.


nas-tabchiche (approved):
LGTM

PR #861, opened 2024-09-23
Enhance/tprm

Mohamed-Hacene:
- Add a "send mail" button to send the internal questionnaire
- Align TPRM permissions with base access control model
- Export the questionnaire
- Allow base audit view for third party
- Add a tag to know if requirements contain questions
- Improve EntityAssessment/Audit ownership
- Improve requirement assessment edit page server complexity


frontend/src/lib/components/DetailView/DetailView.svelte
nas-tabchiche, line 1: We should not put model-specific logic and components here, but rather use slots.

frontend/src/routes/(app)/(internal)/entity-assessments/[id=uuid]/+page.server.ts
```diff
@@ -1,7 +1,49 @@
...
+import * as m from '$paraglide/messages';
+import { setFlash } from 'sveltekit-flash-message/server';
+import { safeTranslate } from '$lib/utils/i18n';
 
 export const load: PageServerLoad = async (event) => {
 	return loadDetail({ event, model: getModelInfo('entity-assessments'), id: event.params.id });
 };
+
+export const actions: Actions = {
+	mailing: async ({ request, fetch, cookies }) => {
```
Comment (nas-tabchiche) [inline frontend/src/routes/(app)/(internal)/entity-assessments/[id=uuid]/+page.server.ts:17]: `mailing` is not a very explicit name. Would we not gain better readability if we actually stated what this action does? E.g. `emailQuestionnaire`...
The URL of the endpoint should also be more specific than `/entity-assessments/mailing`

Comment (Mohamed-Hacene) [reply frontend/src/routes/(app)/(internal)/entity-assessments/[id=uuid]/+page.server.ts:17]: Yes I agree for the name, we'll need to change it anyway to add another action to send the external questionnaire by mail.
However, the endpoint is `/compliance-assessments/mailing` because we send the audit and not the entity-assessment. It is not explicit because it could be useful to send other audits than questionnaires in the future with the same endpoint as it is the same process but maybe with a different template.

Comment (nas-tabchiche) [reply frontend/src/routes/(app)/(internal)/entity-assessments/[id=uuid]/+page.server.ts:17]: Was a typo, meant compliance-assessment/mailing.
The point is that the word 'mailing' does not explicitly convey what the endpoint actually does. That is, emailing the questionnaire to the designated representatives

Comment (Mohamed-Hacene) [reply frontend/src/routes/(app)/(internal)/entity-assessments/[id=uuid]/+page.server.ts:17]: Actually not, the endpoint is sending an audit to its authors so the name cannot be focused on the questionnaire, but it should be more precise I agree, like "send_mail" or "send_audit" maybe

nas-tabchiche (approved):
LGTM

PR #869, opened 2024-09-25
corrected message error on front when conflicting parameters

melinoix:
Fixed message error on front end that was not showing properly in the case of conflicting parameters ( same name, version, etc...)


Mohamed-Hacene (approved):
0k

PR #870, opened 2024-09-25
chore: Add nested evidence list in policies (like in applied controls)

Axxiar:
If you go inside an applied control, you’ll see below its details, a table with its evidence and the possibility to manage them from the page.

We should see the list of evidence linked to a policy and be able to add them from the detail view like in an applied control.


Mohamed-Hacene (approved):
0k

PR #876, opened 2024-09-28
Revert "fix delete library broken"

ab-smith:
This reverts commit 3d2d6a73d3e2e2f001d088df61561c1fd75fe0c9.


eric-intuitem (approved):
Hotfix but we still need to fix the regression on delete library.

PR #879, opened 2024-09-29
fix missed change for library CRUD

eric-intuitem:
fix missed change


nas-tabchiche (approved):
This fixes both the show and delete issues. Merging.

PR #888, opened 2024-10-01
Add a global setting for the default language

monsieurswag:
The translations are missing.
The global setting work but is not effective yet.


Mohamed-Hacene:
Splitting this PR in two different parts to avoid CI issues

PR #891, opened 2024-10-02
Add library content type filter

monsieurswag:


eric-intuitem (dismissed):
Looks good. Can we add the same filter for loaded libraries?

nas-tabchiche (approved):
LGTM

PR #896, opened 2024-10-03
Modify the evidence file stream behavior to stop reading when being terminated by the client

monsieurswag:


Mohamed-Hacene (approved):
0k

PR #898, opened 2024-10-04
Ca 471 implement license expiration

nas-tabchiche:


eric-intuitem:
- Expiration env variable shall be in the format yyyy-mm-dd. This is currently not supported.
- The expiration date (and the number of seats) shall be logged on startup.
- Translations are necessary.

Mohamed-Hacene (approved):
0k

PR #921, opened 2024-10-11
feat: improve risk assessment duplication

Mohamed-Hacene:
Currently, if we duplicate a risk assessment in a different domain, the applied-controls, threats and assets will stay attached to the previous domain. To avoid this, we are duplicating these objects in the new domain to make it easier to track them and avoid conflicts.


eric-intuitem (dismissed):
In duplicate_related_objects, the variable duplicate_risk_assessment should not be used, instead an additional argument "target_folder" should be added, with target_folder=duplicate_risk_assessment.project.folder.
Indeed, is is better to avoid global variables, all the more if they are defined later.

eric-intuitem:
If the obj is in a parent folder of the target, then it should be kept as is. This means we should invert the if / elif conditions in duplicate_related_objects.

eric-intuitem:
The duplicate method shall be documented.

Mohamed-Hacene:
> If the obj is in a parent folder of the target, then it should be kept as is. This means we should invert the if / elif conditions in duplicate_related_objects.

This is actually what the function does. But we must check at first if the object exists in the target folder and then check the parent folders to avoid conflicts, because you can have two or more objects with the same name between the target folder and its parents.
I've made some changes to make it clearer, because the function could be better organised as you mentioned.

nas-tabchiche (approved):
LGTM, gj

PR #926, opened 2024-10-11
Avoid published objects to be visible inside enclaves

eric-intuitem:
enclave does not allow published objects visibility


backend/iam/models.py
```diff
@@ -703,7 +703,9 @@ def get_accessible_object_ids(
                         ):
                             permissions_per_object_id[object.id].add(p)
 
-        if hasattr(object_type, "is_published"):
+        if folder.content_type != Folder.ContentType.ENCLAVE and hasattr(
```
Mohamed-Hacene: We can't check for enclave content type on `folder` variable because we need to check every folder where the user has a view permission (the `folder` variable here can be a folder where the user has no permissions like the Global one for example. It defines the starting folder in which we look for accessible objects). we need to check it for each folder in `folders_with_local_view`.

eric-intuitem: A user confined to an enclave should not have any role assignment outside of the enclave folder. If this is true, then the proposed change avoids that the published objects are visible inside the enclave.

Mohamed-Hacene (approved):
0k

PR #927, opened 2024-10-11
show questions in cmmc

eric-intuitem:


ab-smith (approved):
ok,
we need to review this id generation approach (& and *) as it can breaks down the road

PR #932, opened 2024-10-14
add to mappings a import library button

melinoix:
Add to mappings tab an import mappings button to library


eric-intuitem:
Almost OK, just automatically add a filter on "mapping" object type when displaying the libraries.

nas-tabchiche:
> Almost OK, just automatically add a filter on "mapping" object type when displaying the libraries.

It is not currently possible to pass filters through URL in the frontend. As this PR depends on another, yet to be fulfilled task, I suggest we merge it.

PR #937, opened 2024-10-14
Ca 468 enterprise name should be optional for enterprise version and should be used for main entity

nas-tabchiche:


Mohamed-Hacene (approved):
0k

PR #940, opened 2024-10-15
Add applied control duplication

monsieurswag:


Mohamed-Hacene (changes requested):
- [ ] Use a slot for duplication button inside the DetailView
- [ ] Add a checkbox to bring the evidence if needed

monsieurswag:
We don't need a to use a slot now, the code can be easily adapted to add duplication for other objects.
The checkbox to bring the evidences has been added, check its behavior in the backend (some evidences are just linked while other are duplicated depending on their folder)

backend/core/views.py
```diff
@@ -1039,6 +1043,47 @@ def get_timeline_info(self, request):
...
+            effort=applied_control.effort,
+            cost=applied_control.cost,
+        )
+        if data["duplicate_evidences"]:
+            duplicate_related_objects(
+                applied_control, duplicate_applied_control, new_folder, "evidences"
+            )
+            duplicate_applied_control.save()
+
+        return Response({"results": "applied control duplicated"})
```
nas-tabchiche: Let's return the actual controls as JSON

Mohamed-Hacene (dismissed):
- Field `has_evidence` (and `eta_missed` by the way) should be excluded. We cannot add evidence from the applied control anymore.
image
- The checkbox has unrelated text with this PR
image

ab-smith:
Pensez au squash au moment du merge

eric-intuitem (approved):
LGTM

eric-intuitem (approved):
LGTM

PR #941, opened 2024-10-15
chore: Fix thirdParty translations

Axxiar:
Translating “third party” to “gestion des tiers” results in wrong badge for TP users.

The ‘third party’ label in the sidebar should have its own identifier.


nas-tabchiche (approved):
LGTM

PR #943, opened 2024-10-15
Added a next link when editing audit in analytics 

melinoix:
Added a next link when editing audit in analytics to go back to analytics ( the cancel and update button were just updating the data but were not returning to a page).


Mohamed-Hacene (approved):
0k

PR #947, opened 2024-10-16
Ca 519 move from local items to safe translate to translate dynamic strings

monsieurswag:


monsieurswag:
This PR will surely cause a few silent translation errors but it's necessary to make the code cleaner

Mohamed-Hacene (approved):
Great, 0k

PR #948, opened 2024-10-16
fix: Error not displayed in nested list forms

melinoix:
fixed the errors that were not displaying in the nested list forms while it was good for the list forms.


Mohamed-Hacene (approved):
0k

PR #950, opened 2024-10-16
chore: improve question layout for Firefox

Axxiar:
When the question is too long in requirement assessments edit page, the answers are not aligned correctly.


nas-tabchiche (approved):
LGTM

PR #954, opened 2024-10-17
Display folder field in evidence when creating from it from an applied control

nas-tabchiche:
As an applied control may be global, inheriting its domain may cause errors if the user attempting to create an evidence inside it may not have the permission to create an evidence in the global domain.


Mohamed-Hacene (changes requested):
We also need to filter the "available" domains, because we could add an evidence in applied control with a domain which has no parent relationship

Mohamed-Hacene (approved):
Let's merge it for now, to be able to add an evidence in a global applied control withtout having a global role.

PR #956, opened 2024-10-17
Revert "Display folder field in evidence when creating from it from an applied control"

nas-tabchiche:
Reverts intuitem/ciso-assistant-community#954


Mohamed-Hacene (approved):
0k

PR #960, opened 2024-10-18
filter selected implementation groups in table mode

melinoix:
fix : filter requirement assessments by selected implementation groups in table mode.


nas-tabchiche (approved):
LGTM

Mohamed-Hacene (commented):
It could be solution to do it, however, we loose information by excluding basic requirements in the table mode which could be helpful to fill it. We should discuss about it before merging.

Mohamed-Hacene (approved):
0k

PR #961, opened 2024-10-18
chore: remove description inheritance from reference control to appli…

Mohamed-Hacene:
…ed control during suggestion


nas-tabchiche (approved):
LGTM

PR #970, opened 2024-10-22
deselect a choice in flash and table mode

melinoix:
Can deselect a choice in flash and table mode. If its a question in table mode, it can deselect and put null value , if its a status or a result , the deselection will put again on Todo or Non assessed.


eric-intuitem (approved):
LGTM

PR #971, opened 2024-10-22
Improved error handling when license expiration is unset

melinoix:
Improving the error handling when license expiration is unset by catching error when licence_expiracy = "unset"


enterprise/backend/enterprise_core/views.py
```diff
@@ -203,8 +203,11 @@ def get_build(request):
     LICENSE_SEATS = settings.LICENSE_SEATS
     LICENSE_EXPIRATION = settings.LICENSE_EXPIRATION
     try:
-        expiration_iso = datetime.fromisoformat(LICENSE_EXPIRATION)
-        license_expiration = date_format(expiration_iso, use_l10n=True)
+        if LICENSE_EXPIRATION and LICENSE_EXPIRATION.lower() != "unset":
```
nas-tabchiche: I would advise to wrap the date cast in a try/except block, instead of explicitly checking for an "unset" string.

enterprise/backend/enterprise_core/views.py
```diff
@@ -203,8 +203,11 @@ def get_build(request):
...
+            expiration_iso = datetime.fromisoformat(LICENSE_EXPIRATION)
+            license_expiration = date_format(expiration_iso, use_l10n=True)
+        else:
+            license_expiration = "No expiration date set"
```
nas-tabchiche: We should have an error identifier here instead of an english string, e.g. "noExpirationDateSet". This would make it easier to translate it in the frontend.

Mohamed-Hacene (dismissed):
The error is also triggered here:
image

nas-tabchiche (approved):
LGTM

PR #974, opened 2024-10-22
Ca 548 remove svg from accepted logo formats

Axxiar:
Enterprise : Remove svg from accepted logo formats


ab-smith (dismissed):
extensions are not aligned between back and front,
also let's keep only png, jpg, jpeg

nas-tabchiche:
Looks good to me, we should squash the migrations so that only a single migration file is brought to `main`. Let's discuss that tomorrow morning

eric-intuitem (approved):
LGTM

PR #977, opened 2024-10-23
svelte 5 migration prep

ab-smith:


ab-smith:
the missing part will be to manage the multi slot components, we only have two and this will have to be done as part of the migration

nas-tabchiche (approved):
Looks good to me

PR #978, opened 2024-10-23
Fix TableRowActions button sizes

monsieurswag:
Quick fix to make the TableRowActions buttons as they were before with their right size, the current TableRowActions are too small.


ab-smith (approved):
I've removed that on purpose to complement it with a hover effect but results in an unwanted wiggle for some tables, we can revert it for now

PR #981, opened 2024-10-25
display toast on login when license is nearing expiration

melinoix:
Display toast when logging in when license is nearing expiration. I created a env variable to choose this "nearing expiration" in days, you can change it with "export PUBLIC_LICENSE_EXPIRATION_DAYS=<nb in days>", otherwise the default one is 30 days.


eric-intuitem (approved):
LGTM

PR #999, opened 2024-10-30
Fix 500 error with CJIS flash mode

eric-intuitem:
avoid crash on None when implementation group is selected.


ab-smith (approved):
functional ok

PR #1005, opened 2024-11-02
add mitre attack v16

eric-intuitem:
fix urn for v14 to enable proper upgrade
add updated library
change default pagination in backend (workaround, we need a proper fix for that).


ab-smith (approved):
the pagination is not ideal we need to rework this asap

PR #1007, opened 2024-11-03
remove useless ig code

eric-intuitem:
This code provokes log errors and is useless. Translating IG needs another approach.


nas-tabchiche (approved):
LGTM

PR #1021, opened 2024-11-07
fix loaded library issue

eric-intuitem:
previous migration (0020) has altered the optionLabels for frameowrk


nas-tabchiche (approved):
LGTM

PR #1023, opened 2024-11-08
Handle SVG logos

melinoix:
The SVG logos were not handled because of the 'charset-8' specification in the mimeType.


eric-intuitem (approved):
LGTM

PR #1040, opened 2024-11-12
Move TOTP modal trigger logic to FormCard component

nas-tabchiche:
Since this component is responsible for triggering the TOTP input modal on the login page, the actual triggering logic belongs in it, rather than in its parent page.


Mohamed-Hacene (dismissed):
0k

PR #1041, opened 2024-11-12
Fix question reset in requirement assessments after framework update

monsieurswag:


eric-intuitem:
We need to refactor the data model for answers.

Mohamed-Hacene (approved):
0k

PR #1042, opened 2024-11-12
Fix eager datetime coercions

melinoix:
fixed eager datetime coercions by precising more what a date format is to the ISO regex so when you type "1000" as a name it will not show a date but just the number.


eric-intuitem (changes requested):
This is not solving the issue properly, as "2024-12" would still be displayed erroneously. The formatting of date fields should depend on the field, not on its content.

eric-intuitem (approved):
LGTM

PR #1043, opened 2024-11-12
feat: Recommend setting groups after user creation

Axxiar:
After user creation, it is not obvious that the user has no rights. We should add a recommendation to edit the user and add it to relevant groups to grant appropriate rights.


eric-intuitem (approved):
LGTM. Will need to add all translations with fink.

PR #1044, opened 2024-11-13
Ca 545 add ref id to all editable objects

melinoix:
Added a reference ID to all editable projects : risk-assessments, domains, projects, compliance-assessments, applied-controls.


nas-tabchiche (dismissed):
In almost all cases, a single PR should not have more than one single migration per app. This is mergeable once migrations are squashed and tests pass.

melinoix:
-Have to modify tests/Api tests 
- projects tab have internal reference to replace & adjust the tests too
- domain is quite different so adapt the migration to it ( iam file instead of core file )

eric-intuitem (approved):
LGTM

eric-intuitem (approved):
LGTM

eric-intuitem (approved):

eric-intuitem (approved):

PR #1045, opened 2024-11-13
Use AllowUserOptions in AutocompleteSelect instead of the custom addOption() function

melinoix:
Refactoring by using allowuseroptions and not the custom addOption function


nas-tabchiche (changes requested):
On the right track. What remains to do before we merge it is to remove the `createFromSelection` prop and just pass `allowUserOptions` as a prop to the `AutocompleteSelect` component and forward it to `MultiSelect`.
This would simplify the component as it would remove a redundant if statement.

melinoix:
on the commit message : useless code not useful x)

nas-tabchiche (approved):
LGTM

PR #1059, opened 2024-11-19
Add priority to Applied controls

melinoix:
Added the field priority on applied controls with colored flags on the detailView and the modeltable.

### DIFF
--- a/frontend/src/lib/components/DetailView/DetailView.svelte
+++ b/frontend/src/lib/components/DetailView/DetailView.svelte
@@ -273,15 +273,19 @@
 												)?.urlModel
 											}/${value.id}`}
 											<a href={itemHref} class="anchor">{value.str}</a>
-										<!-- Shortcut before DetailView refactoring -->
-										{:else if value === "P1"}
-											<li class="fa-solid fa-flag text-red-500"></li> {m.p1()}
-										{:else if value === "P2"}
-											<li class="fa-solid fa-flag text-orange-500"></li> {m.p2()}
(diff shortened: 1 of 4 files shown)


backend/core/views.py
```diff
@@ -2494,7 +2502,7 @@ def export_mp_csv(request):
         "csf_function",
         "reference_control",
         "eta",
-        "effort",
+        "priority" "effort",
```
ab-smith: ,

backend/core/models.py
```diff
@@ -1310,6 +1310,21 @@ class Status(models.TextChoices):
...
+        ("P4", _("P4")),
+    ]
+
+    priority = models.CharField(
+        max_length=20,
+        choices=PRIORITY,
+        null=True,
+        blank=True,
+        verbose_name=_("Priority"),
+    )
```
nas-tabchiche: This is not compliant with the specs. Priority has to be an integer, "P1", "P2"... are only labels.

nas-tabchiche (changes requested):
`priority` has to be a number.
django's [PositiveSmallIntegerField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#positivesmallintegerfield) seems like a good pick for this.

eric-intuitem:
When I edit an applied control, the priority is not kept, it is set to --.

eric-intuitem (changes requested):
Priority shall be correctly initialised when entering the edit page.

ab-smith (approved):
functional ok

PR #1060, opened 2024-11-19
disk usage

ab-smith:
- **first step for du**
- **Add usage**


backend/core/views.py
```diff
@@ -2355,14 +2358,27 @@ def get_csrf_token(request):
     return Response({"csrfToken": csrf.get_token(request)})
 
 
+def get_disk_usage():
+    path = "./db/"
```
ab-smith: Maybe build the path per OS? using settings?

eric-intuitem (approved):
LGTM

PR #1070, opened 2024-11-21
Feat/improve mapping inference additional information

Mohamed-Hacene:
We cannot automatically transfer score and observation during mapping (in case of different score scale and contextual observation). However, we should have these informations in the requirement, to let the user choose if he wants to copy paste them.
Example:
image


eric-intuitem (changes requested):
The suggested values (score, annotation) should be copied anyway, it is easier to adjust them rather than copy/paste them. Btw when the relation is equal, most often we want to keep the score and annotation.

PR #1074, opened 2024-11-21
Ca 419 persist applied filters in url params

nas-tabchiche:


eric-intuitem (changes requested):
There is an issue when selecting a filter on object type for a library, going to another page, and coming back to the library: the object type disappears from the filter list. a page reload restores the filter list.
Moreover, when selecting a provider, the whole filter disappears after going to another page and coming back.

eric-intuitem (approved):
LGTM

PR #1078, opened 2024-11-22
add ref id to risk scenario

melinoix:
added ref id to risk scenario to replace hidden R.id (Kept on migration by putting each R.id in ref_id.


eric-intuitem (changes requested):
We should be able to see the ref_id field, and to edit it, like the name field. It is only pre-filled with R.x.

eric-intuitem (approved):
LGTM

PR #1085, opened 2024-11-25
Add community/enterprise update script

Mohamed-Hacene:
Aims to make the locale instance update easier, faster and safer


eric-intuitem (approved):
LGTM, assuming now that we are in the enterprise directory when launching the enterprise script, which is more natural.

PR #1087, opened 2024-11-27
Ca 616 serialize requirement node fields in requirement assessment

Mohamed-Hacene:


ab-smith (approved):
functional ok for 3rd parties

ab-smith (approved):
ok, but we need #1111 before releasing

PR #1118, opened 2024-12-02
Create base models for EBIOS RM

nas-tabchiche:


eric-intuitem (commented):
ROTO -> RoTo ?

eric-intuitem (commented):
Is the main entity included?

PR #1130, opened 2024-12-04
Workshop 1

Mohamed-Hacene:


backend/core/serializers.py
```diff
@@ -600,6 +601,14 @@ class ComplianceAssessmentWriteSerializer(BaseModelSerializer):
...
         allow_null=True,
     )
+    ebios_rm_studies = serializers.SlugRelatedField(
+        many=True,
+        slug_field="id",
+        queryset=EbiosRMStudy.objects.all(),
+        required=False,
+        allow_null=True,
+        write_only=True,
+    )
```
nas-tabchiche: EBIOS RM stuff leaking on `core` should be avoided unless absolutely necessary. Since this field would only be needed in EBIOS RM contexts, what about having a serializer inheriting from this one in the `ebios_rm` app, with the additional field?

nas-tabchiche: Nevermind, let's take this shortcut.

frontend/src/routes/(app)/(internal)/ebios-rm/[id=uuid]/workshop-one/base/+page.server.ts
Comment (nas-tabchiche) [inline frontend/src/routes/(app)/(internal)/ebios-rm/[id=uuid]/workshop-one/base/+page.server.ts:1]: This route should be called `baseline` instead of `base` to reflect the naming used by the ANSSI

nas-tabchiche (approved):
LGTM.

PR #1151, opened 2024-12-07
word export

ab-smith:
- **prep**
- **stable point**
- **Starting point with templating engine**
- **pass figures**
- **update gitignore**


eric-intuitem (approved):
LGTM

PR #1165, opened 2024-12-09
Hotfix for mail language

eric-intuitem:
We can set the LANGUAGE_CODE now.
The real fix is to ask for user default language in the form.


nas-tabchiche (approved):
LGTM

PR #1169, opened 2024-12-10
Fix language issue for mails

eric-intuitem:
Avoid skipping mo files


ab-smith (approved):

ab-smith:
https://forum.djangoproject.com/t/cease-and-desist-versioning-of-compiled-mo-files/23516

PR #1173, opened 2024-12-11
fix: make email comparison for SAML case insensitive

fastlorenzo:
Email comparison when doing SSO with SAML is currently case sensitive.
However, email addresses should be treated as case insensitive.

This PR makes sure that:
1) Email address of each user is saved as lowercase in the database
2) Email address received from IDP when doing SSO is converted to lowercase prior to comparison

### DIFF
--- a/backend/iam/adapter.py
+++ b/backend/iam/adapter.py
@@ -54,7 +54,7 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
     def pre_social_login(self, request, sociallogin):
         email_address = next(iter(sociallogin.account.extra_data.values()))[0]
         try:
-            user = User.objects.get(email=email_address)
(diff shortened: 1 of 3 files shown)


eric-intuitem:
LGTM. @nas-tabchiche ?

ab-smith:
we will merge it after the Ebios RM sprint and fix the test on our side before releasing

fastlorenzo:
> we will merge it after the Ebios RM sprint and fix the test on our side before releasing

Thanks, no worries, now that we know the current behavior we've fixed it in the IDP by forcing lowercase email conversion

nas-tabchiche:
Hi @fastlorenzo, we just merged #1498 which should make functional tests pass on this PR. I can't manually re-run the tests (maybe PR is too old). I'm closing this and re-opening to force CI to run.

PR #1176, opened 2024-12-11
translate emails for TPRM and pwd reset

eric-intuitem:


ab-smith:
how is this passed to the html templates of the mails?

eric-intuitem:
> how is this passed to the html templates of the mails?

This is managed automagically by Django templating system, and the templates were already internationalised.

ab-smith (approved):

PR #1180, opened 2024-12-12
Workshop 2 UX

Mohamed-Hacene:


ab-smith (approved):
functional ok but ux improvment are still needed

PR #1189, opened 2024-12-14
Update anssi-guide-hygiene-detail.xlsx

eric-intuitem:


ab-smith (approved):
functional ok,
strange though that the name of the applied control doesn't follow the selected language

eric-intuitem:
This will be fixed in #1191

PR #1196, opened 2024-12-16
Update AttackPath and OperationalScenario models

nas-tabchiche:
- Make the relationship between AttackPath and OperationalScenario
one-to-one
- Rename OperationalScenario.description to operating_modes_description
- Use NameDescriptionMixin to AttackPath


Mohamed-Hacene (approved):
0k

PR #1198, opened 2024-12-16
fix: Library uuid replaced by name in matrix view

Axxiar:
When displaying a risk matrix in the catalog, we can see the field library with a UUID. 
This should be replaced by the name of the library.


ab-smith:
I don't see it and I cannot reproduce the before and after

nas-tabchiche (approved):
LGTM

PR #1199, opened 2024-12-16
Order tabs following logical order

melinoix:
Order tabs following logical order, not alphabetical order anymore so we can choose in the code what is the most important tab to display first. 
If the order is not specified for an Urlmodel, it will just do by alphabetic order as before


nas-tabchiche (approved):
LGTM

PR #1200, opened 2024-12-16
Ca 695 automatically generate a scenario in the risk analysis for each operational scenario created

Mohamed-Hacene:


nas-tabchiche (approved):
LGTM, merging.

PR #1208, opened 2024-12-17
Temporary fix to get all the ids displayed in table mode

### LINKED ISSUE #1204: Table mode doesn't show controls identifiers anymore
**Describe the bug**
Since I believe the change introduced in https://github.com/intuitem/ciso-assistant-community/pull/1190 , we cannot see anymore the identifiers of the controls when using table mode.
This was really useful to be able to navigate through the different controls.

**To Reproduce**
1. Go to an assessment
2. Click on Table Mode

**Expected behavior**
I would expect to see the different identifiers of the controls (as it was shown before, but maybe change the way it is displayed for better readability).

**Screenshots**
Example below with CyFun (CCB) audit:
image

The structure with the identifiers that should be displayed in table mode:
image

 - Device: All
 - OS: All
 - Browser All
 - Version 1.9.29

**Additional context**
N/A

ab-smith:
should fix #1204


eric-intuitem (approved):
LGTM

PR #1214, opened 2024-12-20
Add null checks in table filters

nas-tabchiche:
This fixes a bug when user navigates from a project detail view to its
parent folder.


Mohamed-Hacene (approved):
0k

Issue #1135, opened 2024-12-05
Unable to use CA with a Helm deployment


dr-lux:
**Describe the bug**

After a tweaked (more on that in `How to reproduce`) deployment of CA through Helm, I was able to log in. However in many main page content (like `/analytics`, `/projects`, ...)  I got `Internal Error 500`.

From my logs I got this error `django.db.utils.OperationalError: no such column: core_project.internal_reference`

Here's the full logs of the `ciso-assistant-backend` in my Pod:

```text
[2m2024-12-05T10:59:41.710137Z[0m [[32m[1minfo     [0m] [1mrequest_started               [0m [[0m[1m[34mdjango_structlog.middlewares.request[0m][0m [36mciso_assistant_url[0m=[35mhttps://<DOMAIN>[0m [36mip[0m=[35m127.0.0.1[0m [36mrequest[0m=[35mGET /api/iam/current-user/[0m [36mrequest_id[0m=[35mfeb2c693-c32b-410f-963d-c0d9e63a0bab[0m [36muser_agent[0m=[35mnode[0m [36muser_id[0m=[35mNone[0m
[2m2024-12-05T10:59:41.747670Z[0m [[32m[1minfo     [0m] [1mrequest_finished              [0m [[0m[1m[34mdjango_structlog.middlewares.request[0m][0m [36mciso_assistant_url[0m=[35mhttps://<DOMAIN>[0m [36mcode[0m=[35m200[0m [36mip[0m=[35m127.0.0.1[0m [36mrequest[0m=[35mGET /api/iam/current-user/[0m [36mrequest_id[0m=[35mfeb2c693-c32b-410f-963d-c0d9e63a0bab[0m [36muser_id[0m=[35m3e49b149-b4ce-45d1-bca9-eea958c2f33a[0m
[2m2024-12-05T10:59:41.751427Z[0m [[32m[1minfo     [0m] [1mrequest_started               [0m [[0m[1m[34mdjango_structlog.middlewares.request[0m][0m [36mciso_assistant_url[0m=[35mhttps://<DOMAIN>[0m [36mip[0m=[35m127.0.0.1[0m [36mrequest[0m=[35mGET /api/projects/[0m [36mrequest_id[0m=[35md8a8e55c-1851-4c7b-aa89-9a46c4447594[0m [36muser_agent[0m=[35mnode[0m [36muser_id[0m=[35mNone[0m
[...]

Issue #1123, opened 2024-12-04
Helm: Prevent the Use of Let's Encrypt for Certificate Generation


siddjellali:
## Description:
I’m experiencing an issue with using Let's Encrypt as the certificate provider in my setup. I use Traefik as a reverse proxy to manage TLS certificates, but I’m getting a TLS error because Caddy starts without a valid certificate when deployed using the Helm Chart.

In my Docker Compose setup, a tls internal option was added to Caddy, which resolved the issue. Is it possible to add a similar tls internal option to the Helm Chart to ensure that Caddy can start properly without relying on Let's Encrypt? This would also prevent conflicts with Traefik.

## Steps to Reproduce:

- Deploy the application using the Helm Chart with Caddy.
- Configure Traefik as the reverse proxy to handle TLS certificates.
- Observe that Caddy starts without a valid certificate, resulting in a TLS error.

## Expected Result:
Caddy should use an internal TLS configuration (tls internal) to avoid the error and integrate correctly with Traefik.

## Environment:

- Reverse Proxy: Traefik
- Deployment: Helm Chart
- Specific Configurations: TLS certificate management via Traefik

## Proposed Solutions:

1. Add an option to the Helm Chart to enable a `tls internal` setting for Caddy, similar to the Docker Compose configuration.
2. Add an option to disable Caddy completely and expose the frontend service directly, allowing Traefik to manage reverse proxying and TLS certificates without conflicts.

siddjellali:
Hi, 

It's not exactly the same issue.
Can we propose something to disable caddy ?

ab-smith:
Let's see if the work on #1470 can help with this.

ab-smith:
left-over, TLS via existingSecret/certificateSecret is supported for a while now.

PR #1216, opened 2024-12-20
Add more flexibility in Helm chart

Nathanael-Mtd:
To make CISO Assistant more flexible for Kubernetes users, I made some (breaking) changes on the Helm Chart : 

- Add Ingress config in values (annotations and ingressClassName)
- Enable tls on Ingress
- Caddy removal (useless if ingress.tls used)

I don't think it will be a big pain to migrate chart with these changes.

It will requires using Cert-Manager for automatic letsencrypt/ACME/PKI certificate generation, or just create secret manually alongside the chart deploy.

I will be happy to add custom TLS secret creation from values file if it can help some users to avoid doing manual things.

That PR can probably fix these issues : #1123 #1135

### DIFF
--- a/charts/ciso-assistant/templates/ingress.yaml
+++ b/charts/ciso-assistant/templates/ingress.yaml
@@ -1,16 +1,19 @@
+{{- if .Values.ingress.enabled }}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
+  {{- with .Values.ingress.annotations }}
   annotations:
-    nginx.ingress.kubernetes.io/backend-protocol: HTTPS
-    nginx.ingress.kubernetes.io/ssl-passthrough: "true"
-    nginx.ingress.kubernetes.io/ssl-redirect: "true"
+    {{- range $key, $value := . }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
+  {{- end }}
   name: ciso-assistant-{{ .Values.clientName }}
   labels:
     client: {{ .Values.clientName }}
     {{- include "common.labels" . | nindent 4 }}
 spec:
-  ingressClassName: nginx
+  ingressClassName: {{ .Values.ingress.ingressClassName }}
   rules:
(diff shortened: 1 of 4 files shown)


Nathanael-Mtd:

ab-smith:
This is awesome @Nathanael-Mtd  thank you for the contribution!
I was thinking more about having caddy as optional rather than just removing it

Nathanael-Mtd:
Maybe, but I don't think chaining multiple reverse proxy, or adding reverse proxy inside pods are good ideas overall, specifically if there are no added value.

Also I think that frontend and backend should be separated in two Pods, and avoiding using statefulset, because in case of replicas > 1, we will have multiple distincts databases running.
Better using Deployments, if we want to scale backend, we can use a volume in ReadWriteMany mode to share SQLite DB across all backend pods.

If you're ok with that, I can make these changes, and also adding an option to keep Caddy as a separate Deployment if you think that's always useful.

ab-smith:
I tell you what: let’s start a different chart for these options: deployment-based, postgres instead of sqlite (to support multiple clients, the backend in this case) and we drop caddy in this flavor. What do you think?

Nathanael-Mtd:
Seems fine for me, we can keep an option to use SQLite for simple deployments.
How do you want to proceed on that "flavor" ? Different name/folder/repo ? And about the current chart, you want to keep and maintain it, or deprecate it later ?

ab-smith:
Yup, we can go for a different folder in this repo, ciso-assistant-alt or something like that, and I’ll handle the distribution afterwards.
we will keep both and see how the community responds and adjust accordingly.
We can try your sqlite idea with ReadWriteMany, just that I’ve had bad experiences and mixed feedback when people twisting it for concurrency but could be a good learning opportunity.
P.S : thank you again and feel free to jump in to the discord server.

Nathanael-Mtd:
Fine, I will create an another PR for the new chart if it's ok for you.

ReadWriteMany was just an idea to avoid impossible scalability in case of Statefulset, but to be fair, managing the backend scalability with database is generally a pain 😅

PR #1218, opened 2024-12-21
Update iso27001-2022.yaml and rename ITAR lib

eric-intuitem:


ab-smith (changes requested):
6.1 should not be assessable

eric-intuitem:
Indeed, fixed.

ab-smith (approved):

ab-smith (approved):

PR #1223, opened 2024-12-21
ebios-rm small fixes

eric-intuitem:
error in the matrix name translation
Remove default matrix choice


ab-smith (approved):
functional ok

PR #1232, opened 2024-12-23
hotfix: applied controls creation from risk scenario edit view

Mohamed-Hacene:


backend/core/views.py
```diff
@@ -1307,7 +1307,9 @@ class RiskScenarioViewSet(BaseModelViewSet):
     ordering_fields = ordering
 
     def _perform_write(self, serializer):
-        if not serializer.validated_data.get("ref_id"):
```
Mohamed-Hacene, line 1310: When we create an applied control from a scenario, the validated data only contains the applied control created, not the risk assessmentq so an error occured

backend/core/views.py
Mohamed-Hacene, line 1: When we create an applied control from a scenario, the validated data only contains the applied control created, not the risk assessment so an error occured

nas-tabchiche (approved):
LGTM, merging

PR #1234, opened 2024-12-23
Leverage RBAC for global settings

nas-tabchiche:
When writing a custom `get_object` method in a viewset, we must perform call `self.check_object_permissions(request, obj)`. Otherwise, `RBACPermissions.has_object_permission` will **not** be called.


Mohamed-Hacene (approved):
0k

PR #1261, opened 2024-12-26
Add eco radar ebios study

Mohamed-Hacene:


ab-smith (changes requested):
let's stick to the colors I've put initially as they're purposefuly aligned with the official document, even if they can be improved

ab-smith (approved):

PR #1266, opened 2024-12-27
Refactor and update CI deps

ab-smith:


Mohamed-Hacene (commented):
Error in the workflows using `env`:
`The workflow is not valid. .github/workflows/functional-tests.yml (Line: 25, Col: 14): Unrecognized named-value: 'env'. Located at position 1 within expression: env.UBUNTU_VERSION`
It seems to be a well known issue for some time now: https://github.com/actions/runner/issues/2372

Mohamed-Hacene (approved):
0k

PR #1291, opened 2024-12-31
Create YAML for DORA Spanish translation provided by thidalgosalvador

eric-intuitem:


ab-smith (approved):
ok, but we should review if the first parts are really not assessable and consider taking them out to improve readability and usability (eg .in the case of table mode)

PR #1297, opened 2025-01-02
Support question-only requirements

monsieurswag:


nas-tabchiche (changes requested):
Okay for compliance assessments. However, when opening a framework's tree view in /frameworks or /libraries, question-only requirements still appear as blank text.
To replicate, import the ITAR Compliance Program Guidelines, navigate to its detail view and open element 7C.

nas-tabchiche (approved):
Good to go now, to merge once CI is green.

PR #1298, opened 2025-01-02
Fix observation and score copying during the mapping process

monsieurswag:


backend/core/views.py
```diff
@@ -2491,6 +2497,14 @@ def perform_create(self, serializer):
                         "source_requirement_assessment"
                     ]["id"]
                 )
+                if same_minmax_score and target_framework == source_framework:
```
Mohamed-Hacene, line 2500: No need to check if it is the same framework here, the score scale is sufficient

Mohamed-Hacene (approved):
0k

PR #1302, opened 2025-01-02
Prevent assigning parent assets to primary assets

nas-tabchiche:
Users used to be able to add parent assets to a primary asset by setting its type to support, assigning parent assets, and setting its type back to primary.
This PR makes it so that it is no longer possible, and also hides the parent assets field if the asset being written is primary.


ab-smith (approved):
functional ok

PR #1304, opened 2025-01-03
Ca 746 clicking on the box of the filter closes it it should only close when clicking outside the box

monsieurswag:


Mohamed-Hacene (approved):
0k

PR #1318, opened 2025-01-06
Include search params in edit next URL when coming from list

nas-tabchiche:
I purposefully did not include next urls coming from detail views as it requires additional testing.
Including url search params in next urls coming from list views is safe however, as search params are only used for filtering in list views.


Mohamed-Hacene (approved):
0k

PR #1323, opened 2025-01-07
Set framework page title to framework name

nas-tabchiche:


Mohamed-Hacene (approved):
0k, can be merged after fixing the conflict

Mohamed-Hacene (approved):
0k

Issue #1342, opened 2025-01-09
The value 4 should not be allowed in as a value in security_objectives


gbyx3:
When creating assets via the api (POST /api/assets/) the `security_objectives` values is not properly validated.
It checks for an integer but not for a max value of >=3.
This causes an internal server error and the asset page fails to load.
This issue is similar to #1329, the fix for this issue does not solve this.

**To Reproduce**
1. Make a POST request to /api/assets/
2. Payload:
```json
{
  "is_published": true,
  "name": "Some name",
  "type": "PR",
  "folder": "<CHANGE ME>",
  "security_objectives": {
    "objectives": {
      "confidentiality": {
        "value": 4,
        "is_enabled": true
      }
    }
  }
}
```
3. Refresh the Assets page under Organization

**Expected behavior**
Block the request with "4 is not a valid integer for value"...

**Screenshots**

**Environment (please complete the following information):**
 - Device: Mac and PC
 - OS: OsX, Debian
 - Browser Firefox
 - Version 133.0.3

**Additional context**
See thread "Asset import with API" in the "tech-chat" channel on discord

gbyx3:
Should be solved in #1343  and #1344

PR #1343, opened 2025-01-09
The value 4 should not be allowed in the security_objectives_display …

gbyx3:
…array

This does not fix input validation, it does however prevent the application from breaking if a invalid value of 4 is entered. Similarly to how 5 and above works today.


gbyx3:
This solves #1342 from breaking.
I does NOT solve the input validation.

More testing and validation from the core team would be advised!

gbyx3:

nas-tabchiche (approved):
Looks good to me. Thank you for your contribution

PR #1344, opened 2025-01-09
Cap asset security objective max value to 3

nas-tabchiche:
For now, we only manage 4-value security objective scales.


eric-intuitem (approved):
LGTM

PR #1347, opened 2025-01-10
Hotfix: inlang build

nas-tabchiche:
https://github.com/opral/inlang-paraglide-js/issues/304#issuecomment-2582087107


Mohamed-Hacene (approved):
0k

PR #1354, opened 2025-01-12
Specify database dump format

nas-tabchiche:
This PR lays the groundwork for domain import/export.
The bulk of it will be located under the `serdes` django app, as it will be the basis of the upcoming work on general database dumps, moving away from stock django's `dumpdata` and `loaddata` management commands.


Mohamed-Hacene (approved):
0k

PR #1361, opened 2025-01-15
docs: domain import/export specification

eric-intuitem:
- add domain import/export
- add draft for asset compliance

## Summary by CodeRabbit

- **New Features**
	- Added support for domain import/export functionality in the PRO version
	- Introduced asset compliance tracking with new compliance indicators and observations

- **Documentation**
	- Updated data model documentation with detailed specifications for new features
	- Added descriptions for new data entities related to asset compliance

### DIFF
--- a/documentation/architecture/data-model.md
+++ b/documentation/architecture/data-model.md
@@ -1442,3 +1442,47 @@ erDiagram
 - EBIOS-RM objects are defined within a dedicated Django "application" ebios_rm.
 - There is no object for "strategic scenarios", as they result directly from attack paths and corresponding feared event (which is the title of the strategic scenario).
 - the current and residual "criticity" are calculated on stakeholders, so they are not seen as fields.
+
+## Domain import/export
+
+### MVP
+
+- It is possible to export a domain, and reimport it on another instance.
+- The imported domain is created as a new child domain of the global domain.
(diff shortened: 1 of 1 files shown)


documentation/architecture/data-model.md
```diff
@@ -1442,3 +1442,48 @@ erDiagram
...
+## Domain import/export
+
+### MVP
+
+- It is possible to export a domain, and reimport it on another instance.
+- The imported domain is created as a new child domain of the global domain.
+- The name of the domain is not exported. The name of the imported domain is selected at import. It shall be unique.
+- Subdomains are not exported, only the domain itself.
+- Objects with an URN that are referenced shall be available on the target. If not, an error message is displayed to point to the missing library.
+- Objects without URN that are referenced in the domain are created in the target domain, even if they come from an upper domain.
```
nas-tabchiche: Wouldn't it be more straightforward to look at the object's `library` foreign key relationship rather than its URN for those two points?

Mohamed-Hacene: I agree with you; technically, it should be equivalent since the URN field is populated only when the library field is filled. However, checking the library field is more consistent in this case, as the URN is merely a unique identifier.

Mohamed-Hacene (approved):
0k

PR #1364, opened 2025-01-16
asset remove bv field

ab-smith:
- **remove bv from asset creation form**
- **remove from schema and tests**


eric-intuitem (approved):
LGTM

PR #1368, opened 2025-01-16
fix broken pdf when exporting risk analysis and the associated action plan

ab-smith:
- **split matrix into two pages to avoid breaks**
- **clean up**
- **this looks better**
- **no more breaks**


eric-intuitem (approved):
LGTM

PR #1376, opened 2025-01-18
export/import domain

eric-intuitem:


backend/core/models.py
```diff
@@ -1,6 +1,7 @@
 import json
 import os
 import re
+import hashlib
```
ab-smith, line 4: improvement: maybe we can align the import style between the files

enterprise/frontend/src/routes/(app)/(internal)/folders/[id=uuid]/export/+server.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const GET: RequestHandler = async ({ fetch, params }) => {
+	const endpoint = `${BASE_API_URL}/folders/${params.id}/export/`;
+
+	const res = await fetch(endpoint);
+	if (!res.ok) {
+		error(400, 'Error fetching the dump file');
+	}
+
+	const fileName = `ciso-assistant-domain-export-${new Date().toISOString()}.bak`;
```
Comment (ab-smith) [inline enterprise/frontend/src/routes/(app)/(internal)/folders/[id=uuid]/export/+server.ts:14]: improve: slugified domain name on the filename

ab-smith (approved):
functional ok with small improvement to add on a different PR

PR #1380, opened 2025-01-18
Translate CCB in French and define score definition for documentation

eric-intuitem:


ab-smith:
translation is ok, but not the documentation score labels

ab-smith (approved):
Functional ok

PR #1394, opened 2025-01-22
Feat(lib) d3fend

eric-intuitem:
Add Mitre D3FEND
Add default value for library_publication_date
ruff format on tools


ab-smith:
we should not use this kind of syntax for the branches:

image

ab-smith (approved):

PR #1401, opened 2025-01-22
fix: make score read only in entity assessment questionnaire in item view

melinoix:
make score read only in entity assessment questionnaire in item view and show both scores if there is the documentation score or just a classic score if there is only the implementation score

## Summary by CodeRabbit

- **Localization**
	- Added new score-related translation keys for English and French languages.
	- Introduced localized labels for documentation and implementation scores.
	- Added indication for absence of evidences in both languages.
	- Improved messaging for missing libraries in the import section.

- **User Interface**
	- Updated scoring component rendering logic in the compliance assessment form.
	- Implemented conditional display of score based on assessment state.
	- Enhanced visual representation of scores with new components, including checkboxes and progress indicators.


eric-intuitem (changes requested):
"Score: null" should be shown as "Score: -"
"No answer" and "No observation" are not localised.
When there are no observation, we should have "No observation" directly (as it is done well for "no answer").
Is is worth to have the observation folded? Why not show it when it exists?
"Evidence" not localised.

Mohamed-Hacene (changes requested):
- [ ] We cannot add an evidence from the classic table mode anymore
image
- [ ] It could be better to align the score on left (put documentation score under implementation in this case)
image
- [ ] Put `--` if not scored instead of 0 (which means it's scored)
image

Mohamed-Hacene (approved):
0k

PR #1443, opened 2025-01-28
feat:  add a progress field on applied controls

melinoix:
This is a new field. It is an integer between 0 to 100 (percentage). Defaults to 0. Cannot be Null.

When the status is set to “active”, this field is set to 100.

## Release Notes

- **New Features**
	- Added a progress tracking field for applied controls.
	- Introduced ability to filter applied controls by progress.

- **Internationalization**
	- Added "Progress" label in English and French translations.

- **User Interface**
	- Enhanced applied control policy form with a progress score component.
	- Updated detail view to include the progress field.

- **Backend**
	- Implemented automatic progress field updates based on control status.


ab-smith:
@melinoix  the color scheme should be inverted: going from red (0) to green (100)

ab-smith (approved):

PR #1474, opened 2025-02-05
Serialize the last migration name for each app

nas-tabchiche:


nas-tabchiche:
Closed in favor of #1475.

PR #1484, opened 2025-02-08
feat: allow helm chart secretes to be provided via external secret

nparfait:
New Features
-  values for django secret and smtp password can be provided via an external secret.

## Summary by CodeRabbit

- **Chores**
	- Enhanced secure configuration management by conditionally using pre-existing credentials.
	- Reduced reliance on hardcoded sensitive data for backend services and email configurations.
	- Improved configuration clarity to reinforce overall security and flexibility.
	- Added new fields in the configuration for existing secrets related to SMTP and Django.

### DIFF
--- a/charts/ciso-assistant-next/templates/backend/deployment.yaml
+++ b/charts/ciso-assistant-next/templates/backend/deployment.yaml
@@ -103,11 +103,11 @@ spec:
           - name: EMAIL_HOST_USER
             value: {{ .Values.backend.config.smtp.primary.username | quote }}
           {{- end }}
-          {{- if .Values.backend.config.smtp.primary.existingSecret }}
+          {{- if .Values.backend.config.smtp.existingSecret }}
           - name: EMAIL_HOST_PASSWORD
             valueFrom:
               secretKeyRef:
-                name: {{ .Values.backend.config.smtp.primary.existingSecret }}
+                name: {{ .Values.backend.config.smtp.existingSecret }}
                 key: email-primary-password
           {{- else if .Values.backend.config.smtp.primary.password }}
           - name: EMAIL_HOST_PASSWORD
@@ -116,6 +116,7 @@ spec:
                 name: {{ include "ciso-assistant.fullname" . }}-backend
                 key: email-primary-password
           {{- end }}
+          {{- if .Values.backend.config.smtp.rescue.enabled }}          -
           - name: EMAIL_HOST_RESCUE
             value: {{ .Values.backend.config.smtp.rescue.host | quote }}
           - name: EMAIL_PORT_RESCUE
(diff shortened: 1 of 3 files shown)


nparfait:

Nathanael-Mtd:
Maybe we should make secrets config more "clear" by creating a backend-smtp secret and backend-django one.
Like that we can declare only one time the existingSecret for SMTP and one for Django.
It will make the conditions on deploy more simple to define. 

Also, can be a good idea to send all smtp parameters into the same secret, not only passwords. 

What do you think about it ?

nparfait:
> Maybe we should make secrets config more "clear" by creating a backend-smtp secret and backend-django one. Like that we can declare only one time the existingSecret for SMTP and one for Django. It will make the conditions on deploy more simple to define.
> 
> Also, can be a good idea to send all smtp parameters into the same secret, not only passwords.
> 
> What do you think about it ?

Yes, could use a single existingSecret field for SMTP, however there may not be any rescue config? so would have to include and enabled field/flag or similar as there may not be a rescue secret key. Thoughts?
 
I don't think the other fields should be moved as they don't need to be secrets and would impact visibility.
The way we have secrets deployed is using a generic external secrets chart together with external secrets operator integrated with AWS Secrets. Adding the host/port config to the secrets I don't think is required/beneficial.

Nathanael-Mtd:
About the rescue part, yes we should have an enabled option for conditional , but if I'm not wrong, the backend fail if there are no rescue host configured. If you can check and confirm that.

To be honest don't really understand the need of rescue smtp server, that's the first time I saw that implementation inside that type of application. 🤷

Ok I see, let's keep only passwords in secrets, but one for smtp, one for Django key.

eric-intuitem:
Agreed that the rescue smtp server is overkill, we had an unreliable one initially, thus this choice. But EMAIL_HOST_RESCUE is optional, you can simply ignore it in the Helm chart.

nparfait:
> About the rescue part, yes we should have an enabled option for conditional , but if I'm not wrong, the backend fail if there are no rescue host configured. If you can check and confirm that.
> 
> To be honest don't really understand the need of rescue smtp server, that's the first time I saw that implementation inside that type of application. 🤷
> 
> Ok I see, let's keep only passwords in secrets, but one for smtp, one for Django key.

Added an enabled flag to disable rescue smtp server, and moved the existing secret key up 1 level so it applies to both primary/rescue.

Nathanael-Mtd:
It will be good to me if the django Secret will be separated from the SMTP one also on integrated config, to keep the same logic as external one. 
Also, the value name of the secret is wrong, needs to remove the "Key" in name.

To be coherent, if we make separation of SMTP and Django for external Secrets, we need to do the same for the integrated ones.

About the rescue SMTP, I will probably remove that config on my future PR, but if you want you can remove it now to simplify the chart.

ab-smith (approved):

PR #1490, opened 2025-02-10
ci: use !cancelled() rather than always()

nas-tabchiche:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always


nas-tabchiche:
@coderabbitai full review

Mohamed-Hacene (approved):
0k

PR #1512, opened 2025-02-18
fix: mapping documentation score transfer when cloning

Mohamed-Hacene:
## Summary by CodeRabbit

- **New Features**
	- Introduced a documentation quality metric into the scoring system for compliance assessments.
	- Enhanced assessment creation logic to automatically factor in the documentation score when a relevant baseline is provided.
	- Improved control flow for handling different frameworks during compliance assessment creation.

### DIFF
--- a/backend/core/models.py
+++ b/backend/core/models.py
@@ -2927,11 +2927,7 @@ class ComplianceAssessment(Assessment):
                     assessment.result = baseline_assessment.result
                     assessment.status = baseline_assessment.status
                     assessment.score = baseline_assessment.score
-                    assessment.documentation_score = (
-                        baseline_assessment.documentation_score
-                        if baseline_assessment.documentation_score
-                        else None
-                    )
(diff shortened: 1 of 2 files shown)


backend/core/models.py
```diff
@@ -2927,6 +2927,11 @@ def create_requirement_assessments(
                     assessment.result = baseline_assessment.result
                     assessment.status = baseline_assessment.status
                     assessment.score = baseline_assessment.score
+                    assessment.documentation_score = (
+                        baseline_assessment.documentation_score
+                        if baseline_assessment.documentation_score
```
nas-tabchiche: This would set the documentation_score to None if it equals 0 in the baseline. What about `assessment.documentation_score = baseline_assessment.documentation_score`? This is what we already do with `score`.

Mohamed-Hacene: You're totally right!

eric-intuitem (approved):
LGTM

PR #1528, opened 2025-02-21
fix: store risk matrix json definition as a dict, not a str

nas-tabchiche:
This is required to be able to query inside a `RiskMatrix` object's `json_definition` field.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the processing of risk data by removing unnecessary JSON parsing, resulting in more efficient data handling and improved performance.
	- Simplified the retrieval of risk matrices and risk scoring by directly accessing usable data formats.
	- Updated the import process to store risk matrix data as native Python dictionaries instead of JSON strings.
	- Introduced a migration to update existing records to store JSON data as dictionaries instead of strings.


eric-intuitem (approved):
LGTM

PR #1536, opened 2025-02-24
Fix: Avoid multiple loads of demo data + feat: loading animation during import of demo data 

melinoix:
The demo data domain is created always with the same name. If the domain already exists a popup will say that demo data is already existing.

Additionally, during loading of data, an animation is shown.

## Summary by CodeRabbit

- **New Features**
  - Introduced a loading spinner and enhanced error notifications during demo data import for clearer user feedback.
  - Added new localized messages informing users when demo data has already been imported (available in both English and French).

- **Refactor**
  - Streamlined the demo data import process by standardizing the identifier, ensuring consistency across imports.

- **Bug Fixes**
  - Improved error handling for demo data loading, providing specific messages for different error responses.


frontend/src/lib/components/SideBar/SideBar.svelte
```diff
@@ -218,20 +220,31 @@
...
+		flash.set({ type: 'success', message: m.successfullyImportedFolder() });
+
 		await goto('/folders', {
 			crumbs: breadcrumbs,
 			label: m.domains(),
 			breadcrumbAction: 'replace'
 		});
+
 		invalidateAll();
+		loading.set(false);
```
nas-tabchiche: Nit: a writable store seems overkill here. A simple boolean is enough to serve this purpose.

PR #1537, opened 2025-02-24
fix: Remove child assets from parent assets field

melinoix:
Removed child assets from parent assets field, only primary assets can be chosen in supporting assets

## Summary by CodeRabbit

- **New Features**
	- Enhanced asset filtering to allow users to refine search results by excluding subordinate assets.
	- Improved asset selection in forms with an updated autocomplete that includes detailed URL parameters to filter available options.


frontend/src/lib/components/Forms/ModelForm/AssetForm.svelte
```diff
@@ -110,7 +110,7 @@
 	hidden={data.type === 'PR'}
 	multiple
 	{form}
-	optionsEndpoint="assets"
+	optionsEndpoint="assets?type=PR"
```
Mohamed-Hacene: We don't want only `Primary assets` as parents. A `Support asset` can also be a parent. Here we want to display candidates eligible for "kinship". To do this, the proposed assets must not be descendants.

eric-intuitem (approved):
LGTM

PR #1567, opened 2025-02-27
feat: add children assets in assets detail view

melinoix:
Adding a children assets in assets detail view can be useful to see al the descendance and so the importance of an asset .
If 1 is the parent of 2 and 2 is the parent of 3, 1 will have 2 and 3 as children assets.

## Summary by CodeRabbit

- **New Features**
  - Assets now display associated child assets for improved hierarchical navigation.
  - Updated multilingual support with new labels for child assets in English and French.
  - Enhanced asset relationship mapping across the platform.

- **Enhancements**
  - Added a property to directly access child assets in the Asset class.
  - Included child asset identifiers in the AssetReadSerializer for better data representation.
  - Enhanced the DetailView component to render child assets when available.
  - Introduced a utility function for string truncation in the DetailView component.

### DIFF
--- a/backend/core/models.py
+++ b/backend/core/models.py
@@ -1653,6 +1653,12 @@ class Asset(
             sub_children.update(child.get_descendants())
         return sub_children
 
+    @property
+    def children_assets(self):
+        descendants = self.get_descendants()
+        descendant_ids = [d.id for d in descendants]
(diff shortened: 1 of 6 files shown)


eric-intuitem (approved):
LGTM

PR #1582, opened 2025-02-28
feat: warn the user during domain export if an object is out of scope

melinoix:
During domain export, we sometimes need to take objects outside of the domain which is exported (Global, or sub-domains mostly). This PR warn the user that some objects out of this domain will also be exported but only when it’s the case.

## Summary by CodeRabbit

- **New Features**
	- When exporting folder data, users are now prompted with a confirmation modal if items beyond the designated scope are detected. The modal displays warning details and a list of these objects before proceeding.
	- Enhanced user messaging—with updates in both English and French—provides clearer guidance and improved feedback during the export process.
	- A new method to check for out-of-scope objects related to folders has been implemented, improving data management.
	- A new component for displaying confirmation modals has been introduced, enhancing user interaction during data exports.
	- Export forms now require user confirmation when out-of-scope objects are present, preventing accidental exports.
	- Folder detail views include checks for out-of-scope objects, ensuring users are informed before exporting.


Mohamed-Hacene (changes requested):
Nice feature, however folders detailed page seems really slowed compared to main. Need perf investigation.

Mohamed-Hacene:
Closing for now. To be reopened after merging main with svelte5 and solve perf issues.

PR #1591, opened 2025-03-02
fix: unique constraint violation for mapping_urn

### LINKED ISSUE #1590: prepare_mapping.py fails mapping_urn unique constraint when importing libraries with same source framework
When attempting to import two mapping libraries from the same source framework (but with different target frameworks), the import fails for the second library due to a unique constraint violation on `mapping_urn`.

Upon reviewing the script, it appears that `source_ref_id` is used instead of `ref_id` to get the urn, causing it to be based solely on the source framework.

https://github.com/intuitem/ciso-assistant-community/blob/92e4002c8c52e8d1a55c1d520365568dc029c4f4/tools/prepare_mapping.py#L54

JosephSaliba01:
fix #1590
Uses the `ref_id` variable, which combines both the source and target framework, to get the urn.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the Excel export so that mapping reference identifiers now combine key details for improved clarity.

### DIFF
--- a/tools/prepare_mapping.py
+++ b/tools/prepare_mapping.py
@@ -51,7 +51,7 @@ ws.append(["library_provider", packager])
 ws.append(["library_packager", packager])
 ws.append(["library_dependencies", f"{source_library_urn}, {target_library_urn}"])
 ws.append(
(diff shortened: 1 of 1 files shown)


JosephSaliba01:

eric-intuitem (approved):
LGTM.

PR #1603, opened 2025-03-04
feat: make gunicorn binding port parametric

srfaber:
Since Cloud Run does work with Port 8080. Cloud Run assigns a dynamic port via $PORT. ${PORT:-8000} → Uses the PORT environment variable if set; otherwise, defaults to 8000.

## Summary by CodeRabbit

- **Chores**
  - Improved server configuration to dynamically assign its binding port via environment settings, ensuring enhanced deployment flexibility while retaining a reliable default.


srfaber:

ab-smith:
Thank you @srfaber !
it’s better to have the same structure as the previous parameters for readability.

srfaber:
@ab-smith you're right. Let me change.

ab-smith:
Awesome! almost there: just prefix it with GUNICORN_ to avoid confusion with other possible env variables :)

ab-smith (approved):

PR #1676, opened 2025-03-14
fix(helm): add ingress.tls.enabled flag

smonfort-cacd2:
Follow-up MR for #1663 

Do not use the global TLS activation flag to control the TLS attribute on the ingress ressource, so that the pod can be deployed behind a reverse proxy that performs the TLS termination out of the Kubernetes cluster.

In this situation, global.tls is set to true (for HTTPS scheme in the URL) but TLS should not be configured in the ingress.
Ingress TLS configuration can still be activated as previously by switching ingress.tls.enabled to true.

To improve readability, all ingress TLS related attributes have been moved to ingress.tls block.

## Summary by CodeRabbit

- **Documentation**
  - Updated version and application badges.
  - Revised configuration descriptions to clarify TLS settings for URLs versus ingress.

- **Refactor**
  - Streamlined TLS configuration by migrating ingress-specific settings under a dedicated section.
  - Updated logic for TLS activation and URL scheme determination for improved flexibility.


smonfort-cacd2:
Thanks @Nathanael-Mtd for your feedback and major contributions on the Helm chart. 
My point is that the chart should allow https URL without requiring tls on the ingress, especially when TLS termination is not performed by the ingress controller but an external reverse proxy/load balancer (such as AWS ALB).
That's why I proposed to decoupled TLS activation on the ingress and https scheme for the application external URL.
Please let me know if the use case requires more detailed explanations, even if I admit it may require to set the configuration twice.

Nathanael-Mtd:
I'm fine with your changes, that's better, but it will be nice to add an OR condition on ciso-assistant.url template to handle cases when users forgot to set global.tls to true but set ingress.tls.enable to true. In that precise case, with your changes, the URL will use http scheme instead of https.

You can do that by changing end of templates/_helpers.tpl on line 73 : 
```diff
{{/*
Define complete url based on scheme and domain
*/}}
{{- define "ciso-assistant.url" -}}
- {{- $scheme := ternary "https" "http" .Values.global.tls -}}
+ {{- $scheme := ternary "https" "http" (or .Values.global.tls .Values.ingress.tls.enable) -}}
{{- printf "%s://%s" $scheme .Values.global.domain -}}
{{- end -}}

```

smonfort-cacd2:
You're right, changed applied, thanks.

Nathanael-Mtd:
Just saw a typo, on my diff example I've used ingress.tls.enable instead of ingress.tls.enabled.
I think helm will throw an error. Don't forget to run `helm lint` before pushing, because right now there are no helm lint checks on GitLab actions

charts/ciso-assistant-next/templates/_helpers.tpl
```diff
@@ -70,6 +70,6 @@ app.kubernetes.io/component: {{ .component }}
 Define complete url based on scheme and domain
 */}}
 {{- define "ciso-assistant.url" -}}
-{{- $scheme := ternary "https" "http" .Values.global.tls -}}
+{{- $scheme := ternary "https" "http" (or .Values.global.tls .Values.ingress.tls.enable) -}}
```
ab-smith: `enabled`

ab-smith (approved):

PR #1724, opened 2025-03-20
fix: eager i18n function call in unsafeTranslate function

nas-tabchiche:
## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced the translation feature by tightening validation, ensuring that only valid translation functions are executed. This improvement increases reliability and prevents errors in language handling for a smoother user experience.


ab-smith (approved):
functional ok

PR #1737, opened 2025-03-21
feat: add abilities to dynamically add env vars from Values.yaml in H…

Siryu6:
To simplify integration from external database this will allows to dynamically override EnvVars from ConfigMap and SecretRef.


ab-smith:
Welcome @Siryu6 and thank you for this suggestion. You need to accept the CLA for the first checks to get triggered.
Regards,

Nathanael-Mtd:
Can you adapt the code on lines L33-L35 to use .Values.backend.env with your additions, to avoid setting two times env vars ?

If you can do the same for Huey container (on the same Pod as the backend) and frontend.

Nathanael-Mtd:
Just checked, you should test, but maybe your addition is not needed and the current `.Values.backend.env` option answer your needs

PR #1756, opened 2025-03-24
fix: whitespaces detailview elements showing correctly & ellipses for too big descriptions in modeltable.

melinoix:
Fixed elements white spaces in the detailviews that were not showing properly. (ex: threats detailview)

Added a condition in ModelTable in the case an element have more than 300 chars;  it crop it to add "..." for a more digestive UI (ex: threats modeltable).

## Summary by CodeRabbit

- **Style**
  - Refined list display to preserve line breaks, improving readability.

- **New Features**
  - Introduced automatic truncation for lengthy text in table cells, showing an ellipsis when content exceeds 300 characters.


eric-intuitem (approved):
LGTM with change

PR #1760, opened 2025-03-25
prevent / in domain and perimeter name 

melinoix:
To avoid user confusion in the frontpages using this [folder]/[object] or [perimeter]/[object], we prevent the usage of '/' in the name of a domain and a perimeter. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation to prevent the use of forward slashes in names, ensuring consistency and accuracy during data entry.
  - Automatically updated existing folder and perimeter names by replacing any forbidden slashes with underscores to maintain proper naming conventions.


ab-smith:
Can we check on the front as well?

melinoix:
> Can we check on the front as well?

@ab-smith The backend error is sent on the front end directly. Are you talking about checking the condition "no / in name" directly on the front component?

nas-tabchiche (changes requested):
On the backend, this should be done at the serializer level. See https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation
Let's keep a single generic validator for both models, but instead of overriding `clean`, just explicitly state the `name` field.

We should also normalize the existing perimeters and folders' names that contain a forward slash using a data migration, as merging this PR as-is would create errors on edit forms, which will be frustrating for the users.

nas-tabchiche (approved):
LGTM

eric-intuitem (approved):
LGTM

PR #1764, opened 2025-03-26
feat: link solutions to assets

melinoix:
Solution object can now add an asset and asset object can see the solutions.

We consider that solutions can have multiple assets (like one solution for multiple assets) and inverse (multiple solutions for one asset).

## Summary by CodeRabbit

- **New Features**
  - Solutions now support asset associations, allowing users to link multiple assets.
  - An improved autocomplete field in solution forms makes asset assignment easier.
  - Enhanced filtering options enable users to search and sort solutions by their associated assets.


nas-tabchiche (approved):
LGTM

PR #1830, opened 2025-04-12
Fix locale and provider filters

monsieurswag:


ab-smith:
@monsieurswag  models changed with no migration?

ab-smith (approved):
functional ok

PR #1838, opened 2025-04-13
fix: prevent deletion of builtin objects

eric-intuitem:
preventDelete was ignored

## Summary by CodeRabbit

- **Refactor**
  - Enhanced deletion control logic for table row actions, ensuring that restricted deletion attempts are blocked immediately.
  - The updated evaluation flow provides a more consistent and reliable experience when interacting with deletion features.


ab-smith (approved):
functional ok

PR #1843, opened 2025-04-14
feat: allow reuse of ro/to couples in strategic scenarios

melinoix:
Stopping the limitation of one strategic scenario per SROV.
Erased the use of the filter 'used'

## Summary by CodeRabbit

- **Refactor**
  - Removed an outdated filtering condition that previously limited result sets.
  - Updated the strategic scenario selection process by modifying the criteria used to retrieve available options.


nas-tabchiche (approved):
LGTM. Merging

PR #1844, opened 2025-04-14
Helm : add JSON scheme and change default values

Nathanael-Mtd:
Changes on default values : 
- SMTP :
  - Removal of primary prefix in keys (impact existing user config)
  - Port 25 by default
  - TLS disabled
  - username and password removed
- Persistence disabled
- Ingress disabled

Helm JSON Scheme added to ensure old values are not used, if old values are used, the install will fail.

GitHub actions added to automatically update JSON schema and Helm Docs, and start helm lint.


.github/workflows/helm-generate-dep-docs-schema.yaml.yaml
```diff
@@ -0,0 +1,41 @@
+name: Generate Helm dependencies, docs and schema
```
ab-smith: double yaml extension

.github/workflows/helm-generate-dep-docs-schema.yaml.yaml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          working-directory: charts/ciso-assistant-next
+          input: values.yaml
+
+      - name: Generate helm-docs
+        uses: losisin/helm-docs-github-action@v1
+        with:
+          chart-search-root: charts/ciso-assistant-next
+
+      - name: Commit and push if it changed
```
ab-smith: we don't allow auto commit for now

Nathanael-Mtd: Switched to only checks !

ab-smith (approved):

PR #1847, opened 2025-04-14
fix: ref id should be single line for strategic scenarios

melinoix:
Ref id is now in single line in strategic scenarios instead of multiple lines like description

## Summary by CodeRabbit

- **Refactor**
  - Improved the integration of a form component to ensure a more consistent sourcing approach.
  - These adjustments are entirely behind-the-scenes and do not affect end-user functionality.


nas-tabchiche (approved):
LGTM

PR #1851, opened 2025-04-14
fix: missing implementation groups on exports

Axxiar:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved display logic for implementation groups in compliance assessments, showing the section only when relevant groups are selected.


eric-intuitem:
Perhaps there is a misunderstanding. When no IG is selected, the "IG" functionality is not used, so we use all requirements, regardless of IG. We don't need to list IGs in this situation, because we don't care for them.
I think this is the current behaviour, so probably no need to change.

PR #1872, opened 2025-04-16
fix: audit progress sorting not working

Axxiar:
## Summary by CodeRabbit

- **New Features**
  - Improved sorting of compliance assessments by progress, allowing users to order results by progress in both ascending and descending order.

### DIFF
--- a/backend/core/views.py
+++ b/backend/core/views.py
@@ -3934,6 +3934,23 @@ class ComplianceAssessmentViewSet(BaseModelViewSet):
     ]
     search_fields = ["name", "description", "ref_id"]
 
+    def list(self, request, *args, **kwargs):
+        response = super().list(request, *args, **kwargs)
+
+        ordering = request.query_params.get("ordering", "")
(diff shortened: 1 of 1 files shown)


ab-smith (approved):

backend/core/views.py
```diff
@@ -3934,6 +3934,23 @@ class ComplianceAssessmentViewSet(BaseModelViewSet):
...
+            reverse = ordering_fields[0].startswith("-")  # use only first 'progress'
+            results = sorted(
+                response.data["results"],  # assumes paginated response
+                key=lambda x: x.get("progress", 0),
+                reverse=reverse,
+            )
+            response.data["results"] = results
+
+        return response
+
```
nas-tabchiche, line 3953: This is incorrect, and breaks pagination when ordering by progress. Using aggregation is the correct approach.

https://docs.djangoproject.com/en/5.2/topics/db/aggregation/

Issue #1867, opened 2025-04-15
Filter for action plan not working


newsgrep:
**Describe the bug**
The filter for the action plan does not work.

**To Reproduce**
1. Add the demo data to CA
2. then open: Complinace -> Audits -> "ISO/IEC - International standard ISO/IEC 27001:2022" -> Action plan (https://localhost:8443/compliance-assessments/29989221-7b3c-4a10-891d-3360545a1c65/action-plan?category=policy)
3. Click on filter and select e.g. Category:Policy 
5. notice that the entrys with Category:Process are still displayed.
6. refresh the page, notice that the filter still does not work

**Expected behavior**
The content is filtered.

**Screenshots**
![Image](https://github.com/user-attachments/assets/1bf53e79-2206-42bf-85db-17a0e0563f1a)

**Environment (please complete the following information):**
 - Device: VM/Desktop
 - OS: Linux/Docker
 - Browser Firefox and Chrome
 - Version 2.3.8

Axxiar:
Hello, thanks for reporting :)

newsgrep:
fix looks good. Thank you.

PR #1873, opened 2025-04-16
fix: Action plan filters not working

Axxiar:
This fixes issue #1867 

## Summary by CodeRabbit

- **New Features**
  - Expanded filtering options available on the compliance assessment action plans list, allowing users to filter by multiple fields and date ranges for more precise results.


ab-smith (approved):
ok to fix the regression but we should be able to combine multiple values as well

PR #1888, opened 2025-04-19
Add Italian Cybersecurity Framework

f0r7y:
Framework Nazionale per la Cybersecurity e la Data Protection - Edizione 2025 (v2.1) from https://www.cybersecurityframework.it/

## Summary by CodeRabbit

- **Style**
  - Added a trailing newline at the end of the Contributor License Agreement document.


f0r7y:

eric-intuitem:
Nice work, thanks for adding this IT NIS transposition! Before we can merge I have two quick requests:
1/ Depth = 3 lines
Could you add the depth=3 entries to the implementation‑group lists?
Example: GV.OC‑04 should only appear in IG I / E when you explicitly include it; right now it still shows up, which could lead to inconsistencies later.
2/ Language tag
Because Italian is the default language, you can omit the [IT] tag in columns E and F of the req sheet. Leaving it blank will expose the detail in every UI language.
Minor: if you could also drop the change to the CLA file, that would be great.
Thanks!

f0r7y:
Hi.
How can I drop change in CLA file?
edit language tag
edit depth.
is now ok?
see in attachment
[Framework-Nazionale-CS-DP.xlsx](https://github.com/user-attachments/files/19824613/Framework-Nazionale-CS-DP.xlsx)

eric-intuitem (approved):
LGTM
Will fix details

PR #1924, opened 2025-04-28
feat/pat

nas-tabchiche:
- **limit tokens per user to 5**
- **implement CRUD operations for PATs**
- **fetch user's auth tokens**
- **write AuthTokenCreateSchema**
- **use new PersonalAccessToken model for PATs**
- **create PAT from frontend**
- **use APIView rather than ModelViewSet for PATs**
- **chore: remove dead code**
- **fix PAT retrieval**
- **update PAT create schema**
- **fix schemas**
- **only show PAT once right after it is created**
- **add translations**
- **add button to copy PAT**
- **implement proper PAT deletion in the backend**
- **implement PAT deletion in the frontend**
- **add schema prop to ConfirmModal**


ab-smith (approved):
Functional ok

PR #1956, opened 2025-05-09
feat: support users list import with the data wizard

ab-smith:
## Summary by CodeRabbit

- **New Features**
  - Added support for processing and importing "User" records in the data wizard tool.
  - Introduced a "User" option in the model selection form, allowing users to select and upload user data.

- **Bug Fixes**
  - Improved error handling and validation feedback when importing user records.


scriptforfivem:
Interesting solution to this problem! I've been working on similar challenges recently.

PR #1995, opened 2025-05-18
feat(lib): add fr translation for PCI-DSS

eric-intuitem:
- add fr translation
- fix some errors in ref_id (at 3.5.1)
- use new library format
- fix create_library
- add optional column "urn_id" to use instead of ref_id, useful to fix ref_id errors while keeping the urn unchanged for smooth migration.

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a custom URN suffix using an optional "urn_id" column.
- **Bug Fixes**
  - Improved handling of missing values when extracting requirement node information.
  - Enhanced URN generation logic to ensure uniqueness and prevent duplicate URNs within framework requirement nodes.
- **Documentation**
  - Updated module documentation to mention the new "urn_id" column option.


juliangit971:
The Excel and YAML files seem to be correct after checking and testing them. We may need to do more testing of the create_library changes.

juliangit971 (approved):

PR #2001, opened 2025-05-19
feat(lib): NIST SP-800-66 Rev.2 Restructuring

juliangit971:
In the Excel file :
- The questions for each task were separated into a specific column, and the "answers" sheet was created in order to use the web app's question system. The type of each question was also assigned.
- Implementation groups were assigned to each task in a specific column. There are 3 distinct groups: "Required", "Addressable" and "Neutral".  The “imp_grp” sheet has also been created to use the web app's implementation group system.

In the Python file :
- It has been updated in order to import everything in the correct format into an Excel file, as in the file in this commit.
- Despite the automation achieved, the questions starting with "If" need to be manually reviewed in the Excel file. There are only a few of them.

The YAML file is the updated version of the framwork from this commit's Excel file.

## Summary by CodeRabbit

- **New Features**
  - Enhanced Excel output with structured parsing of sample questions and automatic classification of answer types.
  - Added new columns and sheets for questions, answer types, and implementation group codes in the exported Excel file.
  - Updated metadata to reflect new content and sheet structure.
  - Introduced a new script to sequentially number requirement names in the Excel workbook.

- **Bug Fixes**
  - Corrected a typo in the argument parser description.

- **Chores**
  - Removed unused imports and improved print statement formatting.


juliangit971:
I've regenerated the YAML file with the “--compat” option so that it works seamlessly with the already existing version

juliangit971:
I copied the updated framework in the backend directory so everything should be fine now

Mohamed-Hacene (approved):
0k

eric-intuitem (changes requested):
The urn are changing because the nodeXX compat mode does not match, as we have removed the old questions nodes.
The good approach is to use create_library with the urn_id column filled.
"Note: the urn_id column can be defined to force an urn suffix. This can be useful to fix ref_id errors"

juliangit971:
Thanks for your feedback! It should be fine now

eric-intuitem (approved):
LGTM

PR #2016, opened 2025-05-21
chore: backend urls

Mohamed-Hacene:


Mohamed-Hacene:
Remove data-model label, as it doesn't change the data model. The migration is modified because it was not linted.

eric-intuitem (approved):
LGTM

PR #2066, opened 2025-06-01
images are optimised 

morningstarxcdcode:
Your image file size has been reduced by 21%


ab-smith:
Thank you @morningstarxcdcode , but we cannot accept such changes because some of them are related to logos and the others are involved in tests where the hash and size are checked.
Looking forward to your next contribution :)

PR #2067, opened 2025-06-01
fix: update tests images

ab-smith:
## Summary by CodeRabbit

- **Tests**
  - Updated the expected file size in an attachment test to reflect the new value.


Mohamed-Hacene (approved):
0k

PR #2077, opened 2025-06-03
fix(lib): SOC2 2017 with Rev2022 Framework

tarkadia:
Fixed a typo where a requirement ref ID was named "C6.5.1" instead of "CC6.5.1".
The URN remained unchanged thanks to the "urn_id" column in the Excel file.
The Excel file is now in v2 as well.


eric-intuitem (approved):
LGTM

PR #2078, opened 2025-06-04
fix(lib): Adobe CCF v5 to ISO27001:2022 Mapping

tarkadia:
Fixed an issue where some mapping targets were not assessable.
Targets "9.2" and "9.3" were not assessable because they are groups of assessments. I simply extended the mappings to each of the assessments belonging to each group.

I've left the mapping in v1 just in case.

## Summary by CodeRabbit

- **New Features**
  - Updated mapping file to version 3 with a new publication date of 2025-06-04.
  - Expanded requirement mappings to reference more specific ISO/IEC 27001:2022 subclauses for improved clarity and detail.


tarkadia:
Changes transferred to PR #2179

PR #2094, opened 2025-06-10
fix: inconsistencies for audits created from entity assessments

Mohamed-Hacene:
- When an audit was created from an entity assessment, the folder was correctly set to the new enclave however the perimeter remained that of the entity assessment which is an inconsistency. A new perimeter should be added in the enclave.
- Some Evidence/Control/Exception Forms fixes for folder auto-filling inside an enclave

## Summary by CodeRabbit

- New Features
  - Folder selectors in Applied Control Policy, Evidence, and Security Exception forms now include an additional folder type, offering more choices when linking items.

- Chores
  - Existing audit folders renamed to include perimeter and instance names for clearer organization.
  - Audits are now consistently linked to dedicated perimeters within their enclave folders, improving structure and navigation.


eric-intuitem (changes requested):
A custom migration is needed to keep consistency

eric-intuitem:
Wouldn't it be simpler to make perimeter optional and not use it inside an enclave?

Mohamed-Hacene:
> Wouldn't it be simpler to make perimeter optional and not use it inside an enclave?

It should be simpler yes, for the general case too. However, it will probably mean significant changes on the clasic path too, which we need to evaluate.
I think we should do this when we do it globally.

ab-smith:
Needs reworking and alignment on a new PR

PR #2096, opened 2025-06-10
feat(lib): FINMA 2023/01 Framework + Framework Tool

juliangit971:
Add of the FINMA - Circular 2023/01 "Operational risks and resilience – banks" Framework (French, English, Italian and German).

Nevertheless, I noticed something strange with the URNs after using "convert_library_v2.py". There are URNs with special characters in the generated YAML, such as:
```
- urn: "urn:intuitem:risk:req_node:finma-2023-01:i.-objet-et-champ-d\u2019application"
- urn: "urn:intuitem:risk:req_node:finma-2023-01:v.-garantie-de-la-r\xE9silience-op\xE9rationnelle"
```
I think the "convert_library_v2.py" script should be modified to solve this "problem". I await your opinion on this matter.

_Note: It works perfectly on CISO Assistant despite the special characters, but I think it's better to fix that before merging my PR._

__**UPDATE:**__

The `convert_library_v2.py` script updated from PR #2103 was used to "clean up" URNs that contained special characters and letters with accents.
The script version used corresponds to the 1st commit of the mentioned PR.

**__UPDATE 2:__**

The `footnote_reference_prettifier.py` script has been created. It has been used to make footnote references more visible in the FINMA framework more visible by adding [square brackets] between the numbers next to a word.
The script has been made as generic as possible so that it can be used with any other Excel.
More information in the script's docstring.

I used this script to format the footnote references in the FINMA Framework.

## Summary by CodeRabbit

- **New Features**
  - Added a tool to extract and structure content from the FINMA 2023/01 PDF document.
  - Automatically identifies chapters, subchapters, sections, and paragraphs, and links annotations to relevant content.
  - Outputs structured data with assessability flags and reference IDs into an Excel file for further review.
  - Includes warnings to prompt manual review due to potential extraction inaccuracies.
  - Introduced a utility to enhance footnote references in Excel files by formatting numeric references for improved readability.
  - Updated documentation to include new supported frameworks: Vehicle CyberSecurity Audit v1.1, Cisco Cloud Controls Framework v3.0, and FINMA Circular 2023/01; also updated ISO 27001 versions listed.

PR #2103, opened 2025-06-11
fix(tool): Update convert_library_v2.py

tarkadia:
Added a consistent cleaning mechanism for all URNs in the code using a function called `clean_suffix()`. This function:
- Removes accents
- Replaces spaces with hyphens
- Replaces disallowed characters with underscores (`_`).

This function has been applied wherever necessary

Warnings have been added regarding URNs.
When a URN or URN suffix is automatically cleaned (i.e. modified by `clean_suffix()`), the code compares the original value with the cleaned value. If they differ, a warning is displayed.
They are only displayed if `--verbose` is used, except for key structures such as the main library URN.

The `--verbose` option has been added to provide detailed information on URN cleaning.
- When `--verbose` is used, the script prints warnings if any URN or suffix is changed during cleaning
- Without `--verbose`, only critical URNs (like library URN, object URNs) show warnings

This option could be useful in the future for other purposes.

Tested with the FINMA Framework (PR #2096) which had "dirty" URNs and worked very well.

# **__UPDATE :__**
Changes: 
- The scrip now use a multi mode compatibility system to handle different compatibility modes in the same script.
- Added Compatibility mode "2" so it use the old URN cleanup system (Replace only spaces " " with hyphens "-" sets text in lower case). This mode also preserves the script's old behavior in certain parts of the code prior to the update provided by this branch.
- URNs generated in Compatibility mode 1 (OG compat mode before this update) will not use the new URN cleaning system.
- Version numbers have been added next to compatibility mode descriptions to make it easier to track the age of a compatibility mode.
- The script now saves the YAML file in the current working directory instead of where the Excel file is.
- Verbose messages now start with a 💬 (speech bubble) emoji.
- Added version + parameters used for convert_library_v2.py to the YAML File in the "convert_library_version" field at the top of the file.
- Add Warning system for mapping creation (see corresponding commit message for more information).
- Errors are now properly printed thanks to the "try, except" method.
- The script now exit correctly after an error message.

After a few test, it seems that the script works very well on mappings. I compared the output of this updated version with some original YAML mappings.
I tried the scipt with a few frameworks, but more testing would be needed before validating this update.

## Summary by CodeRabbit

- **New Features**
  - Added a verbose mode that provides detailed warnings when URN suffixes are modified during processing.  
- **Improvements**
  - Enhanced URN suffix normalization by cleaning and standardizing input values for consistency.  
  - Standardized warning messages to improve clarity and context during processing.

### DIFF
--- a/tools/convert_library_v2.py
+++ b/tools/convert_library_v2.py
@@ -14,7 +14,6 @@ Note: the urn_id column can be defined to force an urn suffix. This can be usefu
     
 """
 
-from ast import Nonlocal
 import sys
 import re
 import yaml
@@ -23,11 +22,13 @@ import argparse
 import unicodedata
 import openpyxl
 from pathlib import Path
+from collections import Counter
 
 SCRIPT_VERSION = '2.1'
 
 # --- Compatibility modes definition ------------------------------------------
 # NOTE: No compatibility mode includes another (unless otherwise stated)
+# NOTE: So far, no compatibility mode has an impact on the mapping creation process.
 
 COMPATIBILITY_MODES = {
     0: f"[v{SCRIPT_VERSION}] (DEFAULT) D'ont use any Compatibility Mode",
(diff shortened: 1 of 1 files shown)


tarkadia:
Changes: 
- The scrip now use a multi mode compatibility system to handle different compatibility modes in the same script.
- Added Compatibility mode "2" so it use the old URN cleanup system (Replace only spaces " " with hyphens "-" sets text in lower case). This mode also preserves the script's old behavior in certain parts of the code prior to the update provided by this branch.
- URNs generated in Compatibility mode 1 (OG compat mode before this update) will not use the new URN cleaning system.
- Version numbers have been added next to compatibility mode descriptions to make it easier to track the age of a compatibility mode.
- The script now saves the YAML file in the current working directory instead of where the Excel file is.
- Verbose messages now start with a 💬 (speech bubble) emoji.
- Added version + parameters used for convert_library_v2.py to the YAML File in the "convert_library_version" field at the top of the file.
- Add Warning system for mapping creation (see corresponding commit message for more information).
- Errors are now properly printed thanks to the "try, except" method.
- The script now exit correctly after an error message.

After a few test, it seems that the script works very well on mappings. I compared the output of this updated version with some original YAML mappings.
I tried the scipt with a few frameworks, but more testing would be needed before validating this update.

eric-intuitem (approved):
LGTM

PR #2122, opened 2025-06-18
feat: added "Go back to Ebios RM study" in every pages of ebios-rm

melinoix:
added "Go back to Ebios RM study" in every pages of ebios-rm

## Summary by CodeRabbit

- **New Features**
  - Added a back navigation button to multiple EBIOS RM workshop pages, allowing users to easily return to the main EBIOS RM study page. The button includes a left arrow icon and a localized label for improved usability and accessibility.

- **Style**
  - Enhanced page layouts by introducing a consistent navigation element above key content sections.


Mohamed-Hacene (changes requested):
Functional ok, however we have to use Anchor component instead of vanilla <a>

PR #2139, opened 2025-06-19
fix: invalidate table after navigating

nas-tabchiche:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the reliability of the detail view panels when switching between related models, ensuring content updates correctly.


eric-intuitem (approved):
LGTM

PR #2154, opened 2025-06-20
fix: action plan link from ebios RM study

ab-smith:
## Summary by CodeRabbit

- **Refactor**
	- Updated the URL for the final step of the fifth workshop to use "/action-plan" instead of "/remediation-plan".


eric-intuitem (approved):
LGTM

PR #2164, opened 2025-06-24
fix: action plan pdf export when no status in control

Mohamed-Hacene:
## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the status label for items without a status in action plan PDFs to display as "--" instead of "No status".
  - Adjusted the table layout in action plan PDFs so that the "No entries found" message correctly spans all columns.


eric-intuitem (approved):
LGTM

PR #2167, opened 2025-06-24
Add restart: always to frontend service in docker-compose

Qwarctick:
All other services in the stack already have a restart policy (restart: always or equivalent). The lack of one on frontend was causing the stack not to start properly after a server reboot.

This PR fixes that by aligning frontend with the expected behavior.


Qwarctick:

ab-smith (approved):

PR #2173, opened 2025-06-25
Add feared events when generating a risk analysis in Ebios-RM

monsieurswag:


Mohamed-Hacene:
Closing for now as it is basic implementation. Can be reopened when discussed and fully implemented.

PR #2176, opened 2025-06-25
feat: add EBIOS RM selection via context menu

monsieurswag:
## Summary by CodeRabbit

- **New Features**
  - Added the ability to select or deselect items directly from the context menu for various entities, including feared events, stakeholders, attack paths, and operational scenarios.
- **Localization**
  - Introduced new translations for "select" and "deselect" actions across all supported languages.
- **Bug Fixes**
  - Resolved minor issues in translation files, such as missing commas, to ensure proper display and functionality.


ab-smith (dismissed):
the context menu doesn't disappear after the click,
also the click is only working if we click exactly on the text and should be operational even if we click on the row (as we did for applied control context menu)

monsieurswag:
> the context menu doesn't disappear after the click, also the click is only working if we click exactly on the text and should be operational even if we click on the row (as we did for applied control context menu)

The ContextMenu now disappears on click.
The ContextMenu displays whenever i click anywhere on a row on my side (tested on both firefox and chromium) i can't reproduce this bug.

nas-tabchiche (approved):
LGTM. The interface of `changeSelected` could be cleaner, but given that `SelectObject` is row-scoped, this implementation is good enough.

Mohamed-Hacene:
@monsieurswag can be merged after fixing conflicts

PR #2179, opened 2025-06-25
feat(lib): Adobe CCF v5 Mappings

tarkadia:
## Mappings
The following mappings have been created :
- Adobe CCF v5 ➡️ BSI C5 2020 (Imported & updated from PR #1968)
- Adobe CCF v5 ➡️ NIST CSF 1.1
- Adobe CCF v5 ➡️ Cyber Essentials: Requirements for IT infrastructure v3.1
- Adobe CCF v5 ➡️ SOC2 Rev2022 (Imported & updated from PR #2076)
- Adobe CCF v5 ➡️ FedRAMP Rev. 5
- Adobe CCF v5 ➡️ PCI DSS 4.0
- **[NEW]** Adobe CCF v5 ➡️ ISO27001:2022 (updated to v2)

## ~~New Tool~~ (Moved to PR #2226)
~~The new tool `mapping_expander.py` has been created (imported from PR #2076)~~

## ~~Other Updates~~
Minor updates has been made to the following scripts :
~~- `convert_library_v2.py`:~~ (Moved to PR #2196)
    - ~~The script now checks if a mandatory field for mappings ("source_node_id", "target_node_id" or "relationship") is missing. If so, a readable error is raised.~~
    - ~~The tip printed at the end of the script about the `--verbose` option will no longer appear when `--verbose` has been used.~~
~~- `mapping_expander.py`:~~ (Moved to PR #2226)
   - ~~Fixed exception file path~~
    - ~~Line number of the row where a warning or an exception has been issued is now also printed~~
    - ~~**[NEW]** It is now possible to expand the `source_node_id` column using the `--reverse` parameter. The code has been restructured in the process.~~

## Misc
Some cleaning has been made in the files :
- Moved the file `mapping-adobe-ccf-v5-to-iso27001-2022.xlsx` from `mapping-adobe-ccf` to `adobe/mappings`
- Removed old version of ENS Framework Excel file
- Removed old version of PCI DDS Framework Excel file

## Summary by CodeRabbit

* **New Features**
  * Added new mapping libraries enabling bidirectional mappings between Adobe CCF v5 and ISO/IEC 27001:2022, as well as Adobe CCF v5 and NIST CSF 1.1.
  * Introduced a script to export mappings from Adobe CCF v5 Excel files for easier data handling.

* **Enhancements**
  * Updated the Adobe CCF v5 to ISO/IEC 27001:2022 mapping library with improved metadata, refined mapping granularity, and added reverse mappings for comprehensive traceability.

PR #2196, opened 2025-06-27
fix(tool): Update convert_library_v2.py

tarkadia:
Added ability to use the `urn_id` column in compatibility mode 1 (Use legacy URN fallback logic (for requirements without ref_id)).

Updates from PR #2179 have also been transferred here, including:
- The script now checks if a mandatory field for mappings (`source_node_id`, `target_node_id` or `relationship`) is missing. If so, a readable error is raised.
- The tip printed at the end of the script about the `--verbose` option will no longer appear when `--verbose` has been used.

## Summary by CodeRabbit

* **New Features**
  * Improved handling of URN suffix generation, allowing more flexible use of the "urn_id" column in compatibility mode 1.
  * Enhanced validation for requirement mappings, ensuring required fields are present and non-empty.
* **User Experience**
  * Refined messaging for the `--verbose` flag, displaying tips only when appropriate.

### DIFF
--- a/tools/convert_library_v2.py
+++ b/tools/convert_library_v2.py
@@ -750,7 +750,7 @@ def create_library(input_file: str, output_file: str, compat_mode: int = 0, verb
                             ref_id_urn = f"node{counter - counter_fix}-{counter_fix}"
                         else:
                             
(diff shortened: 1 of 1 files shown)


eric-intuitem (approved):
LGTM

PR #2205, opened 2025-06-27
feat(lib): SCF 2025.2.2 Framework + Update SCF 2024.2 Framework

tarkadia:
## SCF 2025.2.2 Framework
Add of the **Secure Controls Framework (SCF)** ~~2025.1.1~~ **2025.2.2** Framework.

**[UPDATE 09/26]** After discussions with the team, we decided to create a whole new framework for the new version of SCF. I also updated the files to the latest version of SCF, as this PR was quite old.

## SCF 2024.2 Framework
Updates for the SCF 2024.2 Framework:
- Fixed min score to `0` instead of `1`
- Added version to framework name

## Other Changes
- `convert_library_v2.py`: Fixed an issue when a row in the `description` column for the Score Sheet was empty

### DIFF
--- a/backend/library/libraries/scf-2025.2.2.yaml
+++ b/backend/library/libraries/scf-2025.2.2.yaml
@@ -25,22 +25,22 @@ objects:
     scores_definition:
     - score: 0
       name: Not Performed
-      description: None
+      description: null
     - score: 1
       name: Performed Informally
-      description: None
+      description: null
     - score: 2
       name: Planned & Tracked
-      description: None
+      description: null
     - score: 3
       name: Well Defined
-      description: None
+      description: null
     - score: 4
       name: Quantitatively Controlled
-      description: None
+      description: null
     - score: 5
       name: Continuously Improving
-      description: None
+      description: null
     implementation_groups_definition:
     - ref_id: tier1
       name: Tier 1 - Strategic


eric-intuitem:
Isn't it possible to update SCF2024 in SCF2025?

tarkadia:
Not really. There's a mistake in the structure of the original framework where the "Physical & Environmental Security" section appears 3 times, and consequently, divides this section into 3. On top of that, the new requirements added to the framework are all over the place. It'd get very tricky to update this framework over time, considering that the original framework works with the old URN creation system. Not to mention the potential requirements that can be removed after an update.

tarkadia:
After a few verifications, I found that updating SCF 2024 to SCF 2025 wouldn't be a problem after all. There are only new rerquirements added, none removed. The only problem will be with the framework URN, as the URN contains the framework's version number.

tarkadia:
The problem is that sticking with the old nodes naming system could be constraining in the future for this framework if we intend to introduce new features based on the URN of the nodes. What's more, switching to the v2 version of node naming will save us from wasting time putting the right node number in the urn_id column for section nodes. I think the best solution would be to create a mapping between SCF 2024 and SCF 2025, then update SCF from version 2025 for future versions.

tarkadia:
After discussions with the team, we decided to create a whole new framework for the new version of SCF. I also updated the files to the latest version of SCF, as this PR was quite old.

PR #2215, opened 2025-06-30
feat: allow supporting assets for feared events

Mohamed-Hacene:


eric-intuitem (changes requested):
Nice. But now we should indicate if each asset is primary or support in the dropdown menu (e.g. "DEMO/Ecommerce portal [Primary]"). Variant: show primary assets first, then a separator, then secondary assets.

ab-smith (changes requested):
needs reworking after the hotfix,
no need to indicate the primary/supporting at this stage.

ab-smith (approved):

eric-intuitem (approved):
LGTM

PR #2220, opened 2025-07-01
feat(PLT-000): add init containers to helm chart

nicolaslaunaySO:
Add init containers to the front-end and back-end pods, to run custom scripts like SSO setup for example.

### DIFF
--- a/charts/ciso-assistant-next/Chart.yaml
+++ b/charts/ciso-assistant-next/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: ciso-assistant
 description: A Helm chart for CISO Assistant k8s's deployment
 type: application
-version: 0.4.3
+version: 0.5.0
 appVersion: "v2.5.10"
 icon: https://intuitem.com/ciso-assistant.svg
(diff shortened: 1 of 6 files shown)


charts/ciso-assistant-next/values.yaml
ab-smith, line 1: Can you please update the comment with this?   
```
   # -- Init containers for the backend
    # @security Init containers run with the pod's security context and can access
    # all mounted volumes and secrets. Only use trusted images and review commands carefully.
    # initContainers:
    # - name: init-container
    #   image: busybox
    #   command: ["sh", "-c", "echo 'Hello, World!'"]
```

charts/ciso-assistant-next/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: ciso-assistant
 description: A Helm chart for CISO Assistant k8s's deployment
 type: application
-version: 0.4.3
+version: 0.5.0
```
ab-smith, line 5: can you align this with the recent version please?

PR #2227, opened 2025-07-01
feat(i18n): greek translation

ab-smith:
## Summary by CodeRabbit

* **New Features**
  * Added Greek as a supported language across the application, including language selection menus and localization files.
  * Updated language lists and locale settings to include Greek, enhancing internationalization options.
  * French localization now includes translations for both Greek and Ukrainian languages.


eric-intuitem (changes requested):
We need to solve the locale issue provoking the regression for functional tests.

PR #2232, opened 2025-07-03
chores: switch all cookies to httpOnly

ab-smith:


nas-tabchiche (changes requested):
The `from_login` and `show_first_login_modal` cookies are accessed from the browser, therefore setting them as `httpOnly` cookies will break existing client-side behavior.

ab-smith:
checks confirmed the breaking of the functionalities, objective met and we need just to better document this to avoid false positives on security checks.

PR #2233, opened 2025-07-03
fix: publish qualifications in root folder

Mohamed-Hacene:
Publish Qualification objects in the root folder so they can be viewed by any role other than administrator.

## Summary by CodeRabbit

* **New Features**
  * All qualifications are now marked as published.
* **Refactor**
  * Enhanced the qualification model with improved publishing behavior.


backend/core/migrations/0080_auto_20250703_1217.py
```diff
@@ -0,0 +1,16 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0079_finding_evidences_findingsassessment_evidences"),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            sql="UPDATE core_qualification SET is_published = TRUE;",
+            reverse_sql="UPDATE core_qualification SET is_published = FALSE;",
+        ),
```
nas-tabchiche: Let us use rely on the ORM instead of firing raw SQL queries here. We support multiple database backends, and this may shoot us in the foot some day.

nas-tabchiche (changes requested):

eric-intuitem (approved):
LGTM

PR #2238, opened 2025-07-04
feat(tool): prepare_mapping_v2.py

tarkadia:
## Script
Works exactly like `prepare_mapping_v1.py`, except that it uses the v2 library format for the output Excel file.
It also has an error handling system and better code structuring than v1.

## Update
In v1, there was an issue when the URN ID was longer than the usual `node_base_urn:node_id` format. In fact, when we had nested requirements and the nested requirement didn't have a `ref_id`, the algorith making the URN (in `convert_library.py`) for the node would take the `node_base_urn`, followed by the node ID one level higher than the current one, with a number at the end to identify the node.
We could have something like this for exmaple :
```
node_base_urn:previous_node_id:1
```

The problem here is that v1 of `prepare_mapping.py` was taking the very last bit of the URN after the last `:` as the correct node ID, whereas it's completely wrong. Taking the last part of the URN in the example (here `1`) would mean that we're referring to the URN `node_base_urn:1` and not the actual URN `node_base_urn:previous_node_id:1`.

To fix this issue, the URN is now split at `target_node_base_urn:` and we take the last part of the split URN.
This way, the node ID to be used for mapping will be `previous_node_id:1` instead of `1`.

This problem occured when I wanted to make a mapping with NIS2 (requirement `12.5.1` had the node ID `12.5:1`).

## Summary by CodeRabbit

* **New Features**
  * Introduced a command-line tool to generate an Excel mapping file between two YAML-based frameworks.
  * The generated Excel file includes metadata, mapping templates, guidelines, and detailed sheets for both source and target frameworks.
  * Provides clear error messages for invalid input files and confirms successful file creation.

### DIFF
--- a/tools/prepare_mapping_v2.py
+++ b/tools/prepare_mapping_v2.py
@@ -1,0 +1,164 @@
+"""
+Prepare Mapping v2
+
+This script generates a mapping Excel file between two YAML-based frameworks
+used in CISO Assistant. It extracts relevant metadata from the source and target
+YAML files, builds a structured Excel workbook with mapping sheets, and saves it
+under a filename derived from the two frameworks' reference IDs.
+
+Usage:
+    python prepare_mapping.py source.yaml target.yaml
+"""
+
+
+import openpyxl
+import argparse
+import yaml
+
+
+def generate_mapping_excel(source_yaml, target_yaml):
+    packager = "intuitem"
+
+    print(f"⌛ Parsing \"{source_yaml}\" & \"{target_yaml}\"...")
+    # Load YAML files
+    with open(source_yaml, "r", encoding="utf-8") as file:
+        source = yaml.safe_load(file)
+    with open(target_yaml, "r", encoding="utf-8") as file:
+        target = yaml.safe_load(file)
+
+    # Extract metadata from source and target libraries and frameworks
+    source_library_urn = source["urn"]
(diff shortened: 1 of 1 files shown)


tarkadia:
## Update
In v1, there was an issue when the URN ID was longer than the usual `node_base_urn:node_id` format. In fact, when we had nested requirements and the nested requirement didn't have a `ref_id`, the algorith making the URN (in `convert_library.py`) for the node would take the `node_base_urn`, followed by the node ID one level higher than the current one, with a number at the end to identify the node.
We could have something like this for exmaple :
```
node_base_urn:previous_node_id:1
```

The problem here is that v1 of `prepare_mapping.py` was taking the very last bit of the URN after the last `:` as the correct node ID, whereas it's completely wrong. Taking the last part of the URN in the example (here `1`) would mean that we're referring to the URN `node_base_urn:1` and not the actual URN `node_base_urn:previous_node_id:1`.

To fix this issue, the URN is now split at `target_node_base_urn:` and we take the last part of the split URN.
This way, the node ID to be used for mapping will be `previous_node_id:1` instead of `1`.

This problem occured when I wanted to make a mapping with NIS2 (requirement `12.5.1` had the node ID `12.5:1`).

Mohamed-Hacene (approved):
0k

PR #2250, opened 2025-07-07
fix: undefined in perimeters status french translation

Axxiar:
## Summary by CodeRabbit

* **Style**
  * Improved the appearance of select dropdowns in several forms by disabling the "double dash" option for specific fields.


eric-intuitem (approved):
LGTM

PR #2255, opened 2025-07-07
feat: select implementation groups at campaign creation

Mohamed-Hacene:
## Summary by CodeRabbit

* **Refactor**
  * Updated the visibility logic for the implementation groups selection field to improve consistency with other form fields. Visibility is now managed internally rather than through conditional rendering.


Mohamed-Hacene:
Closing for now, it will be used in a future improvement

PR #2263, opened 2025-07-08
fix: nist csf 2.0 scores localization

nas-tabchiche:


ab-smith:
doesn't seem to do what's mentioned on the title:
image

tarkadia:
That's very weird, it works very well on my end with the latest version of CISO Assistant:
image

But it made me notice another problem. In "Table mode", the score translations don't work. Whether I choose Spanish or French, the English version is always displayed:
image
image

PR #2288, opened 2025-07-15
fix: nested model tables missing values 

Mohamed-Hacene:
## Summary by CodeRabbit

* **Refactor**
  * Improved the way table fields are handled, allowing separate configuration for table headers and body columns.
  * Enhanced logic for updating table fields to better distinguish between default and custom configurations.


ab-smith (approved):
functional ok

PR #2312, opened 2025-07-21
test: Follow-ups module functional tests

Axxiar:


Axxiar:
Commit ["can specify addbutton value in createItem" - 8b50653](https://github.com/intuitem/ciso-assistant-community/pull/2312/commits/8b50653cb588629ed75dc56ea9a34706bb693f0e) allows us to specify which “addButton” we are targeting using its value when we call createItem, which is useful when there are multiple addbuttons (e.g., multiple tabs with an addbutton inside each one). If not specified it will click the first one found.

PR #2320, opened 2025-07-22
feat(lib): Example Framework

tarkadia:
# Example Framework

The structure of an Excel framework file may not be very easy to understand at first glance. That's why `example_framework.xlsx` has been created to help understand the structure of a framework.

The sheets and cells are colored to make it easier to understand the structure of the file, and notes have also been added to some cells. If a cell has a note, you will find a colored triangle at the top right of the cell (red in Excel, brown in LibreOffice). Hover your mouse over the cell with this triangle to see the note.

# Documentation

- `README.md` in `/tool`
    - Added documentation for the `example_framework.xlsx` file
    - Fixed documentation for `Threats`, `Reference controls`, `Implementation groups` and `Requirement mapping sets`

## Summary by CodeRabbit

* **Documentation**
  * Added a new "Example framework" section to the README, introducing an example Excel file with interactive notes to clarify framework structure.
  * Updated documentation to specify that the `name` and `ref_id` fields are now mandatory for `_content` tabs in `threats`, `reference_controls`, `requirement_mapping_set`, and `implementation_groups` objects.


eric-intuitem:
"Not all sheets are needed to create a framework. Just use the ones you need when creating a framework."
-> When creating a library, keep only the tabs that are necessary. For example, if no implementation groups are used, remove the corresponding tabs.

tarkadia:
It should be good now

eric-intuitem:
Perhaps using extracts of real examples would be clearer (e.g. an extract of CISA vendor assessment, NIST CSF, ...)

tarkadia:
The main problem is that there's no example of a framework having absolutely every sheets at the same time. What we can do is recommend that users consult the list of framework Excel files we have in the `./tools/excel` directory if they really want to see a concrete case. The example framework is mainly there to make the Excel file structure easier to understand by seeing every possible cases.

PR #2325, opened 2025-07-23
test: functional tests for analytics module

melinoix:
Some functional tests for analytics module by creating different objects and then verifying some of the graphs / tables to appear on the analytics in the context i provided.

### DIFF
--- a/frontend/tests/functional/detailed/analytics.test.ts
+++ b/frontend/tests/functional/detailed/analytics.test.ts
@@ -1,5 +1,13 @@
 import { test, expect } from '../../utils/test-utils.js';
 import { LoginPage } from '../../utils/login-page.js';
+import { PageContent } from '../../utils/page-content.js';
+import { Page } from '@playwright/test';
+
+async function redirectToAnalytics(page: Page): Promise<void> {
+	await page.getByTestId('accordion-item-overview').click();
+	await page.getByTestId('accordion-item-analytics').click();
+	await page.waitForTimeout(500);
+}
 
 test('Analytics full flow - creation, validation and cleanup', async ({ page }) => {
 	const loginPage = new LoginPage(page);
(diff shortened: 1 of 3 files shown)


frontend/tests/functional/detailed/analytics.test.ts
```diff
@@ -0,0 +1,281 @@
...
+		await expect(page.locator('#page-title')).toHaveText('Analytics');
+		await expect(page).toHaveURL('/analytics');
+
+		await expect(page.getByTestId('card-controls-total')).toHaveText('46');
+		await expect(page.getByTestId('card-controls-active')).toHaveText('4');
+		await expect(page.getByTestId('card-controls-deprecated')).toHaveText('2');
+		await expect(page.getByTestId('card-controls-to do')).toHaveText('2');
+		await expect(page.getByTestId('card-controls-in progress')).toHaveText('5');
+		await expect(page.getByTestId('card-controls-on hold')).toHaveText('1');
+		await expect(page.getByTestId('card-controls-Missed ETA')).toHaveText('6');
```
nas-tabchiche: This part is brittle, as the counts here depend on the number of items in demo data, which will evolve and thus break this test every time a control is added/removed.
We should not rely on demo data for this test, but rather create the objects manually here.

nas-tabchiche (changes requested):
Let's not import demo data, and perform manual setup instead

frontend/tests/functional/detailed/analytics.test.ts
```diff
@@ -0,0 +1,217 @@
...
+		await expect(page.getByText('International standard ISO/IEC 27001:2022')).toBeVisible();
+		await expect(page.getByText('test-control-2')).toBeVisible();
+	});
+
+	await test.step('Verify redirection in analytics', async () => {
+		await redirectToAnalytics(page);
+
+		await page.getByTestId('card-controls-total').click();
+		await expect(page).toHaveURL('/applied-controls');
+		await redirectToAnalytics(page);
```
nas-tabchiche, line 165: The `redirectToAnalytics` call here might be too early, this makes the test flaky. See https://github.com/intuitem/ciso-assistant-community/actions/runs/16570919501/job/46866024781?pr=2325#step:19:76

I recommend we check the page title after the URL assertion and before redirecting to analytics to prevent race conditions here.
Take a look at https://github.com/intuitem/ciso-assistant-community/blob/main/frontend/tests/functional/user-route.test.ts#L32-L32 for reference

nas-tabchiche (dismissed):
Almost there. This branch can be merged once the flakiness issue is solved

PR #2333, opened 2025-07-24
fix: effort medium option in applied control

Mohamed-Hacene:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved option label rendering in select fields to prevent incorrect translations when the option key is "m".


eric-intuitem (approved):
LGTM

PR #2334, opened 2025-07-24
Fix/suggested control duplicated by lang

Mohamed-Hacene:


frontend/src/lib/components/Forms/Select.svelte
Mohamed-Hacene, line 1: This change has been introduced in the PR #2333, we should merge it first

ab-smith (approved):

PR #2344, opened 2025-07-28
fix: ro/to pertinence filtering and ordering

nas-tabchiche:
Added a queryset annotation for RO/TO pertinence field, which is dynamically computed.
This allows to filter and sort against it in DRF API views.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to filter and sort RoTo items by pertinence in the interface.
  * Pertinence is now always available and precomputed when viewing RoTo items.

* **Improvements**
  * The pertinence field in RoTo items now uses a clearer display label in the interface and API responses for better readability.


eric-intuitem (approved):
LGTM

PR #2346, opened 2025-07-28
fix: risk matrix impact filter options

nas-tabchiche:
This fixes filtering against risk matrix impact (e.g. in feared events or strategic scenarios)

- **properly serialize risk matrix gravity choices**
- **replace bitwise OR with null checks for autocomplete value**


eric-intuitem (changes requested):
Better, but if I have several matrices with various scales, there is an issue selecting and deselecting values.

nas-tabchiche:
Another UX must be devised for this, it is not intuitive as-is.

PR #2349, opened 2025-07-29
test: sensible defaults for e2e-tests script

nas-tabchiche:
Set sensible default values for settings in e2e tests script and update README


Mohamed-Hacene (approved):
0k

PR #2354, opened 2025-07-29
users improvements 2907

ab-smith:
- **feat: add observation field to user**
- **add filters to users list**


eric-intuitem (approved):
LGTM

PR #2367, opened 2025-07-31
fix: add user observation during creation

ab-smith:
Fix the issue of not saving the user observation during its creation.
It was operational only if the user exists.

## Summary by CodeRabbit

* **New Features**
  * Added an optional "observation" field when creating a new user.


nas-tabchiche (approved):
LGTM. Can be merged on `main` once CI is green

PR #2400, opened 2025-08-13
feat: add observation field to applied control, asset and finding

ab-smith:


eric-intuitem (changes requested):
There are a few inconsistencies:
- the observation field in the incident "add an event" is not shown as md, while in fact it is considered as such in the viewer
- Same issue for observation in finding tracking, not shown as md in the edit page, while considered as md in viewer

ab-smith:
> There are a few inconsistencies:
> 
> * the observation field in the incident "add an event" is not shown as md, while in fact it is considered as such in the viewer
> * Same issue for observation in finding tracking, not shown as md in the edit page, while considered as md in viewer

Fixed for findings tracking. For incidents, some extra rework will be needed to adjust to the rendering of each timeline entry, and it was left out of scope on purpose.

eric-intuitem (approved):
LGTM

PR #2406, opened 2025-08-14
feat(pro): 'publish' feature flag

nas-tabchiche:
- **prepare 'publish' form action**
- **l10n**
- **publish object PoC**
- **some l10n**
- **simplifications**


ab-smith:
Dropped for now, we need to rethink this feature

PR #2434, opened 2025-08-24
feat: filters for findings table

ab-smith:
- **feat: filters on Findings view**
- **filter per owner**

## Summary by CodeRabbit

- New Features
  - Findings list now includes multi-select filters for Severity, Status, and Owner, with autocomplete dropdowns for quick selection.
  - Filtered results in the findings view reflect chosen values across all three filters.
  - API enhancements: added support to filter findings by severity and a new endpoint to retrieve users who own at least one finding, enabling owner-based filtering in clients.


eric-intuitem (approved):
LGTM

PR #2438, opened 2025-08-24
feat: Add a no-redirect save button and rename the old one

monsieurswag:
## Summary by CodeRabbit

* **New Features**
  * Two save options on requirement assessments: a primary "Save" that saves and stays on the page, and a secondary "Save and continue" that saves and advances to the next step.

* **Localization**
  * Added "Save and continue" translations across many locales so the new option appears in supported languages.


Mohamed-Hacene (approved):
0K

PR #2439, opened 2025-08-24
fix: Make domain import button clickable only once

monsieurswag:
## Summary by CodeRabbit

- New Features
  - Save buttons now show a loading spinner and message while a submission is in progress.

- Enhancements
  - Pre-validation for folder-import submissions checks input before sending.
  - Save actions are disabled during loading; cursor indicates waiting.
  - Consistent loading and validation behavior applied to both primary and modal save buttons.

- Bug Fixes
  - Prevents duplicate submissions by blocking repeated clicks during active submits.


Mohamed-Hacene (dismissed):
If there is an error during domain import, we get stuck in the loading state:
image

We should use this spinner:
image

eric-intuitem (approved):
LGTM

PR #2499, opened 2025-09-09
feat: filter risk assessments by status "--"

Axxiar:


backend/core/locale/fr/LC_MESSAGES/django.po
Axxiar, line 1: Align french backend translation "En examen" with frontend "En révision"

PR #2506, opened 2025-09-10
perf: conditional serializers on read endpoints

nas-tabchiche:
This PR updates the `SerializerFactory` class to allow for action-specific serializers within a ViewSet.

The primary motivation is to address over-fetching in list endpoints. Previously, our `list` and `retrieve` views used the same serializer, leading to unnecessary data being fetched and serialized for list views, which negatively impacted performance due to either N+1 queries or larger payloads sent over the wire.

We can now define a lightweight `<Resource>ListSerializer` specifically for the list action, while using a more detailed serializer for the retrieve action. The factory will automatically select the appropriate class: `<ResourceList>Serializer` for `list` views, and `<Resource>RetrieveSerializer` for `retrieve` views, with a fallback to a general `<Resource>ReadSerializer` if an action-specific one is not defined.

## Summary by CodeRabbit

- New Features
  - Enhanced list views for Assets, Applied Controls, and Compliance Assessments with richer fields (e.g., progress, findings count, ownership, labels, paths, version).

- Refactor
  - More consistent serializer selection across views, improving reliability of list and detail responses.
  - Compliance Assessment API: streamlined framework object (removed implementation group definition, reference controls, min/max score) for leaner payloads.


ab-smith:
Remember to check that the fields for listSerializer act only on first level and don't generate N+1 (eg. framework on compliance assessment)

nas-tabchiche:
So far, the only field that generates n+1 queries for compliance assessments is `progress`. It's an easy solve though, we need a model manager to annotate the base queryset with the `progress` composite field, which is a bit out of scope for this PR.
I'm opening another PR so that we can merge it ASAP.

EDIT: just opened #2507 to address this.

ab-smith:
First check with 43 applied controls on the listing page:

before
58 requests, 16 duplicates
after
29 requests, 6 duplicates


Within a risk scenario, for the applied-controls endpoint:

before:
57, 16
after:
28, 6

no noticeable gain on the compliance-assessmeent list page

nas-tabchiche:
The gains for the compliance assessments endpoint are mostly on the response size.
On `main` (only demo data loaded): `Content-Length: 39883`,
On this branch (only demo data loaded: `Content-Length: 1933`.

This is mostly due to only serializing the assessment's framework's `name`, `ref_id` and `id`. On `main` we serialize its reference controls, the implementation groups' definitions... Which add up a fair bit to the total response size.

ab-smith:
we'll work on the three-levels serializers as discussed and done on recent PRs but not generic yet

PR #2514, opened 2025-09-12
perf: derive serializer_fields.PathField from serializers.SerializerMethodField to prevent n+1 queries

nas-tabchiche:
We have two ways to compute the path of a given object:
- use the `object.get_folder_full_path` model method, **will cause n+1 queries**
- use the `serializer.context.['optimized_data']` to take advantage of `BaseModelViewSet._get_optimized_object_data`, **costs a single query** for all serialized objects

This PR implements the second option.

> [!NOTE]
> This should be merged before #2511, and #2511 should be adapted to the new serializer field by removing the `source` parameter from the field's arguments.

```diff
-    path = PathField(source="get_folder_full_path", read_only=True)
+    path = PathField(read_only=True)
```

## Summary by CodeRabbit

- Bug Fixes
  - Path data in API responses is now normalized to consistent id/label lists and safely handles empty, invalid, or unexpected values.

- Refactor
  - Path resolution standardized across read serializers for consistent behavior and simpler configuration.

- Performance
  - Data fetching optimized for several list/detail endpoints to reduce database queries and improve response performance.


ab-smith:
@coderabbitai full review

PR #2547, opened 2025-09-17
fix: Add owner to RiskScenario create modal and display it into the detail view

monsieurswag:
## Summary by CodeRabbit

- New Features
  - Assign multiple owners to a Risk Scenario via a new multi-select, autocomplete Owners field in the form.
  - Owners are now displayed on the Risk Scenario detail page, each shown as a clickable link to the corresponding user profile.
  - Improved clarity with an explicit Owners section in the scenario header alongside existing details.


ab-smith:
we want to keep the scenario creation form as minimalist as possible

PR #2568, opened 2025-09-22
Fix: hide edit button when not permitted to edit

monsieurswag:
## Summary by CodeRabbit

- Bug Fixes
  - Fixed inconsistent availability of create, edit, and delete actions by using uniform permission checks, independent of URL parameters.
  - Reduced cases where actions were incorrectly shown or hidden when a model is present.

- UI/UX
  - Context menu now displays edit/view options only when explicitly allowed, preventing misleading edit options.
  - More predictable action visibility in tables and row menus for a clearer, consistent experience.


frontend/src/lib/components/ModelTable/ModelTable.svelte
```diff
@@ -347,36 +347,32 @@
...
-						action: 'add',
-						model: model.name,
-						domain:
-							folderId ||
-							page.data?.data?.folder?.id ||
-							page.data?.data?.folder ||
-							page.params.id ||
-							user.root_folder_id
-					})
-				: Object.hasOwn(user.permissions, `add_${model.name}`)
```
nas-tabchiche, line 370: Removing the `page.params.id` check hides the create button on list views even though the user has the permission to create the resource in a given folder other than global.
If it is defined (i.e. we are inside an object's detail or edit view), we check user permissions agains the object's folder, otherwise against any folder the user has access to.

ab-smith:
re-open after re-alignment and if it's still relevant

PR #2580, opened 2025-09-24
feat: collapsible detail view

martinzerty:
## Summary by CodeRabbit

* **New Features**
  * Detail view now collapses long field lists, showing the first 10 by default with a toggle to view more / view less when additional fields exist.

* **Localization**
  * Added "View more" / "View less" labels across all supported locales so the toggle is translated for users in each language.


Mohamed-Hacene (approved):
0k

PR #2597, opened 2025-09-26
fix: Apply security objective scale via /object endpoint

monsieurswag:


nas-tabchiche (changes requested):
This does not seem to work as intended on my end.

With scale `0-4`:

Image

With scale `1-4`:

Image

PR #2632, opened 2025-10-01
Fix:  the table of content in Audit table mode to handle empty title strings

tchoumi313:
image

## Summary by CodeRabbit

- **New Features**
  - Smarter Table of Contents for compliance assessments: only valid title entries are included, items without a valid reference or name are excluded, and very deep references are filtered out.
  - Heading levels are inferred from reference depth (capped at level 4) so TOC better reflects content hierarchy.

- **Improvements**
  - More responsive TOC updates and controls as assessment content changes.


tchoumi313:
> It would be preferable to only include "title" nodes, i.e. nodes with the following characteristics:
> 
> * non-assessable
> * have children nodes
> * have a non-empty ref_id and/or name (which we will display)
> * have depth <= 4

I think it's okay now

image

ab-smith (approved):

PR #2646, opened 2025-10-03
fix: ebios workshop 3 invalid sort by entities

Axxiar:
In ebios/workshop 3/activity 1, sorting by entity sorts by entity.id, not entity.name

## Summary by CodeRabbit

* **New Features**
  * Enhanced sorting functionality with improved field mapping capabilities, allowing more intuitive ordering of results across the application.
  * Expanded search functionality with additional searchable fields for better discoverability and filtering options.

### DIFF
--- a/backend/core/views.py
+++ b/backend/core/views.py
@@ -637,12 +637,12 @@ class CustomOrderingFilter(filters.OrderingFilter):
         if ordering_list is None:
             return None
 
-        view_mapping = getattr(view, "ordering_mapping", None)
-        mapping = (
-            view_mapping
-            if view_mapping is not None
-            else getattr(self, "ordering_mapping", None)
-        )
+        # Merge ordering_mapping from all classes in the view MRO that define it
+        # directly, going base→derived so subclass entries take precedence.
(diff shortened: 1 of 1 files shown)


eric-intuitem (changes requested):
This breaks the advance sorting for user groups (bug visible in pro version only).
This PR should not touch this part.

backend/core/views.py
```diff
@@ -230,11 +230,48 @@ class Meta:
...
+
+    ordering_mapping: dict[str, str]
+
+    def get_ordering(self, request, queryset, view) -> Optional[list[str]]:
+        """Map ordering terms based on `ordering_mapping` when provided on the subclass."""
+        ordering_list = super().get_ordering(request, queryset, view)
+        if ordering_list is None:
+            return None
+
+        mapping = getattr(self, "ordering_mapping", None)
```
nas-tabchiche: let's get `ordering_mapping` from `view`. This will be a valuable shortcut, as there would be no more need to subclass `CustomOrderingFilter`.

PR #2658, opened 2025-10-06
feat(dispatcher): handle CUD operations on organization models

nas-tabchiche:
- **add a few events**
- **split create from get_or_create**
- **add some create events**
- **register organization CUD messages**


nas-tabchiche:
Not immediately needed, closing for now.

PR #2659, opened 2025-10-06
feat: Completing about localisation

tchoumi313:


eric-intuitem (changes requested):
The pattern seems OK, but we must translate diskSpace, diskUsed at least in en.json, and to be complete in all languages

tchoumi313:
> The pattern seems OK, but we must translate diskSpace, diskUsed at least in en.json, and to be complete in all languages

Yes, Okay

tchoumi313:
> LGTM. Can be merged once CI is green

I don't understanding why this lint-enterprise test keep failing 
image

PR #2664, opened 2025-10-07
feat: Filter assets by class

tchoumi313:
image

## Summary by CodeRabbit

* **New Features**
  * Added a hierarchical Asset Class filter to the Assets list with autocomplete and multi-select; selecting a class now includes all descendant classes automatically.
  * Asset Class options now return clearer full-path labels (value/label) for easier discovery and selection in filters and dropdowns.

### DIFF
--- a/backend/core/views.py
+++ b/backend/core/views.py
@@ -810,7 +810,7 @@ class AssetViewSet(BaseModelViewSet):
     def type(self, request):
         return Response(dict(Asset.Type.choices))
 
-    @action(detail=False, name="Get asset class choices")
+    @action(detail=False, name="Get asset class choices", url_path="asset-class")
     def asset_class(self, request):
         # this is for filters
         return Response(


backend/core/views.py
nas-tabchiche: The `url_path` is missing here, therefore to get the asset class choices, one would have to query the `/assets/asset_class` endpoint. We wish to use kebab case here: `/assets/asset-class`.

Furthermore, the frontend expects `/assets/asset-class`, so we get a 404 when attempting to get asset class choices.

Network trace from my browser on commit `6d413d6c6970`:
```
Request URL: http://localhost:5173/assets/asset-class
Request Method: GET
Status Code: 404 Not Found
```

nas-tabchiche (changes requested):

backend/core/models.py
```diff
@@ -2423,6 +2423,39 @@ def full_path(self):
         else:
             return f"{self.parent.full_path}/{self.name}"
 
+    def get_descendants(self) -> set["AssetClass"]:
```
ab-smith, line 2426: why are we handling descendants for classes filtering? this seems over engineered a bit

tchoumi313: This is to ensure that all the descendants of a selected class are not excluded from the filtering.
For example, let’s say we have Class A and its subclasses A/B and A/C.
If a user selects Class A, we want to keep all assets belonging to Class A as well as those belonging to its subclasses not just the assets directly associated with Class A

tchoumi313:
We choosed to filter the asset classes by exact match instead of doing it in a tree-descendant way.
This behavior may be implemented later on.

PR #2665, opened 2025-10-07
fix: custom ordering on UserGroups is broken

Axxiar:
## Summary by CodeRabbit

* Refactor
  * Improved sorting for localized items and related user groups to be deterministic and consistent across hierarchies, with better performance on large datasets.
  * Added support for both ascending and descending order when viewing sorted results.
* Chores
  * Added a backend dependency to enable optimized, database-side sorting.


nas-tabchiche:
CTEs are quite the heavy artillery, I'm closing this for now so that we can devise a general pattern to address ordering folder paths.

PR #2769, opened 2025-10-28
feat(ux): The buttons are made visible at the bottom of the page

tchoumi313:
image

## Summary by CodeRabbit

* **Refactor**
  * Reorganized table container to a flex-based layout, improving scrolling and overflow behavior.
  * Moved header, body and footer into new wrapper structure to stabilize vertical layout and event handling.
  * Preserved existing row actions, cell renderings, pagination and public API so visible behaviors remain unchanged.


nas-tabchiche:
We're not there yet. A sticky layout is definitely the way to go here. I'm closing this so that we can start fresh.

PR #2777, opened 2025-10-29
feat: add asset class filter

monsieurswag:
## Release Notes

* **New Features**
  * Added asset class filter to the assets list view for easier filtering and discovery.

* **Improvements**
  * Asset class dropdown now displays only classes with associated assets, streamlining the selection experience.


ab-smith:
duplicate with #2664 ?

PR #2799, opened 2025-10-31
feat(ux): Sort domains in asset graph

monsieurswag:
## Summary by CodeRabbit

* **Bug Fixes**
  * Graph visualization now uses consistent alphabetical ordering for domains, producing stable, deterministic category ordering and indices. Graph nodes and categories display predictably across generations, replacing prior insertion-order variability and improving reliability of graph layouts for users.


eric-intuitem (changes requested):
review coderabbitai comment.

nas-tabchiche (dismissed):
This has a wider blast radius than just the assets graph, as `GraphExplorer` is also used for mapping visualizations. This might break series-category sync since the category can be identified by its index in a series, e.g.
```
[{
  "id": 1,
  "category": 1
}, {
  "id": 2,
  "category": 0
}, {
  "id": 3,
  "category": 1
}, {
  "id": 4,
  "category": 0
}];
```

As such, ordering transformations should be done outside of the `GraphExplorer` component. I suggest just doing it in the backend.

PR #2800, opened 2025-10-31
feat: Add status to risk scenario control list

monsieurswag:
## Summary by CodeRabbit

* **New Features**
  * Risk scenarios tables now include a visible "Status" column alongside name, owner, and ETA for quicker status at-a-glance.

* **Refactor**
  * Table filter controls are no longer forcibly hidden for risk scenario views, restoring the default filter UI so users can apply filters directly.


ab-smith (commented):
we should add the filter as well then,

also make sure that risk scenarios on a risk assessment is not shifted and on the global list, especially when enabling the inherent level is enabled

monsieurswag:
> also make sure that risk scenarios on a risk assessment is not shifted and on the global list, especially when enabling the inherent level is enabled

I didn't see anything weird (even after enabling the inherent risk feature flag) in:
- the risk assessment list view
- the risk scenarios list view
- the risk assessment detail view
- the risk scenario detail view

What would the "shifting" effect be ?

eric-intuitem (approved):
LGTM

PR #2833, opened 2025-11-04
hotfix: Fix applied control import from data-wizard not working

monsieurswag:
## Summary by CodeRabbit

* **Refactor**
  * Improved data handling during applied-control creation to prevent unintended fields from being processed.
* **Bug Fixes**
  * Processing of uploaded Excel now returns actual results instead of an empty payload.
  * Removed an extraneous warning log during data processing.


eric-intuitem (approved):
LGTM

PR #2834, opened 2025-11-04
fix: Cost is displayed as JSON in action plan

monsieurswag:
This only display the build cost (not the run/maintenance cost), is this acceptable ?

## Summary by CodeRabbit

* **New Features**
  * Action Plan PDFs now show cost as a single, human-readable string that combines cost parts (e.g., build/run), omits zero-value segments, and separates parts with " | ".
  * Currency symbols are rendered when available, with graceful fallback to generic formatting if currency is unknown.

* **Bug Fixes**
  * Currency symbol placement standardized to appear after numeric values in executive summaries and totals.

### DIFF
--- a/backend/crq/views.py
+++ b/backend/crq/views.py
@@ -407,13 +407,13 @@ class QuantitativeRiskStudyViewSet(BaseModelViewSet):
                 # Format risk reduction display
                 def format_currency(value):
                     if value >= 1000000000:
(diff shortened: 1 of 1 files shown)


monsieurswag:
> Why not use display_cost?

Fixed: i added the function's  logic in the display_cost function instead.

PR #2886, opened 2025-11-17
feat: add xlsx exports for ALL entities that have CSV exports

tchoumi313:
image

## Summary by CodeRabbit

* **New Features**
  * Added XLSX export across assets, incidents, security exceptions, applied controls, risk scenarios, risk assessments, action plans, and findings/compliance assessments.
  * Per-item XLSX downloads for risk assessments and action plans; frontend routes now stream XLSX files.
  * Export UI updated to present CSV and XLSX choices via a popover/dropdown.

* **Improvements**
  * XLSX exports include Excel-safe escaping, text wrapping, and tuned column widths for clearer, safer spreadsheets.


ab-smith:
can we find a generic pattern with less repition for which we just send the fields and it manages the rest is homogeneous way across models?

tchoumi313:
> can we find a generic pattern with less repition for which we just send the fields and it manages the rest is homogeneous way across models?

Okay thanks, at the frontend level I am using a general pattern for models. I am going to think about doing same for the backend.

PR #2897, opened 2025-11-18
feat: Sync applied controls to their reference control

monsieurswag:
## Summary by CodeRabbit

* **New Features**
  * Reference control updates now propagate to linked applied controls.
  * Per-control toggle to enable/disable synchronization (default: enabled).
  * UI shows sync option when a reference control is present.
* **Behavior Changes**
  * Reference controls that have synced applied controls can no longer be deleted until sync is removed.
* **Localization**
  * Added sync label and help text in 18+ languages.


ab-smith:
Don't think we need a migration or a sophisticated solution for this. We can consider multiple lightweight implementations:
- a button on the reference control to propagate some of the info (ref_id, description, function, category) to the children
- a button on the applied control to resync with the ref control

PR #2905, opened 2025-11-19
feat: Automatically load/update appropriate libs and improve mapping engine invalidation

monsieurswag:
Note for myself: Failing tests has to be fixed:
- The problem can likely be reproduced by creating a new DB from scratch
- The bug is surely caused by importing the `core/signals.py` file which imports the `Framework` model (which may not be created in the database since it may this file may be imported and therefore executed before the Framework table creation migration was applied).

What this PR does:
- Load custom libraries on upload.
- Update custom libraries on updated version upload.
- Automatically update libraries with `autoload=True` at startup.
- Refresh the MappingEngine in-memory requirement mapping set list when a requirement mapping set is added/updated/deleted.
- Refresh the MappingEngine in-memory (frame.urn => framework score) dict when a framework is added/updated/deleted.

Using signals to invalidate the MappingEngine cache gives the ability to the cache to react to `on_delete=models.CASCADE` originated deletions.
It's also more convenient for transaction support as the post_save and post_delete signals are only fired when a transaction executes successfully.
 

## Summary by CodeRabbit

* **New Features**
  * Added automatic update mechanism for auto-loaded libraries to load and refresh library content.

* **Improvements**
  * Enhanced cache invalidation when framework and mapping data changes, improving performance consistency.
  * Improved error handling and logging for library update operations.

* **Changes**
  * Removed framework comparison restriction in library mappings.

### DIFF
--- a/backend/library/management/commands/autoloadlibraries.py
+++ b/backend/library/management/commands/autoloadlibraries.py
@@ -41,4 +41,4 @@ class Command(BaseCommand):
                         error_msg=error_msg,
                     )
         except Exception:
-            pass
+            logger.error("Failed to update libraries", exc_info=True)


monsieurswag:
Clean version of this PR here: https://github.com/intuitem/ciso-assistant-community/pull/3626

PR #2910, opened 2025-11-20
feat: exclude third parties from sso at creation

tchoumi313:
## Summary by CodeRabbit

* **New Features**
  * Allow marking new accounts as third-party.
  * Add third-party checkbox to the user creation form.
  * Frontend schema updated to include an optional third-party flag.

* **Bug Fixes / Behavior Changes**
  * Local-login preference now defaults based on third-party status during creation.
  * Local-login field visibility is conditional on context and third-party selection.
  * Switching an existing account to third-party enforces local-login.


eric-intuitem (changes requested):
This should be a one-liner, that sets keep_local_login=True for third-party account.

frontend/src/lib/components/Forms/ModelForm/UserForm.svelte
```diff
@@ -65,14 +65,28 @@
...
+{/if}
 
-{#if context !== 'create'}
-	<Checkbox
-		{form}
-		field="keep_local_login"
-		label={m.keepLocalLogin()}
-		helpText={m.keepLocalLoginHelpText()}
-	/>
+{#if shape.keep_local_login}
```
nas-tabchiche: checking against `shape` to render a field is a dangerous practice, introspection is tough against zod schemas and must be kept at a bare minimum. All such checks in this diff should be removed before we merge.

frontend/src/lib/components/Forms/ModelForm/UserForm.svelte
```diff
@@ -65,14 +65,28 @@
 {#if shape.is_active}
 	<Checkbox {form} field="is_active" label={m.isActive()} helpText={m.isActiveHelpText()} />
 {/if}
+{#if shape.is_third_party}
```
nas-tabchiche: ditto

nas-tabchiche (changes requested):
Looks functional, checks against `shape` are to be removed before we can merge.

Mohamed-Hacene:
Already implemented by #2933

PR #2923, opened 2025-11-21
fix: evidence label removal bug

tchoumi313:
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures filtering labels are consistently handled during record updates: missing or empty labels from the frontend now clear existing labels rather than being ignored.
  * Fixes form submission edge cases by allowing update requests to be modified when needed so label changes are reliably processed across different form types.


eric-intuitem (dismissed):
This breaks the API tests

PR #2936, opened 2025-11-25
feat: sanity check before deleting items

tchoumi313:
Screenshot from 2025-11-25 14-57-53

## Summary by CodeRabbit

* **New Features**
  * Added cascade-preview endpoint and UI: modals fetch/display grouped, expandable lists of items that would be deleted or lose links; loading/error states and improved accessibility. Prompt modal accepts new props (id, action, body component, debug).

* **Bug Fixes**
  * Updated deletion flow for an assessment to adjust what is removed and added logging around folder handling.

* **Localization**
  * Added cascade-related translations (warning, fetch error, affected notices/hints) across many locales.


ab-smith:
what if we're deleting a compliance-assessment that has +1200 requirements?
will we be able to detect 2nd order relationships? like the one between a compliance-assessment and an applied control?

tchoumi313:
> what if we're deleting a compliance-assessment that has +1200 requirements? will we be able to detect 2nd order relationships? like the one between a compliance-assessment and an applied control?

No we will not see +1200 Requirements

For the moment no we are not detecting  2nd level M2M relationships

PR #2958, opened 2025-11-28
fix: reload mapping engine and frameworks cache on LoadedLibrary save

nas-tabchiche:
## Summary by CodeRabbit

* **Bug Fixes**
  * Saving a library now detects related requirement mappings and frameworks and schedules their loading as background tasks upon commit.
  * This ensures library metadata and related engine data stay synchronized without delaying or blocking the save operation, improving responsiveness and eventual consistency.


Mohamed-Hacene (changes requested):
1. **Load**: Engine is correctly triggered on mapping library import, but the new mappings still don't show up as available in audits.
2. **Unload**: Engine cache is not refreshed when a mapping library is unloaded.

nas-tabchiche:
Closing this in favor of a signals-based approach

PR #3028, opened 2025-12-09
feat: remove useless log

tchoumi313:
Added config to silence the django-structlog’s middleware request

## Summary by CodeRabbit

* **Chores**
  * Added a logging filter to suppress irrelevant 404 entries for requests targeting the /api/build endpoint.
  * Registered and applied the filter to console logs and to file-based logging when enabled.
  * Improves operational log clarity by reducing noise, making logs easier to scan and troubleshoot.


eric-intuitem:
The log does not appear anymore, and we don't want to hide the symptom, but to fix the root cause.

PR #3075, opened 2025-12-19
feat: Make data wizard update objects

monsieurswag:
The update doesn't concern the following objects:
- ComplianceAssessment
- RiskAssessment
- FindingAssessment

These objects are created with the f"{object_type}_{timestamp}" format which is meant to be a unique name generated by the backend (not user-provided and therefore not updatable)


ab-smith:
Fixed in #3384

PR #3083, opened 2025-12-23
fix: cascade deletion of attachment with evidence

melinoix:
**Ciso Assistant have this following issue:**

If we create an evidence, create a revision containing an attachment. We have normally 3 cases:

- Erasing the entire evidence deleting all evidenceRevisions and their attachments
- Erasing the entire Revision and the attachment linked to it
- Only Erasing the attachment in the detail view revision.

With this 3 cases, only one works now which is the last one. The other ones have an issue: not deleting the attachment when deleting the evidence / evidenceRevision object.

The last one is working because it uses backend function delete_attachment and the others are only erasing their object. (Django does NOT delete FileField files when a model instance is deleted)

 Fixed by changing the delete function by first deleting all attachments contained by the models.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of attachments when evidence items are deleted. Associated attachments are now properly removed from the system.

### DIFF
--- a/backend/core/models.py
+++ b/backend/core/models.py
@@ -3762,6 +3762,13 @@ class Evidence(
         super().save(*args, **kwargs)
         self.revisions.update(is_published=self.is_published)
 
+    def delete(self, using=None, keep_parents=False):
+        for rev in self.revisions.all():
+            if rev.attachment:
+                rev.attachment.delete(save=False)
+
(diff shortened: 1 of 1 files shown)


Mohamed-Hacene (approved):
0k

PR #3109, opened 2025-12-29
remove useless ref_id  in Personal Data form

tchoumi313:
Personal data form was having an ID field but the model lacks a matching ref_id field 

## Summary by CodeRabbit

* **Bug Fixes**
  * Reference ID field removed from the personal data form — it will no longer appear when empty.
  * Reference ID no longer shown on asset detail views.
  * Validation/behavior adjusted so omitted Reference ID values are treated consistently (no empty-string default).

* **Tests**
  * Improved toast handling in functional tests to better synchronize visibility checks and reduce flakiness.


nas-tabchiche (changes requested):
Let's just delete the `ref_id` field in `PersonalDataForm.svelte` and its associated schema instead of hiding/commenting it out.

PR #3130, opened 2026-01-03
fix: sqlite instead of pg for backend code coverage

eric-intuitem:
pg is way too slow for this test

## Summary by CodeRabbit

* **Chores**
  * Simplified CI workflow by removing the PostgreSQL dependency so backend tests and coverage run without a local database service.
* **Documentation**
  * Added a short module docstring describing the FilteringLabel model to clarify its purpose.


eric-intuitem:
Almost same speed

PR #3133, opened 2026-01-04
updated ACN framework

f0r7y:
## Release Notes

* **Updates**
  * Updated framework version and publication date to align with latest standards
  * Modified assessability status for multiple security controls across governance, identity, and protection domains
  * Refined control descriptions and policy references for improved clarity and alignment
  * Restructured control relationships and dependencies to reflect updated governance hierarchy


eric-intuitem:
Thanks for the contribution!
As many nodes go from assessable=true to assessable=false, updating the version is not the best option. It seems more relevant to create a new framework, with a different URN, and to provide a mapping between the two frameworks.
With the current approach, assessments done with the current version will be lost after update.

f0r7y:
Hi
This version, in addition to including the updates released a few days ago by ACN, contains the exact number of requirements. In the previous version, the requirements were split by individual request, but the count was incorrect.

I understand what you're saying, but this is the correct version of the framework. The old one doesn't make sense.

eric-intuitem:
@f0r7y to avoid any breaking change, I propose that the new library uses different URNs than the previous one. We can even remove the old yaml file, but let's not break what users may have done with the existing library. I let you determine the best option for the urn change.

eric-intuitem:
The content of this PR has been largely taken in account in the recent update. So we can close it. Thanks @f0r7y for the insight.

PR #3166, opened 2026-01-11
feat(ux): improve donut chart

eric-intuitem:
Show relevant data directly, no need for hover.
This is much simpler for screenshots

## Summary by CodeRabbit

* **New Features**
  * Donut charts now display percentages by default with larger, overflow-safe labels and clearer center emphasis.

* **Bug Fixes / Behavior Changes**
  * Comparison result ordering updated, which may change the order/detection of differences in audit comparisons.

* **Tests**
  * Toast-check helper is now timeout-aware and resilient to missing toasts.

### DIFF
--- a/frontend/tests/utils/base-page.ts
+++ b/frontend/tests/utils/base-page.ts
@@ -71,27 +71,10 @@ export abstract class BasePage {
 		}
 	}
 
-	async isToastVisible(
-		value: string,
-		flags?: string | undefined,
-		options?: { timeout?: number } | undefined
-	) {
+	async isToastVisible(value: string, flags?: string | undefined, options?: {} | undefined) {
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/Chart/DonutChart.svelte
```diff
@@ -50,8 +55,8 @@
 			tooltip: {
 				trigger: 'item',
 				formatter: function (params) {
-					// Return formatted tooltip content with just the name and value
-					return `${params.data.name}: ${params.data.value}`;
```
ab-smith, line 54: why dropping this?

eric-intuitem: Indeed, we can keep the previous layout even if the value is already visible.

PR #3179, opened 2026-01-14
fix: sidebar display issues for custom roles

Mohamed-Hacene:
If a user cannot view global settings, feature flags cannot be fetched, which may cause display issues in sidebar sections governed by feature flags.

## Summary by CodeRabbit

* **New Features**
  * Roles now automatically receive default permissions for folder access and viewing global settings.

* **Improvements**
  * Role creation and updates use bulk operations and transactional updates for better performance and reliability.
  * Non-built-in group names are kept in sync with role assignments.

* **Bug Fixes**
  * Prevented a frontend error when checking compliance assessments during delete-prevention.

* **Chores**
  * Applied a migration to grant the global-settings view permission to existing roles.


backend/iam/migrations/0019_add_view_globalsettings_in_custom_roles.py
```diff
@@ -0,0 +1,27 @@
...
+    Permission = apps.get_model("auth", "Permission")
+
+    permission = Permission.objects.filter(
+        codename="view_globalsettings",
+        content_type__app_label="global_settings",
+    ).first()
+
+    for role in Role.objects.all():
+        if not role.permissions.filter(id=permission.id).exists():
+            role.permissions.add(permission)
```
nas-tabchiche, line 17: in `RoleViewSet._get_default_permissions` we include `view_folder`, but the data migration does not backfill it. Is it intentional?

Mohamed-Hacene: Yes, it has been correctly added since the feature has been released. view_globalsettings was the only one missing.

PR #3180, opened 2026-01-14
fix: change processing.foreignKeys.owner.urlModel from users to entities

nas-tabchiche:
In `privacy.models`, `Processing.owner` is a foreign key to an `Entity`,
not a user.
This fixes a bug where the link to the owner points to /users/<id>
instead of /entities/<id>.
NOTE: We do not display the processing's owner in the UI at the time,
this fix just avoids the footgun.


ab-smith:
The mistake is on the model, not the other way around.
We currently covers the ownership with 'assigned_to' and this is already exposed and used on the form.

PR #3209, opened 2026-01-17
fix(test): fix flakiness caused by some toasts and enterprise settings

eric-intuitem:
1/ Some toasts are flaky, make them optional, while making sure any failure would be detected by the following test (e.g. after changing a password, log in with the new password).
2/ Enterprise settings are loaded asynchronously, so add visibility wait to avoid flakiness.

## Summary by CodeRabbit

* **Tests**
  * Toast helper: added optional mode to avoid failures, safer handling, and richer return when optional.
  * Tests updated to use optional toast checks where toast may be absent.
  * Improved test stability by adding explicit visibility/wait steps and controlled checkbox interactions.


nas-tabchiche (approved):
LGTM

PR #3215, opened 2026-01-19
fix(lib): CIS Controls v8 "reference_control_base_urn" Field

tarkadia:
# Changes

Replaced the packager name `intuitem` with the actual name of the framework packager in the `reference_control_base_urn` field to ensure consistency with the rest of the framework.

## Summary by CodeRabbit

* **Refactor**
  * Control base URN generation now adapts to the configured packager instead of using a fixed reference.
  * Source and target node base URNs are derived dynamically from available framework data rather than constructed statically.

* **Bug Fix**
  * Added validation to ensure frameworks contain requirement nodes; a clear error is raised when nodes are missing.


tarkadia:
It was ultimately decided that the issue was not significant enough to warrant making the change. This will prevent potential bugs for those who have already integrated CIS v8 and wish to update the framework.

PR #3218, opened 2026-01-19
fix(tool): Compute Correct Source & Target Requirement Node Base URNs

tarkadia:
# Changes
Fixed an issue in `prepare_mapping_v2.py` that computed an incorrect `node_base_urn` source and/or target when the packager name wasn't equal to `intuitem`. 

The script was inserting the full URN of a node into the Excel table instead of removing the base URN from the URN.

# Update

The base URN computing system has been modified to take all required node URNs into account for the calculation.

## Summary by CodeRabbit

* **Refactor**
  * Mapping export now derives base URNs dynamically from framework data instead of using static constructions, updating how mapping IDs and sheet entries are generated.
* **Bug Fix**
  * Added validation and explicit warnings for missing or inconsistent framework URNs to surface issues without halting export.

### DIFF
--- a/tools/prepare_mapping_v2.py
+++ b/tools/prepare_mapping_v2.py
@@ -49,7 +49,7 @@ def load_and_validate_yaml(path, label):
 
 
 
-def common_urn_base(req_nodes: List[Dict], *, name: str) -> str:
+def compute_base_urn(req_nodes: List[Dict], *, name: str) -> str:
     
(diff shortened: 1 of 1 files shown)


tarkadia:
The base URN computing system has been modified to take all required node URNs into account for the calculation

eric-intuitem (approved):
LGTM

PR #3224, opened 2026-01-20
fix: make leader field required on team creation

martinzerty:
## Summary by CodeRabbit

* **Refactor**
  * Team records now require a leader to be specified; creating or updating a team without assigning a leader will be rejected.
  * Validation tightened to disallow empty or omitted leader values, so forms and APIs must provide a valid leader identifier.

### DIFF
--- a/frontend/src/lib/utils/schemas.ts
+++ b/frontend/src/lib/utils/schemas.ts
@@ -1412,7 +1412,7 @@ export const teamSchema = z.object({
 	team_email: z.string().email().optional(),
 	folder: z.string(),
 	members: z.array(z.string().uuid().optional()).optional(),
-	leader: z.string(),
+	leader: z.string().uuid(),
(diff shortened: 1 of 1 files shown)


tchoumi313:
Is it not preferable to instead edit the `schemas.ts`? That is changing from `leader: z.string().uuid().optional().nullable(),` to `leader: z.string().uuid(),`

martinzerty:
> Is it not preferable to instead edit the `schemas.ts`? That is changing from `leader: z.string().uuid().optional().nullable(),` to `leader: z.string().uuid(),`

You are probably right ! I'll update it

ab-smith:
> UI looks good, but for the schema to accurately depict what we want to do here, let's remove `nullable` from the `leader` attribute.

the red asterix doesn't show actually if we leave nullable, fixed

PR #3285, opened 2026-01-24
fix: recovery objectives alignment for unset objective

monsieurswag:
When the recovery objective of a primary asset is 0 (meaning it's unset) and the capability of its support asset is greater than 0 the UI display this as an unalignment.
If there's no recovery objective there can't be any unalignment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed verdict calculation to properly handle non-positive values by excluding them from verdict generation, improving data accuracy in edge cases.

### DIFF
--- a/backend/core/models.py
+++ b/backend/core/models.py
@@ -3310,7 +3310,12 @@ class Asset(
             real_value = capabilities.get(key)
 
             verdict = None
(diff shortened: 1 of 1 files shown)


eric-intuitem (approved):
LGTM

PR #3391, opened 2026-02-07
feat: add flag to create IAM groups

eric-intuitem:
IAM groups and role assignments are no longer created for all domains, but only domains that have the corresponding flag set. This will avoid creatings 1000s of groups and role assignments for nothing, and will ease group selection.
During the migration, domains which have not been assigned are flagged to false and corresponding groups and role assignments are erased.

## Summary by CodeRabbit

* **New Features**
  * Added automatic IAM group provisioning for domain folders with enable/disable toggle
  * IAM groups are used to assign roles to users; global folder setting cannot be disabled
  * Added IAM groups status column to folder table view
  * Multi-language support added for IAM group management interface

* **Bug Fixes**
  * Validation prevents disabling IAM groups when users are already assigned to them

### DIFF
--- a/frontend/tests/functional/detailed/settings/general.test.ts
+++ b/frontend/tests/functional/detailed/settings/general.test.ts
@@ -45,9 +45,9 @@ test.describe('General settings', () => {
 	});
 
 	test('risk matrices settings', async ({ riskMatricesPage, settingsPage }) => {
-	await test.step('check default matrix disposition', async () => {
-		await page.getByRole('button', { name: ' Risk matrix settings +' }).click();
-		await expect(page.getByTestId('form-input-risk-matrix-swap-axes')).not.toBeChecked();
+		await test.step('check default matrix disposition', async () => {
+			await page.getByRole('button', { name: ' Risk matrix settings +' }).click();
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/Forms/ModelForm/FolderForm.svelte
```diff
@@ -28,6 +29,16 @@
...
+
+	run(() => {
+		const isEdit = Boolean(object?.id);
+		if (!isEdit && form.data?.create_iam_groups !== true) {
+			form.form.update((currentData) => ({
+				...currentData,
+				create_iam_groups: true
+			}));
+		}
+	});
```
nas-tabchiche: This breaks folder creation (works as expected when it is commented out). What was the intent here?

frontend/src/lib/components/Forms/ModelForm/FolderForm.svelte
```diff
@@ -28,6 +29,16 @@
 		object = {},
 		model
 	}: Props = $props();
+
+	run(() => {
```
nas-tabchiche: `run` is deprecated as per svelte5 https://svelte.dev/docs/svelte/svelte-legacy#run

nas-tabchiche (approved):
LGTM

PR #3396, opened 2026-02-10
fix: align data recipient count in privacy overview

Mohamed-Hacene:
## Summary by CodeRabbit

* **Refactor**
  * Updated internal metrics calculation methodology for processings count and recipients count to improve accuracy and consistency.


ab-smith (approved):
ok we can go with for now and add a more explicit widget for the data transfers later on, just don't leave the comments on the code as we have history if needed

PR #3415, opened 2026-02-11
fix: third party mailing through an actor

Mohamed-Hacene:
## Summary by CodeRabbit

* **Bug Fixes**
  * Email sending is more robust: if a sender/mailer is not available the system logs a warning and skips sending instead of failing.
  * Improved error responses when sending fails and a clear success message when mail is sent.
  * Detects missing mailer configuration and returns a validation warning instead of a silent error.


ab-smith:
@Mohamed-Hacene  check coderabbit feedback please

PR #3452, opened 2026-02-16
fix: remove "save" and "cancel" buttons when markdown in preview mode

martinzerty:
…and not in editor mode"

"

## Summary by CodeRabbit

* **Style**
  * Removed Save and Cancel buttons from the markdown field editor interface; only Edit and Preview buttons remain visible.
  * The Preview button now also saves changes before exiting edit mode so edits are persisted when switching to preview.

### DIFF
--- a/frontend/src/lib/components/Forms/TableMarkdownField.svelte
+++ b/frontend/src/lib/components/Forms/TableMarkdownField.svelte
@@ -33,6 +33,7 @@
 	}
 
 	function preview() {
+		saveChanges();
 		isEditing = false;
 	}
 </script>


martinzerty:
> auto
> 
> The PR has a real data-loss risk.
> 
> The component has two states: Edit mode (textarea) and Preview mode (rendered markdown). The PR removes Save and Cancel from Preview mode, leaving only the Edit button.
> 
> The problem is the preview() function (line 35-37) sets isEditing = false without saving. So the workflow becomes:
> 
>     1. User double-clicks → enters Edit mode
> 
>     2. User types changes → clicks Preview to check rendering
> 
>     3. Now in Preview mode, changes live only in editValue (not saved)
> 
>     4. Before PR: User could click Save directly from here
> 
>     5. After PR: User must click Edit → then Save — an extra round-trip
> 
> 
> The risk: If a user edits text, clicks Preview to verify, and then navigates away (clicks sidebar, closes tab, etc.), their changes are silently lost. Before this PR, Save was accessible from Preview mode, so users could Preview → Save in one step.
> 
> Other observations:
> 
>     * Blast radius: 3 pages affected — auditee assessments, compliance assessments (table mode), and task nodes
> 
>     * Dead code: saveChanges() and cancelEdit() are still defined and used in Edit mode, so no broken references
> 
>     * The preview() function itself is arguably the root issue — it should probably auto-save or at least the Preview button should
>       be a "view-only toggle" rather than a state that can trap unsaved changes
> 
> 
> Recommendation: Rather than just removing the buttons, a safer fix would be either:
> 
>     1. Keep Save in preview mode (only remove Cancel, which indeed makes less sense in preview)
> 
>     2. Or make preview() call saveChanges() before switching to preview mode, so changes are never stranded

I understand the issue, it's indeed a rather big data loss. I think the best option is to get the saveChanges() function called by the preview() function. It's more understandable than having the save button even though the user has edited the markdown textarea

ab-smith:
We're not losing data now but we are not covering the case of user not liking what he/she sees on the preview and we're saving on thier behalf, which is now a pain to go back.
what we really need is fundamentally a visual feedback about the save and cancel.

PR #3515, opened 2026-02-20
fix: disable foreign key constraints during restore

eric-intuitem:
Avoid restoration failure

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of data backup and restoration operations to better handle database integrity constraints and ensure data consistency during critical maintenance workflows.


nas-tabchiche (changes requested):
Both error-recovery paths diagnosed by coderabbit seem legitimately broken to me. Here are its comments:

>Outside diff comments:
>In `@backend/serdes/views.py`:
>- Around line 165-180: The recovery path currently calls
management.call_command("flush") inside the disable_foreign_keys() context but
calls management.call_command("loaddata", "-", format="json", ...) outside it,
so FK checks can break the fallback restore; modify the except block to wrap
both the flush and the subsequent loaddata call in the same
disable_foreign_keys() context (same pattern used in the successful restore
path) so that disable_foreign_keys() surrounds both
management.call_command("flush") and management.call_command("loaddata", "-")
while sys.stdin is set to current_backup.
>- Around line 202-219: The rollback path currently exits the
disable_foreign_keys() context before calling
management.call_command("loaddata", "-") so a FK error during loaddata could
leave the DB flushed; move the loaddata call inside the same
disable_foreign_keys() context used for management.call_command("flush") (i.e.,
ensure both management.call_command("flush", interactive=False) and
management.call_command("loaddata", "-", format="json", verbosity=0) execute
while disable_foreign_keys() is active) and keep the existing exception handling
around that combined block to restore the original backup on any failure.

The patch is small and mechanical:

```python
# In both the except block (~L164) and the license-seat rollback (~L201):
  with disable_foreign_keys():
      management.call_command("flush", interactive=False)
      management.call_command("loaddata", "-", format="json", verbosity=0, exclude=[...])
```

eric-intuitem:
Seems not needed

PR #3525, opened 2026-02-24
fix: suggested items highlighting in autocomplete select

martinzerty:
image

## Summary by CodeRabbit

* **New Features**
  * Applied-controls list now fetches on page load, provides pre-populated options for selection, and refreshes after new controls are added to keep choices up to date.
  * Autocomplete renders only when options are available to ensure stable selection behavior.

* **Style**
  * Suggested options display a blue-indigo translated label when translations are enabled, in both dropdown and selected-item views.


martinzerty:
@coderabbitai review

eric-intuitem (changes requested):
Nice. What we need in addition is to promote/suggest applied controls that match a suggested reference control when selecting existing applied controls.

PR #3530, opened 2026-02-24
fix: preserve MarkdownField value on form submit in preview mode

nas-tabchiche:
When dataType is 'form' (models with attachment field like evidences),
SuperForms serializes from DOM elements. The textarea was removed from
the DOM in preview mode, causing the field value to be lost on submit.
Add a hidden input to retain the value in the DOM during preview.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where markdown field values were not being submitted when preview mode was active. The field data is now properly retained and submitted during preview interactions.


Mohamed-Hacene (approved):
0k

PR #3538, opened 2026-02-24
fix: multiple improvements to solutions management (tprm)

ab-smith:
- **allow solution name to be reused across different entities**
- **prevent translating name in table and detailview**
- **better lookup for solutions**
- **format**

## Release Notes

* **Improvements**
  * Enhanced solution information display across the application to include and properly render additional provider entity details.
  * Corrected field rendering behavior to properly display key identifiers and reference fields without applying unnecessary translation operations.
  * Optimized display handling of long-form content with consistent automatic text truncation applied across all data tables and views.


eric-intuitem (approved):
LGTM

PR #3552, opened 2026-02-25
fix(lib): ANSSI - MonAideCyber Questionnaire

tarkadia:
# UPDATE

In the end, the framework's YAML has been removed

# Important Fix
Fixed a major issue with URNs I didn't notice because of the fact that the `convert_library_v2.py` changed during the building of the framework (when I merged `main` into my branch). It's way better to use the real IDs comming from the official questionnaire rather than arbitrary numbers in URNs (because of how questions are retrieved from the original source). This will ease future updates of the framework.


tarkadia:
@eric-intuitem before publishing a new release, please review and merge this PR

tarkadia:
@eric-intuitem Why removing the `name` column ? It's very important in this questionnaire to have a visual enumeration of questions so that people knows which questions the sub-questions belong to. It's not very handy to answer the questionnaire without that.

On top of that, it will be very annoying for people to find back which question they were treating without having a number they can refer to. They also use question numbers in the official questionnaire, probably for the same reason.

However, I agree that the `ref_id` column is unnecessary.

PR #3572, opened 2026-02-27
test(tooling): add parallel functional test runner with per-test isolation

nas-tabchiche:
Add a new test runner that executes Playwright functional tests in parallel, each in an isolated slot with its own backend, frontend, MailHog, and SQLite database. Output is organized into per-run timestamped directories with per-test-file subdirectories. Uses blob reporter per test and merges into a unified HTML report. Includes summary.json, auto-prune of old runs, `--report` flag to view reports, `--list-runs`/`--list-tests` for discoverability, and `--keep-runs` to control history retention.


nas-tabchiche:
there are simpler approaches, to be resumed at a later time

PR #3584, opened 2026-03-03
fix: correct rbac bypass on get_controls_info function 

melinoix:
The get_controls_info function has a RBAC bypass because of those lines :

`for ac in AppliedControl.objects.all():`
`for ra in RiskAssessment.objects.all():`
`for audit in ComplianceAssessment.objects.all():`

The code had also another problem of a really big complexity causing lots of requests with .distinct and count() that is costful

## Summary by CodeRabbit

* **Bug Fixes**
  * Controls graph now respects folder and permission scope, showing only authorized items and pruning unused audit/risk nodes.
  * Audit and risk counts on controls updated to reflect only referenced/accessible data, reducing noise.

* **Refactor**
  * Data retrieval for controls tightened and optimized to limit payload size and improve performance.

### DIFF
--- a/backend/core/views.py
+++ b/backend/core/views.py
@@ -13059,7 +13059,7 @@ class TaskTemplateViewSet(ExportMixin, BaseModelViewSet):
             if task_identifier in tasks_to_process_ids:
                 processed_tasks_identifiers.add(task_identifier)
 
-                task_template = TaskTemplate.objects.get(id=task_template_id)
+                task_template = self.get_queryset().get(id=task_template_id)
                 try:
                     task_node, created = TaskNode.objects.get_or_create(
                         task_template=task_template,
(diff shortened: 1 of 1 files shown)


backend/core/views.py
```diff
@@ -13036,7 +13059,7 @@ def _parse_due_date(val):
             if task_identifier in tasks_to_process_ids:
                 processed_tasks_identifiers.add(task_identifier)
 
-                task_template = self.get_queryset().get(id=task_template_id)
+                task_template = TaskTemplate.objects.get(id=task_template_id)
```
ab-smith: I believe we've fixed this on a previous PR, this looks like a miss (see the other two)

melinoix: yup its a miss let me correct

PR #3585, opened 2026-03-03
feat(table mode): add annotation visibility for non assessable requirements

tarkadia:
# Why ?

Some frameworks such as **FINMA - Circular 2023/01** (Discussion #2204), **NIS2 - Annex to the Commission Implementing Regulation 2024/2690 with Technical**, **RBI Master Direction 2023** ~~and another framework I'm working on (#3593)~~ have additional info that doesn't really belong in the `Description` field and should be put in the `Annotation` field. The problem is that CISO doesn't show them when we are in "Table Mode", which I think is the only place where it would be relevant to show them

## Summary by CodeRabbit

* **New Features**
  * Enhanced the compliance assessments interface with annotation display for non-assessable requirements. A new toggleable annotation block provides show/hide controls with visual state indicators, and annotation content is rendered with full formatting support.

### DIFF
--- a/frontend/src/routes/(app)/(third-party)/compliance-assessments/[id=uuid]/table-mode/+page.svelte
+++ b/frontend/src/routes/(app)/(third-party)/compliance-assessments/[id=uuid]/table-mode/+page.svelte
@@ -446,6 +446,41 @@
 									<MarkdownRenderer content={requirementAssessment.requirement.description} />
 								</div>
 							{/if}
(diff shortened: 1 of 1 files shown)


ab-smith:
debatable, it could be interesting to pass instructions to third-parties but we're lacking space on this view

tarkadia:
Since this topic requires further thought, I'll convert this PR as Draft

tarkadia:
Will be discussed later

PR #3593, opened 2026-03-03
feat(lib): ITSP.10.171 - Protecting specified information in non-Government of Canada systems and organizations Framework (FR+EN)

tarkadia:
# New Framework

Add of the [Protecting specified information in non-Government of Canada systems and organizations (ITSP.10.171)](https://www.cyber.gc.ca/en/guidance/protecting-specified-information-non-government-canada-systems-and-organizations-itsp10171) Framework (English & [French](https://www.cyber.gc.ca/fr/orientation/protection-linformation-designee-organisations-systemes-ne-relevant-pas-gouvernement-canada-itsp10171#defn-hame%25C3%25A7onnage)).

# Important Notes

~~Some information of the framewok (the 2 footnotes transformed into annotations) is invisible because CISO doesn't display annotations on non assessable requirements (see #3585).~~
Update: The nodes originally having annotations have been removed from the framework, as those nodes ended up being "useless" for us.

In addition, some extra info was not necessary to put into an "Annotation" section, but was still useful for users filling out the audit, so I did some experimenting with CISO to find the best approach. That's why I used special HTML tags such as `<abbr>` and `<sup>` to display a tooltip when hovering over the words. It provides an experience closer to that of the official website.
To do so, I had to modify the `frontend/src/lib/components/MarkdownRenderer.svelte` file by adding the necessary tags and atttributes. Let me know what you think of this approach, I can easily remove those tags if necessary. I do believe these tags could be very useful for some specific frameworks in the future

## Summary by CodeRabbit

* **New Features**
  * Markdown rendering now supports abbreviations and superscripts for improved inline formatting and tooltips.
  * Added a standalone tool to produce a bilingual (FR/EN) ITSP.10.171 Excel workbook with hierarchical requirements, annotations/footnotes, and separate metadata/content sheets.
* **Documentation**
  * ITSP.10.171 added to the list of supported frameworks in the README.


eric-intuitem (approved):
LGTM

PR #3611, opened 2026-03-05
fix(lib): Remove blank Question in CMMC v2 Framework - Node SI.L2-3.14.6

tarkadia:
## Fix

Removal of a useless blank question in requirement `SI.L2 3.14.6` of the **CMMC v2** Framework.
The Excel remains unchanged, except that I added a warning next to the problematic cell, indicating that when exporting the final YAML version, the `question1` element for requirement `SI.L2 3.14.6`, shoud be removed manually. As a result, users who have already filled out a CMMC audit will still have their answers associated with the correct questions.

## Summary by CodeRabbit

* **Updates**
  * Removed two assessment questions from the CMMC 2.0 evaluation framework.
  * Incremented framework version to 6.
  * Updated publication date to March 5, 2026.


eric-intuitem (approved):
LGTM

PR #3626, opened 2026-03-06
fix: proper mapping engine invalidation on mapping library load/unload

nas-tabchiche:
## Summary by CodeRabbit

* **New Features**
  * Mapping cache now loads lazily and exposes a public refresh to initialize or reload mapping data.
  * App startup now registers cache-change handlers so updates are detected automatically.

* **Bug Fixes**
  * Cache refreshes are scheduled after relevant library or framework changes and run after transactions commit.
  * Improved cache consistency and resilience during migrations and model updates.

* **Tests**
  * Marked a database export test as slow for clearer test expectations.


tarkadia (commented):
Works great! The only "issue" is that if we unload a mapping without removing it completely from CISO, the mapping will continue to appear in the suggestions. It's not such a big deal rn since it's rather rare to just unload a mapping, but it might be important to keep this in mind in case of a future update involving mapping unloading

nas-tabchiche:
The mapping engine stores mapping data from stored libraries, so unloading won't flush cache. This has to be addressed (otherwise, why even bother having being able to unload?), but in a separate PR.
If the PR looks good to you otherwise, you can go ahead and accept it

tarkadia (approved):
Okay then, let's fix that unloading "bug" in another PR.
Otherwise, this PR is Good 2 Go!

PR #3648, opened 2026-03-09
fix: Click "about" then "quick start" fails

melinoix:
When the user clicked the about in the footer modal and then clicked on get started, the json modal display had a tricky issue by posting the second modal opening in plain text instead of json which caught the second modal crashing.

fixing it here 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced modal content rendering to safely handle and display structured data with fallback support for unstructured content.


ab-smith (approved):
we should go with a $derived instead of wrapping a run() but it's a good enough defensive approach for now

PR #3654, opened 2026-03-09
fix: invalid dom error in tree map chart component

melinoix:
if you go on http://localhost:5173/analytics/gdpr
you will have in the navigator console this error :

Uncaught (in promise) Error: Initialize failed: invalid dom. at TreemapChart.svelte:116:19 (anonymous) @ TreemapChart.svelte:116

it is because this component does not handle the case where the length of the tree is 0 when you dont have any elements 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the chart component would attempt to initialize with empty data.

### DIFF
--- a/frontend/src/lib/components/Chart/TreemapChart.svelte
+++ b/frontend/src/lib/components/Chart/TreemapChart.svelte
@@ -112,6 +112,7 @@
 	}
 
 	onMount(async () => {
+		if (tree.length === 0) return;
 		echarts = await import('echarts');
 		chart = echarts.init(document.getElementById(chart_id), null, { renderer: 'svg' });
 


melinoix:
the error is obviously here only if you're instance has a length of 0 on the tree

melinoix:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > frontend/src/lib/components/Chart/TreemapChart.svelte (1)> `114-124`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Handle the empty → non-empty transition.**
> > > This guard fixes the initial invalid-DOM case, but it prevents chart initialization when `tree` is empty on first mount and gets populated later. Since `onMount` only runs once and returns early if `tree.length === 0`, the `chart` variable remains null. Later, when the template renders the container due to tree becoming non-empty, the `$effect` checks `if (chart && translatedTree?.length > 0)` but does nothing because `chart` is null—it never gets created.
> > > 
> > > Suggested fix
> > > ```diff
> > >  let chart: any = null;
> > >  let echarts: any = null;
> > > +let chartEl: HTMLDivElement | null = null;
> > > +
> > > +async function ensureChart() {
> > > +	if (chart || tree.length === 0 || !chartEl) return;
> > > +	echarts ??= await import('echarts');
> > > +	chart = echarts.init(chartEl, null, { renderer: 'svg' });
> > > +	window.addEventListener('resize', handleResize);
> > > +}
> > >  
> > >  onMount(async () => {
> > > -	if (tree.length === 0) return;
> > > -	echarts = await import('echarts');
> > > -	chart = echarts.init(document.getElementById(chart_id), null, { renderer: 'svg' });
> > > -
> > > -	if (translatedTree?.length > 0) {
> > > -		chart.setOption(createOption(translatedTree));
> > > -	}
> > > -
> > > -	window.addEventListener('resize', handleResize);
> > > +	await ensureChart();
> > >  });
> > >  
> > >  $effect(() => {
> > > -	if (chart && translatedTree?.length > 0) {
> > > +	void ensureChart();
> > > +	if (chart && translatedTree.length > 0) {
> > >  		chart.setOption(createOption(translatedTree));
> > >  	}
> > >  });
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ```diff
> > > -	<div id={chart_id} class="{width} {height} {classesContainer}"></div>
> > > +	<div bind:this={chartEl} id={chart_id} class="{width} {height} {classesContainer}"></div>
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@frontend/src/lib/components/Chart/TreemapChart.svelte` around lines 114 -
> > > 124, The early return in onMount prevents creating the echarts instance when
> > > tree is empty initially, so remove the guard that returns when tree.length ===
> > > 0; instead always import echarts and initialize chart via
> > > echarts.init(document.getElementById(chart_id), null, { renderer: 'svg' })
> > > inside onMount, then if translatedTree?.length > 0 call
> > > chart.setOption(createOption(translatedTree)); keep adding
> > > window.addEventListener('resize', handleResize) and ensure any existing logic
> > > that updates the chart (the reactive $: block that checks if (chart &&
> > > translatedTree?.length > 0)) will now work because chart is guaranteed to be
> > > initialized even when tree was empty at mount.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> Outside diff comments:
> In `@frontend/src/lib/components/Chart/TreemapChart.svelte`:
> - Around line 114-124: The early return in onMount prevents creating the echarts
> instance when tree is empty initially, so remove the guard that returns when
> tree.length === 0; instead always import echarts and initialize chart via
> echarts.init(document.getElementById(chart_id), null, { renderer: 'svg' })
> inside onMount, then if translatedTree?.length > 0 call
> chart.setOption(createOption(translatedTree)); keep adding
> window.addEventListener('resize', handleResize) and ensure any existing logic
> that updates the chart (the reactive $: block that checks if (chart &&
> translatedTree?.length > 0)) will now work because chart is guaranteed to be
> initialized even when tree was empty at mount.
> ```
> 
> ℹ️ Review info

I dont think it is necessary here as the user quit the page

PR #3656, opened 2026-03-09
feat: add task template import functionality to data wizard and CLI

tchoumi313:
## Summary by CodeRabbit

* **New Features**
  * Import TaskTemplate and associated TaskNode records from multi-sheet Excel or CSV via the Data Wizard UI and backend flow.
  * New "import_tasks" CLI command to run TaskTemplate imports.
  * TaskTemplate added to the data import UI (perimeter selection disabled where not applicable).
  * TaskTemplate export now includes additional schedule fields and a scheduled date.

* **Documentation**
  * Updated docs with TaskTemplate multi-sheet/CSV import instructions, sheet formats, conflict handling, and import behavior notes.


backend/data_wizard/views.py
```diff
@@ -2494,6 +2731,192 @@ def _process_compliance_assessment(
...
+                            due_date=due_date,
+                            status=status_val,
+                            observation=observation,
+                            folder=template.folder,
+                            scheduled_date=scheduled_date,
+                            to_delete=False,
+                        )
+                        overall_results["task_nodes"]["created"] += 1
+                except Exception as exc:
+                    logger.error("Task node import error: %s", exc)
```
monsieurswag: Our current codebase convention forbid returning stringified error message directly to the client.

It seems the only thing that can raise an Exception inside your try/except block is `TaskNode.objects.create`, so just wrap this call in a try/except  block using specific(precise) `Exception` types (not just the base `Exception` type to not catch unexpected exceptions).
Doing so will make it conventionally safe to return a stringified exception.

Example of what i mean:

```py
try:
  TaskNode.objects.create(...)
except (models.SomeDjangoError, TypeError) as e:
   ...
   overall_results["task_nodes"]["errors"].append(
      {"error": str(e), "record": {"due_date": due_date}}
    )
```

backend/data_wizard/views.py
```diff
@@ -2494,6 +2731,192 @@ def _process_compliance_assessment(
...
+            if ref_id:
+                tmpl = TaskTemplate.objects.filter(
+                    ref_id=ref_id, folder_id=row_folder_id
+                ).first()
+            if tmpl is None:
+                tmpl = TaskTemplate.objects.filter(
+                    name=rec_name, folder_id=row_folder_id
+                ).first()
+            if tmpl is None:
+                tmpl = TaskTemplate.objects.filter(name=rec_name).first()
```
monsieurswag: This query may not be selective enough (not filtering by folder could make the code select an unexpected task template (some user may have task templates with the same name across multiple domains))

backend/data_wizard/views.py
```diff
@@ -2494,6 +2731,192 @@ def _process_compliance_assessment(
...
+        overall_results = {
+            "templates": {},
+            "task_nodes": {
+                "created": 0,
+                "updated": 0,
+                "skipped": 0,
+                "failed": 0,
+                "errors": [],
+            },
+        }
```
monsieurswag: Use the `Result` dataclass in `overall_results["task_nodes"]` if this doesn't cause any limitation.

monsieurswag (dismissed):
Add a simple `backend/cli/sample_task_templates.csv` file containing a few examples of CSV import lines for task templates like what you can see in `backend/cli/sample_assets.csv` `backend/cli/sample_controls.csv`.

Retest the data-wizard again once (by importing a TaskTemplate with its TaskNode) to ensure the latest code modifications didn't break anything.

backend/data_wizard/views.py
monsieurswag: I have no idea why github truncated the highlighted code diff to 4 lines.
Consider the code diff starting at the `TaskNode.objects.filter` call.

backend/data_wizard/views.py
monsieurswag: I have no idea why github truncated the highlighted code diff to 4 lines.
Consider the code diff starting at the `overall_results` variable definition line.

nas-tabchiche (changes requested):
`TaskTemplate` import does not seem to work on my end. Here is what I get when attempting to load the sample excel file provided in the docs:

```
Processing file with model: TaskTemplate, folder: 1a566dea-3aa5-44db-9111-5cd7682cf718, perimeter: null, framework: null, matrix: null
Error parsing Excel file
Traceback (most recent call last):
  File "/Users/nassim/Dev/intuitem/ciso-assistant-community/backend/data_wizard/views.py", line 2345, in process_excel_file
    res = self._process_task_template_excel(
        request,
    ...<3 lines>...
        on_conflict,
    )
  File "/Users/nassim/Dev/intuitem/ciso-assistant-community/backend/data_wizard/views.py", line 3114, in _process_task_template_excel
    TaskTemplateRecordConsumer(base_context)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
TypeError: Can't instantiate abstract class TaskTemplateRecordConsumer without an implementation for abstract method 'create_context'
```

backend/data_wizard/views.py
```diff
@@ -1472,6 +1482,209 @@ def prepare_create(
...
+        ref_id = record_data.get("ref_id")
+        if ref_id:
+            existing = TaskTemplate.objects.filter(
+                ref_id=ref_id, folder_id=folder_id
+            ).first()
+            if existing:
+                return existing
+        return TaskTemplate.objects.filter(
+            name=record_data.get("name"), folder_id=folder_id
+        ).first()
```
monsieurswag, line 1701: The `TaskTemplateRecordConsumer.find_existing` seems to assume a `TaskTemplate` object can be identified either by a `(ref_id, folder_id)` combination OR by a `(name, folder_id)` combination, which is a false assumption.

Two `TaskTemplate` can have the same `name` and the same `ref_id` EVEN WHILE having the same `folder_id` (being in the same domain).

That's kind of a problem, why the `TaskTempate` class doesn't have any `fields_to_check` attribute in the first place ?

Maybe we should think about adding a `fields_to_check` attribute to `TaskTemplate`, but if so, we must ensure it doesn't break anything in the instance of a user currently having 2 `TaskTemplate` which will be considered as conflicting after adding a `fields_to_check` field list.

This shouldn't be fixed by this PR but this SHALL be talked with with the product owner (aka Abder).

backend/data_wizard/views.py
```diff
@@ -1472,6 +1482,209 @@ def prepare_create(
...
+                        obj = candidate
+                        break
+            if obj is not None:
+                ids.add(obj.id)
+            else:
+                logger.warning(
+                    "Task import: could not resolve %s '%s'; skipping.",
+                    model._meta.model_name,
+                    entry,
+                )
```
monsieurswag, line 1754: At `backend/data_wizard/views.py:109` why the logger being used is actually the builtin `logging` python module ?
This error originates from way back this PR, still our new code SHALL use the `structlog` logger (`logger = structlog.get_logger(__name__)`) instead of instead of the `logging` logger.

backend/data_wizard/views.py
```diff
@@ -1472,6 +1482,209 @@ def prepare_create(
...
+        for col_name, model in [
+            ("assets", Asset),
+            ("applied_controls", AppliedControl),
+            ("evidences", Evidence),
+            ("compliance_assessments", ComplianceAssessment),
+            ("risk_assessments", RiskAssessment),
+            ("findings_assessment", FindingsAssessment),
+        ]:
+            raw = record.get(col_name) or ""
+            data[col_name] = self._resolve_m2m_by_name(model, raw) if raw else []
```
monsieurswag, line 1860: At `backend/data_wizard/views.py` lines `1674-1681` multiple models are listed.
Among these models only one seem to have a stringifieid (`__str__`) version with this pattern: `{name} - {version}`.

Despite that `self._resolve_m2m_by_name` threat every single model like it has the same pattern.

Meaning that an `AppliedControl` with a `name` field being `"XXX - YYY"` would not be found as the `self._resolve_m2m_by_name` function would look for `AppliedControl.objects.filter(name__iexact="XXX")` instead of `AppliedControl.objects.filter(name__iexact="YYY")` due to `name_part = entry.rsplit(" - ", 1)[0].strip()` at `backend/data_wizard/views.py:1574`.

This edge case SHALL be fixed.

backend/data_wizard/views.py
```diff
@@ -1472,6 +1482,209 @@ def prepare_create(
...
+                    raw_val = str(record.get(col) or "").strip()
+                    if raw_val:
+                        try:
+                            schedule[opt_key] = [
+                                int(v.strip()) for v in raw_val.split(",") if v.strip()
+                            ]
+                        except (ValueError, TypeError):
+                            pass
+            except (ValueError, TypeError):
+                pass
```
monsieurswag: At `backend/data_wizard/views.py` lines `1646` and `1648` woudln't it be wiser to return an error instead of `pass` (like you do at `backend/data_wizard/views.py:1594`) ?

I guess a user would prefer if the operation failed with an error message instead of the `TaskTemplate` being created with an unexpected schedule.

monsieurswag (changes requested):
There seem to be a bug (i'am not 100% sure this is considered as bug, but as someone oblivious with the export feature, it seems like a bug to me, and a lot of oblivious real users will think the same).

Based on `git blame` you're the one who created the `ExportMixin` class so i guess you're the best suited to know if this behavior is "normal" or not.

(Starting with now `TaskTemplate` objects in my database.)

- I do `python clica.py import-tasks --file cli/sample_task_templates.csv`
- I export my TaskTemplate objects via the `button[data-testid="export-button"]` at the `/task-templates` frontend page as `Excel` which makes me download a `TEST.xlsx` file.
- I delete all the `TaskTemplate` objects from my database.
- I do `python clica.py import-tasks --file TEST.xlsx`

```json
{
  "message": "File loaded successfully",
  "results": {
    "templates": {
      "successful": 0,
      "created": 0,
      "updated": 0,
      "skipped": 0,
      "failed": 0,
      "stopped": false,
      "errors": []
    },
    "task_nodes": {
      "successful": 0,
      "created": 0,
      "updated": 0,
      "skipped": 0,
      "failed": 0,
      "stopped": false,
      "errors": []
    }
  }
}
```

Expected behavior: I expected loading the exported `Excel` file would result in my task-templates being loaded back.

---

I didn't review the excel part of the PR btw, so this review is incomplete but there's a descent amount of stuff to work through nonetheless.

martinzerty:
Tested and worked. Looks good to me

martinzerty (dismissed):
LGTM

ab-smith:
closed due to inactivity. to reopen once updated and ready to merge

PR #3684, opened 2026-03-13
fix: crowded threat list on risk scenario edit page

monsieurswag:
There isn't a perfect solution for this ticket, this solution is the one which address the best the problem being described the ticker (without worsening the `AutoCompleteSelect.svelte` component UI/UX).

## Summary by CodeRabbit

* **Style**
  * Updated default spacing for the autocomplete select component to improve visual consistency.
  * Enhanced multiple selection mode with improved layout positioning and scrollable options list display for better usability when handling multiple selections.


ab-smith:
Not enough, we need a better ui pattern for crowded autocomplete fields everywhere

ab-smith:
The need is covered through #3754

PR #3700, opened 2026-03-16
fix: allow third party to access their profile and settings

martinzerty:
The third party representatives were not able to see their profile and access the settings to change password/set TOTP. Now they can.

## Summary by CodeRabbit

* **Bug Fixes**
  * Personal access token creation and deletion now enforce early authorization, preventing third-party users from performing PAT operations (returns 403).

* **Refactor**
  * PAT UI is conditionally shown based on a new patAllowed flag; MFA schema imports consolidated to a shared utility import.

* **Chores**
  * Added validation schemas for TOTP activation (6-digit) and WebAuthn registration.


monsieurswag:
Try to fix the problem by moving `./(app)/(internal)/my-profile/...` stuff to `./(app)/(third-party)/` instead (if it doesn't work/create problems keep the current the solution).

nas-tabchiche (dismissed):
Please fix the ongoing merge conflicts

Mohamed-Hacene (dismissed):
We should hide PAT, it is not meant to be used by third parties

frontend/src/routes/(app)/(third-party)/my-profile/settings/+page.server.ts
```diff
@@ -174,6 +179,10 @@ export const actions: Actions = {
 		return { recoveryCodes: response.data };
 	},
 	createPAT: async (event) => {
+		if (event.locals.user.is_third_party) {
+			return fail(403, { error: 'Forbidden' });
+		}
```
monsieurswag: Returning `return fail(403, { error: 'Forbidden' });` when a third-party attempts to create a PAT is a clear intent this action is meant to be forbidden.

In this case not preventing a third party from creating a PAT using the `/api/iam/auth-tokens/` backend endpoint is effectively a security vulnerability.

We MUST either keep the same `fail(403, { error: 'Forbidden' })` response AND also make `/api/iam/auth-tokens/` refuse to create the PAT and also return a `HTTP_403_FORBIDDEN` status.

OR if third party should be able to create a PAT then all PAT visibility related changes shall be reverted.

As a proof a third-party is still authorized to created a PAT by the backend use this: `curl -s -H "Authorization: token {THIRD_PARTY_USER_TOKEN}" -X POST -d 'name=abcd1234' http://127.0.0.1:8000/api/iam/auth-tokens/`

frontend/src/routes/(app)/(third-party)/my-profile/settings/+page.server.ts
```diff
@@ -11,6 +11,7 @@ import { z } from 'zod';
 import { AuthTokenCreateSchema } from '$lib/utils/schemas';
 
 export const load: PageServerLoad = async (event) => {
+	const PATAllowed = !event.locals.user.is_third_party;
```
monsieurswag: Always declare variables the nearest possible to where they're first used.
At `frontend/src/routes/(app)/(third-party)/my-profile/settings/+page.server.ts:14` `PATAllowed` is declared.

But it only start being used from line `67` (just declare `PATAllowed` right above `let personalAccessTokens` instead).

frontend/src/routes/(app)/(third-party)/my-profile/settings/+page.server.ts
```diff
@@ -174,6 +179,10 @@ export const actions: Actions = {
 		return { recoveryCodes: response.data };
 	},
 	createPAT: async (event) => {
+		if (event.locals.user.is_third_party) {
```
monsieurswag: Replace `if (event.locals.user.is_third_party)` by `if (!PATAllowed)` as it's more expressive (make the intent clearer) AND is DRYer.

frontend/src/routes/(app)/(third-party)/my-profile/settings/+page.server.ts
```diff
@@ -279,6 +288,10 @@ export const actions: Actions = {
 		return { form };
 	},
 	deletePAT: async (event) => {
+		if (event.locals.user.is_third_party) {
```
monsieurswag: Replace `if (event.locals.user.is_third_party)` by `if (!PATAllowed)` as it's more expressive (make the intent clearer) AND is DRYer.

monsieurswag (dismissed):
Also prevent third parties from creating PAT in the backend if we decide to keep the current behavior of the sveltekit action `createPAT` (in `frontend/src/routes/(app)/(third-party)/my-profile/settings/+page.server.ts:182`).

---

Also we should hide the `<button` at `frontend/src/routes/(app)/+layout.svelte:175` for the third party users.
This button invokes a "commandPalette" which mostly contain links to pages that are both irrevelant, useless and really confusing to the third party user (as they are list views empty of any object).

A third party user only have access to a really restricted set of pages anyway, so the sidebar should be enough for him.

Mohamed-Hacene (approved):
0k

PR #3704, opened 2026-03-16
feat: show that the ID will be generated automatically if left blank

martinzerty:
## Summary by CodeRabbit

* **New Features**
  * Added help text to ID fields indicating that IDs will be automatically generated if left blank.

* **Documentation**
  * Added translation strings for the new help text across all supported languages (Arabic, Chinese, Czech, Danish, Dutch, English, French, German, Greek, Hindi, Croatian, Hungarian, Indonesian, Italian, Lithuanian, Polish, Portuguese, Romanian, Swedish, Turkish, Ukrainian, and Urdu).


Mohamed-Hacene:
Adding the generated ID as a placeholder can also be considered but I'm not sure we can retrieve it easily before creation

martinzerty:
> Adding the generated ID as a placeholder can also be considered but I'm not sure we can retrieve it easily before creation

I guess if the user does not want to specify the ID of the object, it is probably because the ID itself is not relevant to him/her.

ab-smith (approved):

PR #3707, opened 2026-03-17
fix(oidc): deep search for email in nested IdP token claims

Mohamed-Hacene:
## Summary by CodeRabbit

**Bug Fixes**
* Improved social login compatibility by enhancing email extraction to recursively search through nested provider data structures, enabling successful authentication with additional identity provider payload formats and reducing edge-case failures.


Mohamed-Hacene:
> it would be safer to have a max depth for this recursion

Done in #3752

PR #3708, opened 2026-03-17
feat(data wizard): add Asset.is_business_function attribute

martinzerty:
… and the imports in the datawizard

## Summary by CodeRabbit

* **New Features**
  * Added support for an "Is a business function ?" field on assets: shown as a new column in asset exports and accepted during asset import/creation (boolean-like handling of user input).
* **Documentation**
  * Updated import guidance to document the new boolean-like input values ("true"/"yes" or "false"/"no") for the asset field.


backend/data_wizard/views.py
```diff
@@ -568,6 +479,19 @@ def prepare_create(
...
+        if isinstance(record_is_business_function, str):
+            is_business_function = record_is_business_function.strip().lower() in [
+                "lower",
+                "yes",
+                "1",
+                "y",
+                "t",
+            ]
+        elif isinstance(record_is_business_function, bool):
+            is_business_function = record_is_business_function
```
monsieurswag, line 624: Only limit the options making `ibf_to_import` truthy to a single possible string (instead of having 5 different strings (currently being: `["lower", "yes", "1", "y", "t"]`))

nas-tabchiche (approved):
LGTM

PR #3714, opened 2026-03-18
feat: when license is expired, the SSO login fails

martinzerty:
The user cannot login using the SSO if the license of the instance is expired.


martinzerty:
I failed to replicate the bug

PR #3724, opened 2026-03-19
feat: add variable to force TLS1.2 for SMTP

eric-intuitem:
Without this variable the negotiation will start in TLS1.3, which seems to block in some contexts.

## Summary by CodeRabbit

* **New Features**
  * Email communications now support optional, environment-configurable enforcement of TLS 1.2 for outbound mail.
  * Email sending has been updated to use a connection-based approach that preserves HTML message content when provided, improving formatting and delivery consistency.

### DIFF
--- a/backend/ciso_assistant/settings.py
+++ b/backend/ciso_assistant/settings.py
@@ -322,7 +322,7 @@ EMAIL_USE_TLS_RESCUE = os.environ.get("EMAIL_USE_TLS_RESCUE", "False").lower() i
     "1",
     "yes",
 )
-EMAIL_FORCE_TLS12 = os.environ.get("EMAIL_FORCE_TLS12", "False").lower() in (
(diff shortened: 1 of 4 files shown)


backend/ciso_assistant/settings.py
```diff
@@ -321,6 +322,21 @@ def set_ciso_assistant_url(_, __, event_dict):
     "1",
     "yes",
 )
+EMAIL_FORCE_TLS12 = os.environ.get("EMAIL_FORCE_TLS12", "False").lower() in (
```
ab-smith: `EMAIL_FORCE_TLS_1_2` is probably more readable

enterprise/backend/enterprise_core/settings.py
```diff
@@ -319,6 +320,21 @@ def set_ciso_assistant_url(_, __, event_dict):
     "1",
     "yes",
 )
+EMAIL_FORCE_TLS12 = os.environ.get("EMAIL_FORCE_TLS12", "False").lower() in (
```
ab-smith: same as above

PR #3725, opened 2026-03-19
fix: make timerange edition stick on metrics widget (backend not implemented yet)

martinzerty:
Fixes a bug: It was not possible to edit the timerange attribute in the widgets in the metrology's dashboard. The time range was always set to "all_range" in the database and the default rendered in the frontend was "Last Day".
Now, the time_range attribute changes are commited to the database and correcly rendered on the frontend.

## Summary by CodeRabbit

* **Refactor**
  * Improved the time-range selector to use local selection state with controlled synchronization for more stable behavior.
  * Reduced unnecessary prop passing between form components to simplify form interactions and reduce unexpected side effects.

* **Bug Fixes**
  * New widget creation now defaults the time range to "Last 30 days" instead of "All time."


ab-smith:
ok to merge this, but it tackles only the form persistence, not the underlying issue of filtering.

ab-smith (approved):

PR #3726, opened 2026-03-19
fix: change 'Tier' to 'Tiers' in french role name

martinzerty:


martinzerty:
Neither @monsieurswag nor me achieved to apply the translation correctly

eric-intuitem:
The fix shall be made in frontend, as we don't want to use django.po anymore. Fix in #3964.

PR #3752, opened 2026-03-23
chore(oidc): handle lists and add max depth to email lookup recursion

Mohamed-Hacene:
Extends #3707 to improve robustness

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced email extraction during social account authentication to properly handle multiple data formats, including lists and nested structures, with improved safeguards and performance.


ab-smith:
ok but the name is now misleading, we could try something like `_find_email_recursive`

PR #3784, opened 2026-03-26
fix: correct applied_control 'impact' attribute for the csv export

martinzerty:
Replace 'impact' attribute of the applied controls by the one in the db called 'control_impact' in the csv export.
Before, we looked for the attribute named 'impact'. However, as it does not exist, the result for this attribute in the csv export was always empty

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the "Impact" column in Action Plan exports (Excel, CSV, XLSX formats) to correctly display impact values for compliance and risk assessments.


backend/core/views.py
Mohamed-Hacene: We have the same issue here

PR #3813, opened 2026-03-29
chores: update helm chart to match v3.15.0

ab-smith:
## Summary by CodeRabbit

* **Chores**
  * Updated application release version to v3.15.0


ab-smith:
Should be better now

PR #3894, opened 2026-04-09
fix: implementation groups for dynamic frameworks

Mohamed-Hacene:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved implementation group filtering for dynamic framework compliance assessments. The system now correctly respects user-selected implementation groups when applying filters, rather than using uniform filtering. This ensures more accurate assessment results and consistent filtering behavior.


eric-intuitem (approved):
LGTM

PR #3900, opened 2026-04-10
feat: reflect client name change in page's title

Axxiar:
## Summary by CodeRabbit

* **New Features**
  * Application title is now configurable via client settings, allowing organizations to set a custom product name.
  * If no custom name is provided, the app falls back to "CISO Assistant" and continues to append the current page title as before.


eric-intuitem (approved):
LGTM

PR #3902, opened 2026-04-10
fix: handle the case of local MFA while SSO is enabled

ab-smith:
- **option 1**
- **option 2**

## Summary by CodeRabbit

* **Improvements**
  * Multi-factor authentication now intelligently adapts based on authentication method. Users signing in through federated identity providers bypass the local MFA challenge, streamlining their login experience. Standard authentication methods continue to enforce multi-factor protection.


eric-intuitem (approved):
LGTM

PR #3923, opened 2026-04-14
fix: framework duplicate fails with UNIQUE constraint on long names

nas-tabchiche:
When duplicating a framework whose slugified name reaches the 60-char
truncation limit, appending " (copy)" to the name produces the same
slug after truncation. This causes the copy's node URNs to collide
with the source's existing URNs in the database.

Detect URN collisions before creating nodes and disambiguate by
appending a numeric suffix (-2, -3, ...) to the slug when needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved framework duplication to guarantee unique URNs (handles long-name slug collisions, copy-of-copy, and truncation boundaries), preserve source node/choice identifiers needed by expressions, remap depends-on references correctly (single/multihop, multi-answer "all", cross-node) while leaving external URNs intact, preserve null URNs, honor folder placement, and avoid copying compliance assessments, attachments, or related artifacts.

* **Tests**
  * Added extensive tests covering duplication outcomes, library-shaped questions, visibility/outcome expression evaluation on copies, depends-on edge cases, empty frameworks, parent-chain repointing, and scalar-field preservation.


nas-tabchiche:
@ab-smith

ab-smith (approved):

PR #3930, opened 2026-04-15
feat: add writable permission handling to folder tree selection

tchoumi313:
- Updated `get_folder_content` to include a `writable_ids` parameter for determining writable folders.
- Enhanced `org_tree` action in `FolderViewSet` to accept an optional `write_perm` query parameter, allowing for dynamic writable folder identification based on user permissions.
- Modified frontend components to support the new writable feature in `FolderTreeSelect`, ensuring nodes are selectable based on user permissions.
- Added writable permissions to various model forms, enabling specific actions like adding assets, incidents, and other entities based on user roles.

Pasted image (2)

Pasted image (3)

## Summary by CodeRabbit

* **New Features**
  * Folder trees now include a writable flag and visually indicate writable vs non-writable folders.
  * Folder pickers and search results respect writability: non-writable items cannot be selected.
  * Non-writable ancestors auto-expand to reveal writable descendants for easier discovery.
  * Forms provide a default write-permission context so folder selectors load with the appropriate write filter.
  * Folder tree loading can be constrained by a write-permission filter to show writable items.


frontend/src/lib/components/Forms/FolderTreeSelect.svelte
```diff
@@ -31,6 +31,13 @@
 		 * previous optionsEndpoint="folders?content_type=DO&content_type=GL".
 		 */
 		contentTypes?: string[];
+		/**
+		 * Django permission codename (e.g. "add_asset"). When set, each tree node
+		 * returned by org_tree carries a ``writable`` boolean; only writable nodes
+		 * are selectable.  Ancestor-only nodes remain visible for navigation but
+		 * are rendered as non-selectable.
+		 */
+		writable?: string;
```
nas-tabchiche: Good PR, however a few things bother me here about the `writable` prop:
- `writable` will clash with the [`writable`](https://svelte.dev/docs/svelte/stores) store from svelte (that one is non negociable)
- that's one more prop to remember setting. we can most likely set `add_${urlModel.name}` as a sensible default here. That, or make `writable` a required prop so that the linter screams when we forget to set it.

nas-tabchiche (changes requested):

frontend/src/lib/components/Forms/FolderTreeSelect.svelte
tchoumi313: Thanks for the review, great point. I removed the writable prop entirely from FolderTreeSelect and now always infer write_perm from ModelForm (add_<model.name>).
I also removed all explicit writable="add_*" usages across forms, so there’s no extra prop to remember and no naming clash with Svelte stores.

frontend/src/lib/components/Forms/ModelForm.svelte
```diff
@@ -144,6 +144,8 @@
 	}: Props = $props();
 
 	const URLModel = model.urlModel as urlModel;
+	const defaultFolderWritePermission = `add_${model.name}`;
```
nas-tabchiche: Something I didn't catch in my first review:
A user with the `change_<model>` permission, but not `add_<model>` won't be able to edit an object from said model anymore. We must discriminate between adding and changing an object here.

nas-tabchiche (changes requested):

frontend/src/lib/components/Forms/ModelForm.svelte
tchoumi313: Good catch!

frontend/src/lib/components/FocusMode/FolderTreeNode.svelte
```diff
@@ -1,10 +1,13 @@
...
+	import { untrack } from 'svelte';
 	import FolderTreeNode from './FolderTreeNode.svelte';
 
 	export interface TreeNode {
 		name: string;
 		uuid: string | null;
 		content_type?: string;
+		/** When false, the node exists in the tree for navigation only and cannot be selected. */
+		writable?: boolean;
 		children?: TreeNode[];
```
monsieurswag: At `frontend/src/lib/components/FocusMode/FolderTreeNode.svelte:10`:

`writable?: boolean` should be changed to `writable: boolean`.
`children?: TreeNode[]` should be changed to `children: TreeNode[]`.

All `TreeNode` objects currently come from the `/api/folders/org_tree` backend endpoint.
And as you can see at `backend/core/views.py:7290` it always define the `"writable"` and `"children"` keys for the `TreeNode` dicts it returns.

Also replace `uuid: string | null` to `uuid: string` as a `next(Folder.objects.filter(...)).id` can't be `None` in the backend.

And replace `content_type?: string` by `content_type: string` as the `"content_type"` key is always defined by the `/api/folders/org_tree` backend endpoint.

backend/core/helpers.py
```diff
@@ -1938,6 +1938,7 @@ def get_folder_content(
     include_enclaves,
     viewable_objects,
     needed_folders,
+    writable_ids=None,
```
monsieurswag: Always add annotations for python function arguments.

Example at `backend/core/helpers.py:1941` `writable_ids=None,` should be `writable_ids: Optional[set[UUID]] = None,`.

backend/core/helpers.py
```diff
@@ -1949,6 +1950,7 @@ def get_folder_content(
                 "name": f.name,
                 "uuid": f.id,
                 "viewable": viewable_objects and f.id in viewable_objects,
+                "writable": writable_ids is None or f.id in writable_ids,
```
monsieurswag: Always try to put the non-default condition first.
Example at `backend/core/helpers.py:1953`:

```py
writable_ids is None or f.id in writable_ids
```

`writable_ids is None` represents the default (`True`) when there's no `writable_ids`, and `f.id in writable_ids` represents the actual meaning of the `"writable"` key.

Using:

```py
f.id in writable_ids if writable_ids is not None else True
```

Would be clearer as it put the meaning `f.id in writable_ids` (the non-default scenario) right at the beginning of the expression instead.

backend/core/views.py
```diff
@@ -7198,7 +7198,10 @@ def retrieve(self, request, *args, **kwargs):
     @action(detail=False, methods=["get"])
     def org_tree(self, request):
         """
-        Returns the tree of domains and perimeters
+        Returns the tree of domains and perimeters.
```
monsieurswag: Always use imperative style (PEP 257 docstrings).
At `backend/core/views.py:7201` use the imperative style `Return` instead of `Returns`.

frontend/src/lib/components/FocusMode/FolderTreeNode.svelte
```diff
@@ -36,10 +39,30 @@
 	});
 
 	const hasChildren = $derived((node.children ?? []).length > 0);
-	let isExpanded = $state(false);
+
+	// Returns true if `n` or any of its descendants is writable.
```
monsieurswag: Always use JSDoc style comment with imperative style.
At `frontend/src/lib/components/FocusMode/FolderTreeNode.svelte:43` replace:

```js
// Returns true if `n` or any of its descendants is writable.
```

By this:

```js
/** Returns true if `n` or any of its descendants is writable. */
```

backend/core/views.py
```diff
@@ -7222,6 +7225,22 @@ def org_tree(self, request):
                 needed_folders.add(current.parent_folder_id)
                 current = Folder.objects.get(pk=current.parent_folder_id)
 
+        # Optional per-node writable annotation
+        write_perm_codename = request.query_params.get("write_perm")
+        writable_ids = None
+        if write_perm_codename:
+            perm = Permission.objects.filter(codename=write_perm_codename).first()
+            if perm:
```
monsieurswag: Always use `if {...} is not None:` instead of `if {...}:` when checking for None as it's more expressive.

Example at `backend/core/views.py:7233` `if perm:` should be `if perm is not None:`

frontend/src/lib/components/FocusMode/FolderTreeNode.svelte
```diff
@@ -36,10 +39,30 @@
...
 	const isSelected = $derived(node.uuid !== null && focusId === String(node.uuid));
-	// Node is selectable if it has a uuid
-	const isSelectable = $derived(!!node.uuid);
+	// Node is selectable if it has a uuid, matches the allowed content types,
+	// and — when a write_perm filter is active.
+	const isSelectable = $derived(
+		!!node.uuid &&
+			(!node.content_type || contentTypes.includes(node.content_type)) &&
+			node.writable !== false
+	);
```
monsieurswag: At `frontend/src/lib/components/FocusMode/FolderTreeNode.svelte:62` checking `!!node.uuid` is useless as uuid is always an uuid string (using `!!string` when `string.length > 0` will always give `true`).

Using `!node.content_type` is also useless as it's also always a non-empty string.
(Even if it could an empty string/undefined it would still be useless as `contentTypes.includes(node.content_type))` would never be true unless `contentTypes` contained undefined/"")

And replace `node.writable !== false` to `node.writable` as it's exactly the same thing (always avoid negation when it can be avoided).

frontend/src/lib/components/FocusMode/FolderTreeNode.svelte
```diff
@@ -36,10 +39,30 @@
 	});
 
 	const hasChildren = $derived((node.children ?? []).length > 0);
-	let isExpanded = $state(false);
+
+	// Returns true if `n` or any of its descendants is writable.
+	function subtreeHasWritable(n: TreeNode): boolean {
+		if (n.writable !== false) return true;
```
monsieurswag: At `frontend/src/lib/components/FocusMode/FolderTreeNode.svelte:45`, replace `(n.writable !== false)` by `(n.writable)` for the same reason mentionned herebefore.

frontend/src/lib/components/FocusMode/FolderTreeNode.svelte
```diff
@@ -36,10 +39,30 @@
...
-	const isSelectable = $derived(!!node.uuid);
+	// Node is selectable if it has a uuid, matches the allowed content types,
+	// and — when a write_perm filter is active.
+	const isSelectable = $derived(
+		!!node.uuid &&
+			(!node.content_type || contentTypes.includes(node.content_type)) &&
+			node.writable !== false
+	);
 	// Whether this node should be shown at all
 	const isVisible = $derived(!node.content_type || contentTypes.includes(node.content_type));
```
monsieurswag: At `frontend/src/lib/components/FocusMode/FolderTreeNode.svelte:67`, remove `!node.content_type` for the same reason mentionned herebefore.

monsieurswag (changes requested):
The PR is good, everything work as expected.
It can be merged once the following has been corrected.
(It may seem like a lot but this is just syntactic stuff).

The PR should still get a quick test from someone else than this PR's author after these quick syntactic changes have been applied.

monsieurswag (approved):
LGTM

eric-intuitem (approved):
LGTM

nas-tabchiche (approved):
LGTM. To merge once CI is green

PR #3936, opened 2026-04-15
feat: sanitize finding export for round trip

Mohamed-Hacene:
Ensure round-trip compatibility between the follow-up Excel export and the data wizard import by exporting the raw status value (instead of its display label) and adding the missing importable columns: filtering_labels, priority, observation, and vulnerabilities.

## Summary by CodeRabbit

* **New Features**
  * Enhanced Excel export: added columns for Priority, Filtering Labels, Vulnerabilities, and Observation.
  * Status column now exports raw status values.
  * Labels and vulnerabilities exported as pipe-separated strings.
  * Observation column supports text wrapping and emits an empty value when not present.


nas-tabchiche (approved):
LGTM

PR #3970, opened 2026-04-18
feat: assisted mapping creation

ab-smith:


ab-smith:
closed for now, we'll revisit later with a different angle.

PR #3989, opened 2026-04-21
feat: add service account in ciso assistant

tchoumi313:
> Remeber to activate the feature flags 

A Service Account (SA) is a User with:
    - is_service_account = True
    - name (slug)
    - auto-generated email: <slug>@serviceaccount.ciso-assistant.com
    - no password (unusable)
Rules
Max 2 active keys per SA at any time (rotation)
Revoke = delete
SA user can be assigned to any UserGroup (built-in or custom)

image

## Testing the SA token in different state(usergroup assigned, is_active)
Pasted image (5)

Pasted image (6)

## Summary by CodeRabbit

* **New Features**
  * Service account management: admins can create/manage API-only service accounts and their keys (create, list, edit, revoke).
  * API key lifecycle: generate one-time tokens with expiry, limit active keys per account, and edit key metadata.
  * UI & i18n: frontend forms, modals, pages and English/French translations for service-account workflows.
  * Feature flag: toggle service accounts on/off.

* **Bug Fixes / Behavior**
  * Service accounts are disallowed from interactive login and password-reset flows.
* **Permissions**
  * Reader role now includes personal access token permissions.


frontend/src/routes/(app)/(internal)/users/+page.svelte
```diff
@@ -0,0 +1,145 @@
...
+
+	<!--Service accounts tab -->
+	{#if user?.is_admin}
+		<Tabs.Content value="service-accounts">
+			<div class="shadow-lg">
+				<ModelTable
+					source={data.saTable ?? { head: {}, body: [], meta: [] }}
+					URLModel="users"
+					deleteForm={data.deleteForm}
+					baseEndpoint="/service-accounts"
```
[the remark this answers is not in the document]
tchoumi313: deleteForm and URLModel=users is correct since an SA is a User

frontend/src/routes/(app)/(internal)/users/+page.server.ts
```diff
@@ -0,0 +1,140 @@
...
+				return fail(res.status, { form });
+			}
+			Object.entries(response).forEach(([key, value]) => {
+				const msg = Array.isArray(value)
+					? value.map((v: string) => safeTranslate(v)).join(', ')
+					: safeTranslate(value as string);
+				(form.errors as Record<string, string[]>)[key] = [msg];
+			});
+			return fail(res.status, { form });
+		}
```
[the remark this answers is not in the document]
tchoumi313: resolve in commit ahead

frontend/src/lib/components/DetailView/DetailView.svelte
```diff
@@ -173,15 +175,21 @@
...
-			ref: CreateModal,
+			ref,
 			props: {
 				form: model.createForm,
 				model: model,
 				debug: false,
 				additionalInitialData: model.initialData,
-				formAction: `/${model.urlModel}?/create`
+				formAction: isCustomComponent ? '?/create' : `/${model.urlModel}?/create`
 			}
```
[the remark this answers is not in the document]
tchoumi313: Resolve in commit ahead

backend/iam/views.py
```diff
@@ -484,3 +506,315 @@ def post(self, request, *args, **kwargs):
...
+
+        from django.db import transaction
+
+        with transaction.atomic():
+            locked_sa = User.objects.select_for_update().get(pk=sa_pk)
+            active_count = PersonalAccessToken.objects.filter(
+                auth_token__user=locked_sa,
+                auth_token__expiry__gt=timezone.now(),
+            ).count()
+            if active_count >= SA_KEY_LIMIT:
```
monsieurswag: The code is currently broken (i can't create an API key) as `backend/iam/views.py:619` attempts to access `SA_KEY_LIMIT` (which is not current defined in this file).

This MUST be fixed by importing `SA_KEY_LIMIT` from wherever it's defined.

backend/iam/serializers.py
```diff
@@ -127,3 +133,159 @@ class PersonalAccessTokenReadSerializer(serializers.ModelSerializer):
...
+    description = serializers.CharField(
+        source="observation", required=False, allow_blank=True
+    )
+
+    class Meta:
+        model = User
+        fields = ["is_active", "expiry_date", "description"]
+
+
+SA_KEY_LIMIT = 2
```
monsieurswag: At `backend/iam/serializers.py:232` the `SA_KEY_LIMIT` constant should be moved into `backend/iam/models.py` `User` class as `SA_KEY_LIMIT: ClassVar[Final[T]] = 2` as it's being shared to multiple files (`backend/iam/serializers.py`, `backend/iam/views.py`) AND as it's a parameter specific to the `User` (service account).

The exact same can be said for the `SA_EMAIL_DOMAIN` constant at `backend/iam/models.py:1581`.

backend/iam/serializers.py
```diff
@@ -127,3 +133,159 @@ class PersonalAccessTokenReadSerializer(serializers.ModelSerializer):
...
+        if User.objects.filter(email__iexact=email).exists():
+            raise serializers.ValidationError(
+                "A service account with this name already exists."
+            )
+        return value
+
+    def create(self, validated_data):
+        slug = validated_data["slug"]
+        email = f"{slug}@{SA_EMAIL_DOMAIN}"
+        from iam.models import Folder
```
monsieurswag: At `backend/iam/serializers.py:166` the `from iam.models import Folder` import should surely be at the top of the file, the same goes for the following (non-exhaustive) list:

Same at `backend/iam/serializers.py:212` with `from django.utils import timezone`.
Same at `backend/iam/serializers.py:512` with `from django.utils import timezone`.
Same at `backend/iam/serializers.py:262` with `from datetime import date`.
Same at `backend/iam/serializers.py:279` with `from django.utils import timezone`.
Same at `backend/iam/views.py:516` with:

```py
from core.pagination import CustomLimitOffsetPagination
from django.db.models import Count, Q
from django.utils import timezone
```

I didn't list them all, so find the remaining function-scopped imports and put them at the top of the file.
(Except if it was meant to solve a ciruclar reference issue in the first place.)

backend/iam/serializers.py
```diff
@@ -127,3 +133,159 @@ class PersonalAccessTokenReadSerializer(serializers.ModelSerializer):
...
+class ServiceAccountCreateSerializer(serializers.Serializer):
+    """Create a new service account. The email is built from the slug."""
+
+    slug = serializers.CharField(max_length=64)
+    description = serializers.CharField(
+        max_length=500, required=False, allow_blank=True, default=""
+    )
+    expiry_date = serializers.DateField(required=False, allow_null=True, default=None)
+
+    def validate_slug(self, value):
```
monsieurswag: At `backend/iam/serializers.py:150` Add annotation for the `value` argument in `def validate_slug(self, value):`

backend/iam/serializers.py
```diff
@@ -127,3 +133,159 @@ class PersonalAccessTokenReadSerializer(serializers.ModelSerializer):
...
+        value = value.lower().strip()
+        if not _SLUG_RE.match(value):
```
monsieurswag: At `backend/iam/serializers.py:152` use `if _SLUG_RE.fullmatch(value) is not None:` instead of `if not _SLUG_RE.match(value):` as direct None check is more expressive and using `fullmatch` makes the intent of validating the WHOLE `value` string more obvious.

(even though it effectively doesn't change anything as the regex already use `^` and `$` to enforce the `fullmatch` behavior).

backend/iam/serializers.py
```diff
@@ -127,3 +133,159 @@ class PersonalAccessTokenReadSerializer(serializers.ModelSerializer):
     class Meta:
         model = PersonalAccessToken
         fields = ["name", "user", "created", "expiry", "digest"]
+
+
+_SLUG_RE = re.compile(r"^[a-z0-9][a-z0-9\-]{0,62}[a-z0-9]$|^[a-z0-9]$")
```
monsieurswag: Store `_SLUG_RE` directly in the `User` class as a `ClassVar[Final[re.Pattern]]` as it's a `User`-specific constant.

backend/iam/serializers.py
```diff
@@ -127,3 +133,159 @@ class PersonalAccessTokenReadSerializer(serializers.ModelSerializer):
...
+            "is_active",
+            "expiry_date",
+            "date_joined",
+            "active_key_count",
+        ]
+
+    def get_slug(self, obj):
+        return obj.email.split("@")[0]
+
+    def get_active_key_count(self, obj):
```
monsieurswag: At `backend/iam/serializers.py` line `206` and `209` add annotations for `obj` and (more importantly) for the return type of both `get_slug` and `get_active_key_count`.

backend/iam/serializers.py
```diff
@@ -127,3 +133,159 @@ class PersonalAccessTokenReadSerializer(serializers.ModelSerializer):
...
+_SLUG_RE = re.compile(r"^[a-z0-9][a-z0-9\-]{0,62}[a-z0-9]$|^[a-z0-9]$")
+
+
+class ServiceAccountCreateSerializer(serializers.Serializer):
+    """Create a new service account. The email is built from the slug."""
+
+    slug = serializers.CharField(max_length=64)
+    description = serializers.CharField(
+        max_length=500, required=False, allow_blank=True, default=""
+    )
```
monsieurswag: At `backend/iam/serializers.py:145` `max_length=500` should be removed as the `observation` `models.TextField` field it maps to doesn't have any `max_length` set.

(mapping at `backend/iam/serializers.py:177)` AND field definition at `backend/iam/models.py:636`).

frontend/messages/en.json
```diff
@@ -4000,6 +4001,20 @@
...
+	"serviceAccounts": "Service accounts",
+	"serviceAccountsFlagDescription": "Enable service accounts to allow API access via long-lived tokens",
+	"createServiceAccount": "Create service account",
+	"serviceAccountIdentifiedAs": "The service account will be identified as",
+	"serviceAccountSlugHelpText": "Lowercase alphanumeric and hyphens (e.g. ci-pipeline)",
+	"serviceAccountDescriptionHelpText": "Optional — what this SA is used for",
+	"noExpiry": "Leave empty for no expiry.",
+	"keyName": "Key name",
+	"keyNameHelpText": "e.g. \"primary\" or \"ci-pipeline\"",
+	"tokenWarning": "Copy this token now — it will not be shown again.",
```
monsieurswag, line 4048: In `frontend/messages` only the `en.json` and `fr.json` files have been modified (as you can see in the current PR diff).
Translations should be done to all the languages of the app instead (not only for french and english).

frontend/src/lib/components/DetailView/DetailView.svelte
```diff
@@ -173,15 +175,21 @@
 		};
 	});
 
-	function modalCreateForm(model: Record<string, any>): void {
+	const customCreateComponents: Record<string, any> = {
+		'service-account-keys': CreateSAKeyModal
+	};
```
monsieurswag: At `frontend/src/lib/components/DetailView/DetailView.svelte:178` avoid using `Record<string, any>` when there's a better alternative.

In the current situation `ComponentType`/`SvelteComponent` types from `svelte` would be better than `any`.

In the same way the `urlModel` type from `frontend/src/lib/utils/types.ts` may be better than `string` for the key type.

frontend/src/lib/utils/schemas.ts
```diff
@@ -1490,6 +1491,47 @@ export const AuthTokenCreateSchema = z.object({
...
+		{ message: 'Expiry date cannot be in the past' }
+	);
+
+export const ServiceAccountCreateSchema = z.object({
+	slug: z
+		.string()
+		.min(1)
+		.max(64)
+		.regex(/^[a-z0-9][a-z0-9-]*[a-z0-9]$|^[a-z0-9]$/, {
+			message: 'Lowercase alphanumeric and hyphens only, no leading/trailing hyphens'
```
monsieurswag: Even though it's not implicitely enforced since there are already a few untranslated validation error messages in `schema.ts`, we should ideally translate the following string:

- At `frontend/src/lib/utils/schemas.ts:1505` `'Expiry date cannot be in the past'`
- At `frontend/src/lib/utils/schemas.ts:1514` `'Lowercase alphanumeric and hyphens only, no leading/trailing hyphens'`.

frontend/src/lib/components/DetailView/DetailView.svelte
```diff
@@ -173,15 +175,21 @@
...
+
+	function modalCreateForm(model: Record<string, any>, field?: ReverseForeignKeyField): void {
```
monsieurswag: In `frontend/src/lib/components/DetailView/DetailView.svelte:182` isn't the `field` argument of `modalCreateForm(model: Record<string, any>, field?: ReverseForeignKeyField)` dead code ?

It seems to be unused, if it's something you forgot to clean, remove it.

frontend/src/routes/(app)/(internal)/users/+page.svelte
```diff
@@ -0,0 +1,150 @@
...
+		<Tabs.Trigger value="users">
+			<i class="fa-solid fa-user mr-2"></i>{m.users()}
+		</Tabs.Trigger>
+		{#if user?.is_admin && serviceAccountsEnabled}
+			<Tabs.Trigger value="service-accounts">
+				<i class="fa-solid fa-robot mr-2"></i>{m.serviceAccounts()}
+			</Tabs.Trigger>
+		{/if}
+		<Tabs.Indicator />
+	</Tabs.List>
```
monsieurswag: In the `/users` frontend page we should not display the `Tab` component UNLESS the user is an admin AND the `service_accounts` feature flag is set.

Right now the "Users" `Tab` is displayed even when it's the only possible `Tab` (which makes it useless and confusing).

Image

Also the `Tab` component should be inside the `ModelTable` itself (like in the old `/libraries` page or like what can be found in `/incidents/{uuid}` frontend pages)

Image

backend/iam/urls.py
```diff
@@ -40,4 +46,35 @@
...
+        "service-account-keys/",
+        ServiceAccountKeyFlatListCreateView.as_view(),
+        name="service-account-keys-flat",
+    ),
+    path(
+        "service-account-keys/<int:pk>/",
+        ServiceAccountKeyFlatDetailView.as_view(),
+        name="service-account-key-flat-detail",
+    ),
 ]
```
monsieurswag, line 68: The backend currently have duplicated endpoints (endpoints/routes with the exact same logic):

- `/api/iam/service-account-keys/` is a duplicate of `/api/iam/service-accounts/<uuid:sa_pk>/keys/`.
- `/api/iam/service-account-keys/<int:pk>` is a duplicate of `/api/iam/service-accounts/<uuid:sa_pk>/keys/<int:pk>`.

This is NOT acceptable, a REST API MUST never have duplicated endpoints.
So we MUST remove either the scoped endpoints OR the flat ones.

monsieurswag, line 68: Having "flat endpoints" in itself is likely to be considered bad REST API architecture.

But in a certain way it's not a dumb idea, our `EbiosRMStudy`-scoped models (e.g. `RoTo`, `StrategicScenario` etc...) have endpoints of the form `/api/{model}/{pk}` so using flat endpoints for service account key would be consistent with what we currently have.

This "flat" routing model is meant to make different models loosely related (not strictly scoped).

This way of thinking kind of goes with the idea that ciso-assistant is just a GRC database where we store a bunch of objects which are loosely related to each other (kind of like API resources in kubernetes).

But service account keys are a part of the IAM system, so it doesn't really need this flexibility.
Service account keys will never be repurposed or be unscoped from the `User`.
For this reason i think we should remove all flat endpoints (`/api/iam/service-account-keys/` and `/api/iam/service-account-keys/<int:pk>/`).

On this specific subject get someone else's opinion too, as my conclusion might not be the right one.

backend/iam/urls.py
```diff
@@ -40,4 +46,35 @@
...
+    path(
+        "service-accounts/<uuid:sa_pk>/keys/<int:key_pk>/",
+        ServiceAccountKeyDetailView.as_view(),
+        name="service-account-key-detail",
+    ),
+    path(
+        "service-account-keys/",
+        ServiceAccountKeyFlatListCreateView.as_view(),
+        name="service-account-keys-flat",
+    ),
```
monsieurswag: The following `APIView` based classes created by this PR should be renamed e.g:

- `ServiceAccountListCreateView` should be renamed to `ServiceAccountListView`
- `ServiceAccountKeyListCreateView` should be renamed to `ServiceAccountKeyListView`
- `ServiceAccountKeyFlatListCreateView` should be renamed to `ServiceAccountKeyFlatListView`

In REST architecture, a list resource endpoint is implied to support `POST` requests for resource creation, so there's no need to specify that in the class name.

backend/iam/serializers.py
```diff
@@ -127,3 +133,159 @@ class PersonalAccessTokenReadSerializer(serializers.ModelSerializer):
...
+
+
+class ServiceAccountKeyCreateSerializer(serializers.Serializer):
+    """Create a new API key for a service account."""
+
+    service_account = serializers.PrimaryKeyRelatedField(
+        queryset=User.objects.filter(is_service_account=True)
+    )
+    name = serializers.CharField(max_length=64)
+    expiry_days = serializers.IntegerField(min_value=1, max_value=365, default=365)
```
monsieurswag: At `backend/iam/serializers.py:259` the `expiry_days` `default` value shouldn't be `365` as it's inconsistent with the `ServiceAccountKeyCreateSchema.expiry_days` `default` value defined at `frontend/src/lib/utils/schemas.ts:1528`.

frontend/src/routes/(app)/(internal)/service-account-keys/+server.ts
```diff
@@ -0,0 +1,22 @@
...
+	const qs = url.searchParams.toString();
+	const endpoint = `${BASE_API_URL}/${model.endpointUrl}/${qs ? '?' + qs : ''}`;
+
+	const res = await fetch(endpoint);
+	if (!res.ok) {
+		const body = await res.json().catch(() => ({ detail: 'Request failed' }));
+		error(res.status as NumericRange<400, 599>, body);
+	}
+
+	const data = await res.json().catch(() => ({}));
```
monsieurswag: At `frontend/src/routes/(app)/(internal)/service-account-keys/+server.ts:17` change `await res.json().catch(() => ({}));` to `await res.json();` as unhandled errors should never be silenced.
(replacing an invalid JSON body by `{}` is obviously not proper error handling).

frontend/src/routes/(app)/(internal)/users/+page.svelte
```diff
@@ -0,0 +1,150 @@
...
+	// Tab state (persisted in URL)
+	const VALID_TABS = ['users', 'service-accounts'] as const;
+	const rawTab = page.url.searchParams.get('tab') ?? 'users';
+	let tab = $state(
+		user?.is_admin &&
+			serviceAccountsEnabled &&
+			VALID_TABS.includes(rawTab as (typeof VALID_TABS)[number])
+			? rawTab
+			: 'users'
+	);
```
monsieurswag: At `frontend/src/routes/(app)/(internal)/users/+page.svelte:34` the `tab` variable should be defined using the `$derived` rune as its definition is relative to `user` which is itself a `$derived` rune variable.

monsieurswag (changes requested):

monsieurswag (changes requested):
- In the file `frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/+page.svelte` the whole "set password" HTML code MUST be hidden for service account (as it leads to the `/users/{user_id}/edit/set-password` page which is broken for service accounts).

- And fix the new conflicts

PR #4001, opened 2026-04-22
fix(compliance): show ancestors whose implementation groups dont't match selected ones but whose descendants do

### LINKED ISSUE #3985: Framework Builder: Implementation Groups are empty in 27001 audit
**Describe the bug**
I tried to cluster the ISO27001 controls by adding additional Implementation groups. e.g. i added P1 and added it to 5.1,5.2,5.3,5.4,5.5,5.6,5.8 and 5.24 (8 Controls)

When i start an Audit i could select P1 as implementation group, the dashboard shows 8 requirements but under "Associated Requirements" it shows 0 requirements, 
Table mode and Flash mode both show the controls i tagged with P1

**To Reproduce**
1. Create a copy of ISO 27001 in the Framework builder
2. Add an implementation group P1
3. Activate P1 on several controls of the annex
4. Create a new Audit with target framework the copy of 27001 and implementation group P1
5. See thatthe controls are missing on the main audit screen while visible on Table and Falsh mode
**Expected behavior**
Controls visible in the tree view below the audit dashboard 

**Screenshots**

Image

Image

 - Device: Laptop
 - OS: Windows
 - Browser Firefox
 - Version 149.0.2

**Additional context**
Ciso Assistant 3.15.9 Community Edition

nas-tabchiche:
## Summary
- Tree view under Associated Requirements showed 0 requirements when a user filtered an audit by a custom implementation group that only appeared on leaves, if an ancestor node carried a different (pre-existing) IG. Table mode and Flash mode were unaffected because they filter at the RA level and skip the hierarchical filter.
- `filter_graph_by_implementation_groups` short-circuited on a node's own IGs: if the node had IGs of its own and none matched, the entire subtree was pruned — even when descendants did match.
- Fix keeps an ancestor whenever its own IGs match *or* any of its children survived filtering. Purely additive: never keeps fewer nodes than before.

## Reproduction (the reported case)
1. Copy ISO 27001:2022. Both top-level nodes (`core`, `annex-a`) carry pre-tagged IGs (`['Clauses']`, `['SoA']`).
2. In the framework builder, add a custom IG `P1` and tag some `A.5.x` leaves.
3. Publish, create an audit filtered by `P1`.
4. Before the fix: Associated Requirements shows 0; Table mode / Flash mode show the tagged leaves.
5. After the fix: all three views agree.

## Test plan
- [x] New unit tests in `backend/core/tests/test_helpers.py` cover the regression and neighbouring edge cases (5 tests, all passing).
- [x] Full `test_helpers.py` suite still green (38/38).
- [x] Manual: publish a framework with a custom IG tagged on leaves that live under an ancestor with a different IG; audit filtered by the custom IG shows the expected leaves.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixes implementation-group filtering so ancestor nodes are preserved when descendants match selected groups; nodes without specified implementation groups are treated as empty and handled correctly.
* **Tests**
  * Added tests covering filtering behavior: empty selection, ancestor retention, sibling removal, branch pruning, and handling of missing/empty implementation groups.


backend/core/helpers.py
```diff
@@ -489,11 +489,17 @@ def filter_graph_by_implementation_groups(
...
+        # Include a node if:
+        #   - it has any matching IG of its own, OR
+        #   - any of its children survived the filter (i.e., this node is the
+        #     ancestor of a match).
+        # The prior implementation short-circuited on node_groups and dropped
+        # ancestors whose own IGs didn't match, hiding matching descendants —
+        # e.g., ISO 27001's annex-a (IGs=['SoA']) would hide leaves tagged with
+        # a custom IG added via the framework builder.
+        if any(group in node_groups for group in implementation_groups):
+            return True
```
nas-tabchiche, line 503: actual application diff is these 2 lines

ab-smith:
isn't this the expected behaviour? during authoring, we put the IG on the assessable nodes not their parents

tarkadia (commented):
Works great! Howerver, if a parent node is assessable and you select an IG that belongs to the child only and not to the parent, CISO throws a 404 Error when clicking on the parent.

You can reproduce the issue with the following Excel file, by selecting the `Group 2` IG and clicking on the ` A - Level 1` node (a 404 Error also occurs if the parent node doesn't have any defined IGs) : 
[problem.xlsx](https://github.com/user-attachments/files/27058043/problem.xlsx)

Maybe we could fix this in another PR ?

## UPDATE
Fixed in PR #4063

tarkadia:
> isn't this the expected behaviour? during authoring, we put the IG on the assessable nodes not their parents

I think this becomes an issue if we encounter a specific framework where parents are assessable and have different IGs from their children. We do have frameworks with assessable parents, but we didn't encouter this specific issue with IGs until it was reported.

**EDIT :** We DID encounter this issue in the past, but we were using workaround consisting of adding the children's IGs to the parents' IGs

tarkadia (approved):
Good 2 Go

PR #4002, opened 2026-04-22
fix(perf): "Summary" page performance boost

tarkadia:
## Fix

Fixed an issue where the `Summary` section was very slow (it was taking 2 minutes to load 81 audits on my end).
The frontend now makes only 1 call to the backend instead of `1 + F + 2A` (**F**: Amount of folders ; **A**: Amount of audits) in the previous version.
A new dedicated backend endpoint provides minimum info required for the frontend to display the summary.

The following lines of codes in the backend boosted performance :
```py
assessments = list(
    # Reuse get_queryset() so the endpoint keeps the same visibility rules and
    # request-level filtering as the rest of the ComplianceAssessment API.
    self.get_queryset()
    .select_related("folder", "framework")
    .prefetch_related(
        Prefetch(
            "requirement_assessments",
            queryset=RequirementAssessment.objects.filter(
                requirement__assessable=True
            ).select_related("requirement"),
        )
    )
    # We need requirement on each requirement_assessment because the recap
    # calculations inspect requirement.implementation_groups.
    # Keep the response ordering stable so the page does not need to sort again.
    .order_by(Lower("folder__name"), Lower("name"))
)
```

Also, the `Summary` page now loads data asynchronously to ensure that the page has been clicked on.

## Summary by CodeRabbit

* **New Features**
  * Consolidated recap endpoint for single-request recap data
  * Folder-level aggregates and updated donut/percentage visuals from consolidated counts

* **Refactor**
  * Simplified recap page loading to stream folders and reduce rendering complexity
  * Scoring and aggregation logic revised for more accurate NA/score handling

* **Bug Fixes**
  * Improved error handling and empty-state fallback when recap load fails

### DIFF
--- a/backend/core/views.py
+++ b/backend/core/views.py
@@ -10293,17 +10293,11 @@ class ComplianceAssessmentViewSet(BaseModelViewSet):
 
     @action(detail=False, methods=["get"], url_path="recap")
     def recap(self, request):
-        """Return the compact data needed by the /recap page in a single response."""
-        # /recap only renders folder groupings plus per-assessment donut/global scores.
-        # Building that payload here avoids the former frontend request fan-out:
+        """Return the raw data needed by the /recap page in a single response."""
+        # Keep this endpoint presentation-agnostic: it returns counts and scores, while
+        # the Svelte frontend is responsible for colors, percentages, and chart wiring.
+        # This still avoids the former frontend request fan-out:
         # folders -> assessments per folder -> donut/global_score per assessment.
-        result_color_map = {
-            RequirementAssessment.Result.NOT_ASSESSED: "#d1d5db",
-            RequirementAssessment.Result.NON_COMPLIANT: "#f87171",
-            RequirementAssessment.Result.PARTIALLY_COMPLIANT: "#fde047",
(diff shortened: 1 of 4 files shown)


backend/core/views.py
```diff
@@ -10291,6 +10291,155 @@ def get_queryset(self):
     def status(self, request):
         return Response(dict(ComplianceAssessment.Status.choices))
 
+    @action(detail=False, methods=["get"], url_path="recap")
```
ab-smith, line 10295: move this (recap) to a helper file

nas-tabchiche:
Thanks @tarkadia for tackling this. PR makes good perf gains (~7× on a 1k-audit dataset (200k requirement assessments), single round-trip instead of `1 + len(folders) + 2*len(audits)`).

We're going to land #4003 instead. Side-by-side they're close, but #4003 is ~25% faster end-to-end (~10s vs ~13s on the same dataset on M2 macbook pro) because it skips the heavy get_queryset() annotations the recap doesn't need;

Closing in favor of #4003. The streamed {#await} loading state from your frontend is a nice touch though. We should fold that on top.

PR #4014, opened 2026-04-23
perf(compliance-assessments): replace per-row progress walk with aggregate

tarkadia:
## Why 

The goal is to reduce the cost of calculating `progress`, especially during batch operations such as builtin metrics backfilling.

## Changes

In `backend/core/models.py`, the `ComplianceAssessment.progress` property was refactored to use a faster helper: `_get_progress_counts()`.

Before this change, `progress` was using `get_requirement_assessments(include_non_assessable=False)`, which loads a large amount of related data we don't need. It was prefetching Questions, Answers, Choices, Evidences, Applied Controls, and other related objects, even though `progress` only needs to know 3 things :

- How many assessable requirements exist
- Are there any Implementation Groups and have any of them been selected
- How many requirements are already considered assessed

The new implementation keeps the same logic, but computes way faster than before (~1 second instead of almost 1 minute for 81 audits on my end).

There are now 2 different ways to compute progress:
- If no Implementation Groups are selected, it uses a direct database aggregation to compute `total` and `assessed`
- If Implementation Groups are selected, it loads only the minimal fields needed for filtering and counting, then applies the group intersection in Python

## Summary by CodeRabbit

* **Refactor**
  * Improved compliance assessment progress calculation: progress percentages now come from optimized counts for more accurate and consistent reporting.
  * Reduced memory use when filtering by implementation groups by streaming and limiting fields, avoiding loading full records.
  * Better scalability and faster responses for large datasets and filtered views.


backend/core/models.py
```diff
@@ -7841,20 +7841,68 @@ def assign_attributes(target, attributes):
...
+                continue
+
+            total += 1
+            # Keep the existing semantics: a requirement counts as "assessed"
+            # as soon as it has either a non-default result or an explicit score.
+            if (
+                requirement_assessment.result
+                != RequirementAssessment.Result.NOT_ASSESSED
+            ) or requirement_assessment.score is not None:
+                assessed += 1
```
monsieurswag, line 7948: The `) or requirement_assessment.score is not None:` should be replaced by `) and requirement_assessment.score is not None:` At `backend/core/models.py:7895`:

```py
  if (
      requirement_assessment.result
      != RequirementAssessment.Result.NOT_ASSESSED
  ) or requirement_assessment.score is not None:
      assessed += 1
```

With the current code a requirement with a `None` `score` OR with a `NOT_ASSESSED` `result` would still be considered as assessed (would still increment `assessed`).

backend/core/models.py
```diff
@@ -7841,20 +7841,68 @@ def assign_attributes(target, attributes):
...
+            # counters directly without hydrating the requirement tree.
+            counts = requirements.aggregate(
+                total=Count("id"),
+                assessed=Count(
+                    "id",
+                    filter=~Q(result=RequirementAssessment.Result.NOT_ASSESSED)
+                    | Q(score__isnull=False),
+                ),
+            )
+            return counts["total"] or 0, counts["assessed"] or 0
```
monsieurswag: At `backend/core/models.py:7866` the `or 0` is useless as `Count` will never return a non-int value.

backend/core/models.py
```diff
@@ -7841,20 +7841,68 @@ def assign_attributes(target, attributes):
...
+            )
+            .iterator()
         )
+
+        for requirement_assessment in lightweight_requirements:
+            requirement_groups = set(
+                requirement_assessment.requirement.implementation_groups or []
+            )
+            if not (selected_groups & requirement_groups):
+                continue
```
monsieurswag: At `backend/core/models.py:7889` prefer using `not selected_groups.isdisjoint(requirement_groups)` as it's more expressive.

backend/core/models.py
```diff
@@ -7841,20 +7841,68 @@ def assign_attributes(target, attributes):
...
-    def progress(self) -> int:
-        requirement_assessments = list(
-            self.get_requirement_assessments(include_non_assessable=False)
+    def _get_progress_counts(self) -> tuple[int, int]:
+        """
+        Return (total, assessed) counts for assessable requirements.
+
+        This keeps the progress calculation lightweight by avoiding the
+        heavy prefetch graph used to render a full assessment tree.
+        """
```
monsieurswag, line 7905: The `This keeps the progress calculation lightweight by avoiding` message shoudln't be in the docstring.

Read the `sorted` builtin function docstring, it doesn't contain something like `this keeps the sorting RAM usage low by avoiding copying the entire iterable content in memory when sorting the elements.`.

Specifying what the function doesn't do (or avoid) in a docstring is bad.

And specifying it's "lightweight"/"performant" is useless as it should be the norm.
However the opposite could make sense, as a function's consumer should be warned if the function was slow.

backend/core/models.py
```diff
@@ -7841,20 +7841,68 @@ def assign_attributes(target, attributes):
...
+            )
+            return counts["total"] or 0, counts["assessed"] or 0
+
+        selected_groups = set(self.selected_implementation_groups)
+        total = 0
+        assessed = 0
+        lightweight_requirements = (
+            # IG membership lives in a JSONField. To stay DB-portable, stream only
+            # the small set of fields needed for the progress calculation and
+            # apply the set-intersection in Python.
```
monsieurswag: The code is already clear and self-describing, remove this comment.
Only write comments if the intent/logic of your code is likely to be too hard to understand/confusing.

backend/core/models.py
```diff
@@ -7841,20 +7841,68 @@ def assign_attributes(target, attributes):
...
+        for requirement_assessment in lightweight_requirements:
+            requirement_groups = set(
+                requirement_assessment.requirement.implementation_groups or []
+            )
+            if not (selected_groups & requirement_groups):
+                continue
+
+            total += 1
+            # Keep the existing semantics: a requirement counts as "assessed"
+            # as soon as it has either a non-default result or an explicit score.
```
monsieurswag: The code is already clear and self-describing, remove this comment.
Only write comments if the intent/logic of your code is likely to be too hard to understand/confusing.

backend/core/models.py
```diff
@@ -7841,20 +7841,68 @@ def assign_attributes(target, attributes):
...
-            [
-                r
-                for r in requirement_assessments
-                if (r.result != RequirementAssessment.Result.NOT_ASSESSED)
-                or r.score != None
-            ]
+
+        if not self.selected_implementation_groups:
+            # Fast path: when no IG filter is active, the DB can compute both
+            # counters directly without hydrating the requirement tree.
```
monsieurswag: The code is already clear and self-describing, remove this comment.
Only write comments if the intent/logic of your code is likely to be too hard to understand/confusing.

monsieurswag (dismissed):
There is a small logic bug in the "slow path"  `assessed` increment condition which MUST be fixed.

Also remove the comments, the code is already really clean, the comments just make it less readable.

Once these are fixed, this is ready to be merged.

backend/core/models.py
tarkadia: I think you meant `selected_groups.isdisjoint(requirement_groups)` instead of `not selected_groups.isdisjoint(requirement_groups)`, as the **`not`** would mean that if some requirement nodes contains one or more IGs from the selected IGs, the progress will NOT be computed and will be equal to 0.

I tested it, and indeed, putting the `not` returns an incorrect progress for audits with selected IGs.

backend/core/models.py
tarkadia: I think it should be left as is, because sometimes, audits can be assessed by just giving a score to the requirement nodes without giving a specific result on them. Putting `and` would change the original logic. The previous version of the code was like this, with an **OR** :
```py
def progress(self) -> int:
    requirement_assessments = list(
        self.get_requirement_assessments(include_non_assessable=False)
    )
    total_cnt = len(requirement_assessments)
    assessed_cnt = len(
        [
            r
            for r in requirement_assessments
            if (r.result != RequirementAssessment.Result.NOT_ASSESSED)
            or r.score != None
        ]
    )
    return int((assessed_cnt / total_cnt) * 100) if total_cnt > 0 else 0
```

We could've put `and` if the code was written like this:
```py
if (
    requirement_assessment.result
    == RequirementAssessment.Result.NOT_ASSESSED
) and requirement_assessment.score is None:
    continue

assessed += 1
```

tarkadia: @ab-smith can you confirm this?

nas-tabchiche (approved):
LGTM

PR #4016, opened 2026-04-23
fix: enhance autocompleteselect function

tchoumi313:
### Root cause: 

AutocompleteSelect stores its cached value as an array even for single-select fields. In JavaScript, an empty `array [] is truthy`, so the expression `formDataCache['ro_to_couple'] || initialData.ro_to_couple` was always resolved to the (empty) cache array instead of the actual initial value coming from the object data. This caused the feared-events API to be called with an empty ro_to_couples filter, returning no matching option for the feared event, and ultimately silently clearing the field on save.

### Fix: 
Introduced a `toSingleId() normalizer` that collapses array/string/null cache values down to a single scalar ID. The {#key} block and the ro_to_couples filter parameter now use this normalized value, so the component only re-renders and filters when a real ID is present.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the autocomplete field to properly handle empty selections, improving cache reset behavior for more reliable form state management.


ab-smith (commented):
@tchoumi313 are we going to face this issue later for other case where autocomplete is a single choice? why not fix on the component level?

I've approuved it but maybe we can look into this lead if we can have a surgical fix there instead

ab-smith (approved):

tchoumi313:
> @tchoumi313 are we going to face this issue later for other case where autocomplete is a single choice? why not fix on the component level?
> 
> I've approuved it but maybe we can look into this lead if we can have a surgical fix there instead

Yes , you are right, that's what I am doing. I should have marked the pr as draft.

PR #4017, opened 2026-04-23
fix(lib): solve several issues with Framework-Nazionale-C-DP

eric-intuitem:
GV.PO-01.1: regroup subnodes
ID.AM-02: add missing IGs I,E
PR.AA-04: remove IGs
PR.AA-05 and subnodes: add IGs I,E
RS.MA-01.1: regroup subnodes
RC.RP-03.1: rename as rc.co-03.1

## Summary by CodeRabbit

* **New Features**
  * Framework Nacional v2 released with restructured requirement entries and updated implementation group assignments.
  * Added a new recovery communication procedure and streamlined existing recovery nodes.

* **Documentation**
  * Expanded and reformatted descriptions with additional compliance reference links and a Version 2 note.
  * Bumped library version and publication date to reflect the update.


tarkadia (changes requested):
You're removing nodes that were previously assessable by grouping them into a single node (for GV.PO-01.1 and RS.MA-01.1). People who have already created an audit with this framework will lose information.
Also, I don’t really understand why you renamed RC.RP-03.1 to RC.CO-03.1 when it made sense in the original (see image)?

Image

And why remove the IG from PR.AA-04? I don’t understand why that should be the case either

tarkadia (approved):
After some discussion and a minor fix, everything should be fine now. We just need to warn users that some assessable requirement nodes have been merged with their parents nodes

PR #4027, opened 2026-04-24
feat(lib): enrich NIST CSF 2.0 with recommendations

ab-smith:
## Summary by CodeRabbit

* **Updates**
  * NIST CSF 2.0 library bumped to version 6 and enriched with comprehensive reference-control mappings across requirements and functions.

* **Behavior Changes**
  * Score-boundary detection now only considers min/max values, preventing unnecessary compliance assessment updates when only scoring definitions change.

* **Chores**
  * Added a dependency entry linking the library to the control-policy reference set.

### DIFF
--- a/backend/core/models.py
+++ b/backend/core/models.py
@@ -714,6 +714,7 @@ class StoredLibrary(LibraryMixin):
             library_label.garbage_collect()
 
 
+
 class LibraryUpdater:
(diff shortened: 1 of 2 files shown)


eric-intuitem (approved):
LGTM with small fix in the engine to avoid false positive warning.

PR #4063, opened 2026-04-29
fix: 404 error when an assessable parent node doesn't have IG while its child does

monsieurswag:
(A video presenting this bug can be found in jira `CA-1664`).

WITH:
- `COMPLIANCE_ASSESSMENT`  being a `ComplianceAssessment` object
- `FRAMEWORK = compliance_assessment.framework`
- `IGD = FRAMEWORK.implementation_groups_definition`
- `GROUP_NAME = IGD[...]["name"]` (where `IGD is not None`)
- `PARENT_REQ` is a `RequirementNode` with `GROUP_NAME not in `PARENT_REQ.implementations_groups` (where `PARENT_REQ.implementations_groups is not None`)
- `CHILD_REQ` is a `RequirementNode` with `GROUP_NAME in REQ.impementations_groups`  (where `CHILD_REQ.implementations_groups is not None`)
- `PARENT` is a `RequirementAssessment` with `PARENT.requirement == PARENT_REQ` and `PARENT.compliance_assessment == COMPLIANCE_ASSESSMENT`
- `CHILD` is a `RequirementAssessment` with `CHILD.requirement == CHILD_REQ` and `CHILD.compliance_assessment == COMPLIANCE_ASSESSMENT`

In the `/compliance-assessments/{uuid}` frontend page (representing the `COMPLIANCE_ASSESSMENT`, if `GROUP_NAME in COMPLIANCE_ASESSMENT.selected_implementation_groups` (if the group is selected).

Then clicking on the `PARENT_REQ` link in the requirement (in `RecursiveTreeView` at the bottom of the page) will go to a page with a 404 error.

This is due to the `<Anchor` at `frontend/src/routes/(app)/(third-party)/compliance-assessments/[id=uuid]/TreeViewItemContent.svelte:205` having an invalid URL in its `href` prop because of an empty `ra_id`.

The `ra_id` value comes from the  `/api/compliance-assessments/{uuid}` backend endpoint (implemented by the `ComplianceAssessmentViewSet.tree` method).

The reason comes from the `requirement_assessments` definition at line `11485` not containing all the assessable `RequirementAssessment` (only the ones matching the `COMPLIANCE_ASSESSMENT.selected_implementation_groups`).

This cause `get_sorted_requirement_nodes` at line `11526` to not be able to resolve the `PARENT` `"ra_id"` value which leads to this bug in the frontend.

The solution is to implement ALL the assessable requirements by setting `skip_ig_filter=True` so that `"ra_id"` will never fail to be resolved.

It seems like the right solution as the `requirements_assessed` argument name of `get_sorted_requirement_nodes` in `backend/core/helpers.py` implies that the argument should represent an exhaustive list of all assessable requirements.

## Summary by CodeRabbit

* **Bug Fixes**
  * Made the compliance assessment tree endpoint consistently apply requirement-assessment filtering, improving consistency in how assessment data is retrieved and displayed across different framework configurations and selections.


tarkadia (approved):
Good 2 Go

PR #4077, opened 2026-05-05
spec: iam evolution for dynamic groups support and is_published deprecation

eric-intuitem:
- Support for SKIM
- Allow users creation in any folder
- Support "all users" paradigm
- Publication mechanism managed as folder level instead of object level


ab-smith:
we can re-open it when ready

PR #4081, opened 2026-05-05
fix: ebios workshop 3 sort by entity name is broken

Axxiar:
Fixes the sort by entity in Ebios RM, workshop 3, activity 1
Introduces a new `CustomOrderingFilter` that goes with an `ordering_mapping` field that can map the api call name to it's proper model field before applying the backend ordering. Useful for ordering on more complex fields. 

## Summary by CodeRabbit

* **New Features**
  * Enhanced search functionality for stakeholders by entity and category names
  * Added support for sorting/ordering by mapped fields (e.g., folder names)

* **Improvements**
  * User group names now display complete folder paths without truncation


Axxiar:
This approach will be abandoned, traceability problem. The suggested approach would be to keep a custom OrderingFilter that checks for foreign keys if a field `name` exists (fallback to `email` since `User()` doesn't have a `name` field). If so, apply ordering on `foreignkey__name`.

PR #4086, opened 2026-05-07
fix: delete evidence revision attachments when deleting a domain

monsieurswag:
# The problem

If i create an evidence `EVIDENCE` (via the `/evidences` frontend page) with:
- An `attachment` file `evidence.png`
- A `domain` named `DOMAIN`

The attachment file will be stored under `backend/db/attachments`.
If i delete the domain `DOMAIN`, the `EVIDENCE` evidence will be deleted (as expected), but the (now orphaned) `backend/db/attachments/evidence.png` file will remain on the filesystem.

# What this PR does

This PR fixes this by creating a new `_delete_evidence_revision_attachment` signal which is triggered for each `EvidendeRevision` delete (`pre_delete`) operation.
This signal detect the `EvidenceRevision` deletion and remove its attachment from the filesystem accordingly, making it impossible for orphaned files to remain.

When the domain is deleted:

- The domain delete its `Evidence` objects
- The `Evidence` delete its `EvidenceRevision` objects
- The herebefore mentionned signal delete the `EvidenceRevision` attachments.

## Summary by CodeRabbit

* **Refactor**
  * Centralized attachment cleanup for evidence revisions by moving deletion handling to app startup–registered signal handlers.

* **Bug Fix**
  * Deletion proceeds safely when attachment removal fails; failures are logged without interrupting deletion.

* **Behavior**
  * Filename retrieval now returns an empty string when no attachment is present to avoid errors.


backend/core/models.py
```diff
@@ -4517,9 +4511,19 @@ def get_size(self):
...
+    if instance.attachment and instance.attachment.name:
+        try:
+            instance.attachment.delete(save=False)
+        except Exception as e:
+            logger.warning(
+                "Failed to delete evidence revision attachment",
+                revision_id=instance.pk,
+                evidence_name=instance.evidence.name,
+                error=str(e),
+            )
```
nas-tabchiche: let's move this to a `signals.py` file in `backend/core`

nas-tabchiche (dismissed):

backend/core/models.py
monsieurswag: I did this as the deletion logic of a model is usually stored in `models.py`.
But based on django documentation it appears storing this in `signals.py` is indeed more idiomatic, so i will do it.

backend/core/signals.py
```diff
@@ -0,0 +1,21 @@
...
+@receiver(pre_delete, sender=EvidenceRevision)
+def _delete_evidence_revision_attachment(sender, instance: EvidenceRevision, **kwargs):
+    if instance.attachment and instance.attachment.name:
+        try:
+            instance.attachment.delete(save=False)
+        except Exception as e:
+            logger.warning(
+                "Failed to delete evidence revision attachment",
+                revision_id=instance.pk,
+                evidence_name=instance.evidence_id,
```
tchoumi313: The log field is named evidence_name but is assigned instance.evidence_id (the FK value, not a name). The key is misleading for whoever reads the logs.
```suggestion
                  evidence_id=instance.evidence_id,
```

tchoumi313 (commented):
no regression test ??

backend/core/models.py
```diff
@@ -4663,13 +4659,9 @@ def save(self, *args, **kwargs):
...
         super().save(*args, **kwargs)
 
-    def delete(self, using=None, keep_parents=False):
-        if self.attachment:
-            self.attachment.delete(save=False)
-
-        return super().delete(using=using, keep_parents=keep_parents)
-
-    def filename(self):
+    def filename(self) -> str:
```
tchoumi313: The new guard if self.attachment is None never evaluates true. A Django FileField with no file returns a FieldFile None.  It's not actively crashing today only because every current caller (serdes/views.py:676,778, core/views.py:11073) guards attachment first, may break if any future unguarded caller.

```suggestion
  def filename(self) -> str:
      if not self.attachment:
          return ""
      return os.path.basename(self.attachment.name)

```

backend/core/signals.py
monsieurswag: Fixed it.

backend/core/models.py
monsieurswag: I fixed it the way you described.

Even though keeping `if not attachment` isn't REQUIRED (as `self.attachment.name` is `""` and `os.path.basename("")` would return `""` anyway, which make the if condition logically useless).

We should still keep the if condition to make the code clearer the idea that a "Non file field" would have a `name` attribute being an empty string isn't intuitive).

monsieurswag:
> no regression test ??

The probability of this code doing any kind of regression is low.
And this is just a bugfixing PR i want to keep it simple.

And adding regression tests to this isn't trivial as it touch to the filesystem, and some people may store their evidence attachments into S3 buckets.

tchoumi313 (dismissed):
LGTM, can be merge when green

eric-intuitem (approved):
LGTM now

eric-intuitem (approved):
LGTM

PR #4139, opened 2026-05-15
feat(ux): add library update buttons

monsieurswag:
# What it does:

- Create a dedicated `LibraryUpdateButton.svelte` as this PR introduce the need for library update button reusability.
- Display the `LibraryUpdateButton` component on the `/compliance-assessments/{uuid}` frontend page when the library to which this audit's framework belongs to can be updated.
- Use the `LibraryUpdateButton` component to update library library from the `/libraries` list view frontend page.

# Why:

This PR originates from the `CA-1609` Jira ticket.
This ticket originally asks to add this button in the `/frameworks/{uuid}` frontend page, but this has already be done by the #4175 PR.
This PR is still relevant as it fix the same problem for the audit detail view (`/compliance-assessments/{uuid}`) frontend page.

# Note for the reviewer:

The `LibraryUpdateButton.svelte` code as is just a pure copy of the button we previously had within the `LibraryActions.svelte` component.
Your only task is to ensure this PR works as expected and achieve its objective well enough.

## Summary by CodeRabbit

* **Improvements**
  * Compliance assessment API responses now include library details within framework data.
  * Library update UI redesigned: new dedicated update button, clearer loading states, conflict resolution via choice modal, and automatic table refresh after updates.


ab-smith:
@eric-intuitem this is a bit risky to make it too accessible at the audit's level. closing for now and we can revisit this.

PR #2736, opened 2025-10-21
feat: JIRA integration

nas-tabchiche:
## Summary by CodeRabbit

* **New Features**
  * Integration management system enabling connection to external systems like Jira.
  * Jira ITSM integration for automatic synchronization of controls with remote issues.
  * Test connection functionality and webhook support for bidirectional updates.
  * Manage integration configurations and sync mappings through new settings interface.

PR #4147, opened 2026-05-16
fix(core): sync builtin role permissions after every post_migrate emission

nas-tabchiche:
## Problem

On a freshly migrated DB, the built-in `BI-RL-ADM` role (and the other `BI-RL-*` roles) silently end up missing permissions for any app that's positioned **after `core`** in `INSTALLED_APPS`. Today that's `integrations`, `webhooks`, and any new app added below `core`. The visible symptom is that a brand-new superuser cannot use those features: e.g. the Jira / ServiceNow integration pages return an empty config list because the user lacks `view_integrationconfiguration`. Running `manage.py migrate` a second time (no schema changes) "fixes" it.

## Root cause

Django emits `post_migrate` once per app, **in `INSTALLED_APPS` order**. `django.contrib.auth.management.create_permissions` runs per emission and materializes that app's `add_/change_/delete_/view_<model>` rows in `auth_permission`.

`core/apps.py` connects `startup()` with `sender=self`, so `startup()` runs exactly once: when `post_migrate` reaches `core` (position 388 in `INSTALLED_APPS`). `integrations` is at position 393, **after** `core`. By the time `startup()` filters:

```python
administrator_permissions = Permission.objects.filter(codename__in=ADMINISTRATOR_PERMISSIONS_LIST)
…
administrator.permissions.set(administrator_permissions)
```

The four `*_integrationconfiguration` codenames don't exist in `auth_permission` yet, so the role is persisted with them missing. The second `migrate` re-fires `post_migrate`, the permissions exist by then, and the role gets the full set.

This issue has been latent since the original Jira integration PR (#2736) added the integration codenames to `ADMINISTRATOR_PERMISSIONS_LIST` without changing the post_migrate registration.

## Fix

Register a small idempotent `sync_builtin_role_permissions()` handler with **no `sender` filter**. It fires after every app's `post_migrate` emission; the *last* invocation (after the final app in `INSTALLED_APPS` has created its permissions) leaves the built-in roles in their correct final state. The handler is cheap (`get_or_create` + `set`) and guards on the root folder's existence so it bails safely on the very first emissions before `iam` has migrated.

## Reproduction

On `main` before this fix:

```bash
# 1. Wipe DB (including WAL/SHM files; SQLite preserves uncommitted state across just-the-main-file deletes)
rm -f backend/db/ciso-assistant.sqlite3 backend/db/ciso-assistant.sqlite3-{wal,shm}

# 2. Migrate from scratch
cd backend && poetry run python manage.py migrate

# 3. Create a superuser
poetry run python manage.py createsuperuser

# 4. Inspect the admin role
poetry run python manage.py shell -c "
from iam.models import Role
admin = Role.objects.get(name='BI-RL-ADM')
print(sorted(c for c in admin.permissions.values_list('codename', flat=True) if 'integration' in c))
"
```

Expected output:
```
['add_integrationconfiguration', 'change_integrationconfiguration',
 'delete_integrationconfiguration', 'view_integrationconfiguration']
```

Without this fix: `[]`.

Confirming the bug another way: login as the new superuser and visit `/settings/integrations/jira`. You'll see an empty form even after saving; the saved row exists in DB but the user can't read it back. A second `manage.py migrate` makes it work.

## Summary by CodeRabbit

* **Chores**
  * Startup now runs once after migrations to initialize root folder, main record, and mappings; built‑in roles and permissions are synchronized in a single mapping pass.
  * Simplified role assignment flow to improve consistency for audit and third‑party respondent roles.
  * Chat knowledge‑graph prewarm failures are now logged at debug level to aid diagnostics without noisy output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/intuitem/ciso-assistant-community/pull/4147?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ab-smith:
@eric-intuitem ?

eric-intuitem (approved):
LGTM (with Nassim)

PR #4148, opened 2026-05-16
feat(integrations): dynamic field/value mappings for Jira with auto-suggested defaults

nas-tabchiche:
## What this is

The Jira integration page has a UI for picking which AppliedControl fields map to which Jira fields, and for status/priority how local values map to Jira values. This brings Jira in line with the ServiceNow integration, which already uses the same `FieldMapper` component. On table pick, the backend proposes defaults filtered against what the chosen project + issue type actually exposes, so most users won't need to touch the rows manually.

Also fixes an `AutocompleteSelect` bug where mounting with non-empty options and a pre-set form value silently wrote `null` back to the form. The bug was latent across the codebase but only visible here because the field/value mapping rows are the rare case of options-passed-as-prop combined with a saved form value at mount.

## Backend

- `BaseFieldMapper.suggest_mapping_for_table(table_name, client)`. Default returns `{field_map: {}, value_map: {}}` so providers opt in.
- `JiraFieldMapper.suggest_mapping_for_table` walks `_DEFAULT_FIELD_MAP` and keeps only entries whose remote field exists in `client.get_table_columns(table_name)`. For status/priority it then calls `client.get_field_choices` and keeps only the value pairs whose label exists in Jira (case-insensitive match, canonical casing preserved).
- New `suggest_mapping` RPC action on Jira and ServiceNow orchestrators. ServiceNow returns the base no-op so the call doesn't 400.
- `IntegrationConfigurationViewSet.create/update` log the serializer errors when a save is rejected, with `provider_id` and `config_id`.

## Frontend

- `FieldMapper.svelte`: on table pick, fetch columns and `suggest_mapping` in parallel, pre-warm `get_choices` for each suggested value field, then set `fieldMap`, `valueMap`, `columns`, and the form in one render. `onMapsChange` callback so the suggested maps land in the superform store.
- `fetchRpc` checks `res.ok`, parses the backend's `error` / `message`, shows a toast, and returns `null` on failure. Callers default with `?? []` so existing array consumers still work.
- `AutocompleteSelect.svelte`: `selected` is seeded from `\$value` at script init via `initialSelection()`. Stops the "selected → value" `\$effect` from interpreting empty `selected` against a populated form value as a user-clear and writing `null` back on mount.
- Jira and ServiceNow `+page.svelte`: pass `initialConfig={page.data.config.settings}` so the table picker, field map and value map populate from saved settings on page load. Save-failure flash now includes the backend's error detail instead of a flat "Failed to save".

## Recommended merge order

#4147 (post_migrate role-permission sync) is a separate fix for an old bug spotted while working on this branch: on a freshly migrated DB the admin role doesn't get `view_integrationconfiguration` (and other perms for apps loaded after `core` in `INSTALLED_APPS`) until a second `manage.py migrate` runs. Production isn't affected because the backend pod runs `migrate` on every start, but on a fresh dev DB the user who just created an integration can't see it back on reload until they migrate twice. This branch works standalone; recommend merging #4147 first so the fresh-DB dev experience is correct.

## How to test

1. `Settings → Integrations → Jira`. Fill in the Jira server URL, email and API token.
2. Pick a project / issue type from the target table dropdown.
3. Confirm the field and value mapping rows pre-populate. Adjust any that need overriding.
4. Save.

Full setup walkthrough in the docs: https://intuitem.gitbook.io/ciso-assistant/integration/third-party-integrations/jira

## Summary by CodeRabbit

* **New Features**
  * Field mapping UI for Jira & ServiceNow with configurable field_map/value_map and composite table selector.
  * Interactive RPCs to list tables/columns/choices and a new "suggest_mapping" action; base integration supports mapping suggestions.
  * Frontend FieldMapper improvements: choice preloading, configurable labels/help text, toast-backed RPC helper, and synchronized form wiring.

* **Bug Fixes**
  * Improved config error handling to surface validation details and log warnings.
  * Preserved initial selection and capped rendered autocomplete options; ServiceNow table list excludes internal/system tables.

* **Tests**
  * Added tests for dynamic mappings, partial updates, legacy fallbacks, case-insensitive status matching, and description handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/intuitem/ciso-assistant-community/pull/4148?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Mohamed-Hacene (dismissed):
- [x] **1 , field_map "all-or-nothing" fallback** , `backend/integrations/itsm/jira/mapper.py:67-78`
      As soon as the user saves a value_map, the entire default field_map is dropped. If the saved field_map is partial, `name → summary` is lost and Jira rejects the issue creation.
      _Reproduced in testing: `JiraError 400 "summary required"` on creation._

- [x] **2 , Jira Server/DC compatibility broken** , `backend/integrations/itsm/jira/client.py:275-278, 313-316`
      `get_field_choices` and `_get_priority_choices` reach into the SDK's private `self.jira._session` and hardcode `/rest/api/3/...`. Jira Server/DC uses `/rest/api/2/`, so these calls 404 on any non-Cloud instance.
 _Static analysis only as my test instance is Cloud, but any pre-9.x Server/DC instance (no v3) would return 404._

- [x] **3 , `get_table_columns` swallows errors** , `backend/integrations/itsm/jira/client.py:215-217`
      The createmeta exception is silently caught and the function returns only the synthetic column list. An expired token, missing scope, or insufficient permission shows up as "1-row mapper" with no error surfaced to the UI (RPC still returns 200).

- [x] **4 , config schema has no required field** , `backend/integrations/itsm/jira/integration.py:371` + `frontend/.../jira/+page.server.ts:24-26`
      `"required": ["project_key"]` was removed with no replacement, and the zod schema marks the target fields as `optional()`. A configuration can be saved without `table_name`, deferring the crash to the first sync instead of blocking at save time.

- [x] **5 , pull regression on `Lowest` / `Highest` priority** , `backend/integrations/itsm/jira/mapper.py:26-31`
      The previous `PRIORITY_MAP_FROM_JIRA` mapped both `Low` and `Lowest` to 4 (and `High`/`Highest` to 2/1). The new default only has `"4": "Low"`, and the auto-inverted `value_map_to_local` loses the alias. Jira issues with `Lowest` or `Highest` priority are ignored on pull.
      _Not reproduced yet, found via static analysis._

- [x] **6 , ServiceNow: changing table wipes the user mapping** , `frontend/src/lib/components/Forms/FieldMapper.svelte:163-196`
      `BaseFieldMapper.suggest_mapping_for_table` returns `{}` for ServiceNow. `handleTableChange` resets `fieldMap`/`valueMap` before applying the suggestion, so the empty result silently clears the user's existing mapping every time a table is picked.
      _Not reproduced yet, found via static analysis._

frontend/messages/en.json
Axxiar, line 1: what about other locales

Axxiar (changes requested):
- [x] Creating a new applied control fails when adding a description 
(`jira.exceptions.JIRAError: JiraError HTTP 400` `text: Operation value must be a string`)
- [x] Updating a Jira issue name doesn't reflect on linked applied control
the webhook is correctly fired and return 200, but nothing happens
- [x]  Updating a Jira issue status doesn't reflect on linked applied control
- [ ] Deleting a Jira issue does not update the status of the link to "unsynced" on CA
- [x] Trying to remove the selected Target Table from the AutocompleteSelect (see below) freezes the page 
Image

Axxiar (dismissed):
- [ ] Changing a mapped field from Jira issue overwrite the ID with empty array `[]` on CA (only when mapping ref_id to Labels)
Image

- [x] Degraded field is not mapped

Axxiar (dismissed):
One more thing : 
- [x] When connecting an existing applied control to an existing issue, all remote-objects are listed, even those from other issue types. (eg. i get Epics even though i set Tasks as target)

Looks good for the rest

Axxiar (approved):
Okay, looks good

PR #4149, opened 2026-05-17
feat(ai): audit prefill agentic workflow

ab-smith:


ab-smith:
we will revisit later

PR #4155, opened 2026-05-18
fix(backend+ui): blank fields in Matrix Editor after Matrix duplication & YAML Import + fix page title

tarkadia:
## Return & Process missing Fields in Request for Duplicated Matrix & YAML Matrix Import

The `description`, `provider`, and `locale` fields weren't returned by the backend. This caused a blank description on duplication, but it would reappear after a page refresh.
This would also mess up the default locale if the matrix was in a language other than English.
The missing `provider` field doesn't seem to be a big deal as we apparently can't modify its value in the editor, but it's still important to retrieve it from the backend.

## Fix Page Title
The page title would change to `--` after a few seconds being of the page because it was defined in `+page.svelte` instead of `+page.server.svelte`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Draft metadata is now preserved when cloning or importing a risk matrix — description, provider, and locale are correctly returned and applied to the new draft immediately.

* **Improvements**
  * Matrix editor page now includes a localized page title and ensures newly created drafts load with their provider and locale for consistent editing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/intuitem/ciso-assistant-community/pull/4155?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/frontend/src/routes/(app)/(internal)/experimental/matrix-editor/+page.svelte
+++ b/frontend/src/routes/(app)/(internal)/experimental/matrix-editor/+page.svelte
@@ -301,6 +301,9 @@
 				loadDraft({
 					id: result.id,
 					name: result.name,
+					description: result.description,
+					provider: result.provider,
+					locale: result.locale,
 					editing_draft: result.editing_draft
 				});
 				refreshDrafts();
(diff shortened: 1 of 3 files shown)


frontend/src/routes/(app)/(internal)/experimental/matrix-editor/+page.svelte
nas-tabchiche: This is no longer used

nas-tabchiche (approved):
LGTM

PR #4160, opened 2026-05-19
feat(data-wizard): normalize headers

martinzerty:
This PR lowers the header of the input file in the data wizard.

The user can now input a file in the data wizard with for instance : " NaMe    " and the data_wizard will still get this header as the "name" attribute of the wanted object to be imported.

# How to test
- get a test excel file to import in the data_wizard
- change a column name by changing lower case characters to upper case and add one or multiple spaces before and after the attribute
- insert the file in the data wizard

## Summary by CodeRabbit

* **New Features**
  * File imports now automatically normalize column headers by trimming whitespace and converting to lowercase.

* **Bug Fixes**
  * Files with duplicate column names after normalization are detected and rejected, returning a clear error response.

* **Tests**
  * Added tests covering header normalization, numeric-to-string conversion, idempotence, and duplicate-detection behavior.


nas-tabchiche (changes requested):
functional looks okay, but we need at least a minimal test harness before we can merge on `main`

nas-tabchiche (changes requested):
`tests.py` filename does not follow the convention in place. File must be named `test_*.py` to be picked up by pytest https://docs.pytest.org/en/stable/explanation/goodpractices.html#conventions-for-python-test-discovery

tchoumi313 (dismissed):
LGTM, can be merge when green

monsieurswag:
I did some change:

- Created `backend/data_wizard/tests/test_utils.py` (to fix the failing test)
- Refactored the `NormalizeDfColumnsTest` test class.

I did the changes myself instead of waiting for the author to implement them as the PR is already considered "LGTM" by a reviewer.
Waiting for the author to do the changes could slow down the merge too much, which might result in a "stale" PR.

This PR SHALL still be considered as ready-for-merge once the CI is green.

# Reason for these changes

## Adding __init__.py

The failing test comes from a module name conflict with the "test_utils" module name (the conflict is between `backend/data_wizard/tests/test_utils.py` and `backend/app_tests/api/test_utils.py`).

To avoid such conflicts we need to create `__init__.py` files to directories containing tests as described in this [pytest documentation](https://docs.pytest.org/en/latest/explanation/pythonpath.html).

That's why i added the `backend/data_wizard/tests/__init__.py` file.

## Refactor reasoning

We MUST ensure new code changes follows pytest good practices/standards (which mean, not using `unittest`).

Therefore the newly created `NormalizeDfColumnsTest` test class MUST use idiomatic pytest:

- The `NormalizeDfColumnsTest` class SHALL be renamed to a name of the form `Test*` (look for "Prefixed test functions" at [Pytest good integration practices](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html)).
- This class SHALL NOT inherit from `unittest.TestCase`.
- Consequently the tests within class SHALL NOT make usage of `self.assertEqual`.

Keeping unittest is only acceptable for "legacy test file" (meaning old files written before pytest was added to our codebase).

Why [unittest-pytest-guide] doesn't work as a label ?

((Source of the following claims: [unittest-pytest-guide]))

[unittest-pytest-guide]: https://docs.pytest.org/en/stable/how-to/unittest.html

- The pytest doc says unittest-based tests should be "incrementally adapted" to be replaced by pure pytest tests:

> It’s meant for leveraging existing unittest-based test suites to use pytest as a test runner and also allow to incrementally adapt the test suite to take full advantage of pytest’s features.

- Using unittest leaves out a lot of pytest gained benefits.

Search for (CTRL+F): "Benefits out of the box" in [unittest-pytest-guide] to see the list of benefits.

# Note for future test updates

In the future adding extra tests to a non-idiomatic test file should be seen as an opportunity to refactor it to idiomatic pytest.

In most cases, such work can be fully performed with AI as:

- It doesn't introduce any security risk
- If AI is ever "wrong" the test will very likely fail
- Test code depedencies are usually really small so there's no missing context
- This type of refactor is purely deterministic and flat (so there's no bad impact on code quality).

I will do this whole refactor in a separate PR once this one is merged (as i don't want to pollute this PR code diff).

monsieurswag:
Some tests started to fail right after the merge branch 'main' commit (commit hash = `26abbd4964a23035529eeae01788f78d6dd7a7d0`).

If you ever see some test fail without extra commit after the commit mentioned herebefore, just restart the tests until they're green.

These tests are either failing because of `docker pull` commands are being rate-limited (likely throttled).
Or either because of github actions instability.

nas-tabchiche (approved):
LGTM. Can be merged once CI passes

PR #4162, opened 2026-05-19
fix(lib): Add missing EU's CER Directive Excel file

tarkadia:
## Add missing Excel file
The Excel file for the ["EU CER Directive"](https://eur-lex.europa.eu/eli/dir/2022/2557/oj) Framework has been added because it was missin in PR #4097.

## Small Fix
 The useless `dependencies` field has been removed from the YAML, resulting in the creation of a new version of the framework. The content remains unchanged.


eric-intuitem (approved):
LGTM

PR #4190, opened 2026-05-22
fix(lib): add TRUE II n° 901/SGDSN/ANSSI Framework

tarkadia:
## Fix

Add of the REAL ["II n° 901/SGDSN/ANSSI"](https://cyber.gouv.fr/reglementation/cybersecurite-systemes-dinformation/protection-information-sensible-diffusion-restreinte/instruction-interministerielle-n901/) Framework.

The II 901 Framework we had was actually the ["ANSSI - Recommandations pour les architectures des SI sensibles ou DR"](https://messervices.cyber.gouv.fr/guides/recommandations-pour-les-architectures-des-si-sensibles-ou-dr) Framework, a guide that ensures PARTIAL compliance with II 901. This framework has been renamed to `anssi-rec-architectures-systeme-information-sensibles-ou-diffusion-restreinte`, with the right Framework description.

> [!IMPORTANT]
> For `Annexe 2`, it is only usefull for the `Article 14`. I put it in Article 14's annotation at first, but the problem is that it was making the requirement very big. I then decided to put it at the very bottom of the Framework as in the original document, but it's not very practical to scroll all the way down. I think it'd be nice to probably have a system making us able to create a link to jump to another requirement so we don't have to scroll.

## Notes
* The REAL II 901 Framework uses `ii-901_integral` as its base URN, since the ANSSI guide already uses `ii-901`.
* As `Annexe 1` is very long, numbering has been added to the subsections of Objectives to make the Framework easier to read.
* URNs of the wrong II 901 framework (ANSSI guide) remain unchanged.


eric-intuitem (approved):
LGTM

PR #4227, opened 2026-05-29
fix: normalise mixed-scale aggregation in tree and radar displays

nas-tabchiche:
Two display paths missed the normalisation that `_compute_score_for_field`
already does:

- `annotate_tree_with_aggregated_scores` (tree donuts) collapsed missing
  scores via `score or 0` and produced a negative ratio on offset scales
  (e.g. 1..4: `(0 - 1) / 3 = -0.33`). A leaf with `is_scored=True` and
  `score=None` is now excluded from aggregation, same as `is_scored=False`.
  Same fix for `documentation_score`.
- `compare`'s radar maturity-per-top-level was a raw weighted average over
  whatever scales the leaves carried, so a binary 0..1 leaf and a 0..5 leaf
  were averaged on the same axis. Now reuses `_compute_score_for_field` so
  mixed-scale subtrees normalise to ratios and denormalise on the CA scale,
  matching the global score.

Uniform-scale audits are unaffected (normalize + denormalize is identity).
Mixed-scale audits will see different radar values — by design.

## Summary by CodeRabbit

* **Bug Fixes**
  * Leaves marked scored with no numeric score are treated as unscored; offset/alternative scales use the assessment-configured minimum
  * Unset documentation score is treated as explicit zero in documentation rollups and ratios
  * Radar maturity respects mixed-scale normalization, NA anchoring, and maps internal sentinel negatives to zero

* **Tests**
  * Added regression and endpoint tests covering mixed-scale normalization, offset-scale edge cases, unset-documentation semantics, NA anchoring, and authenticated endpoint flows


backend/core/helpers.py
```diff
@@ -442,9 +442,14 @@ def walk(node: dict) -> None:
...
+            # `is_scored` with `score is None` is a data inconsistency; treat
+            # it as unscored to avoid producing a negative ratio on offset
+            # scales (where (0 - min) / range < 0).
+            if is_assessed and score_val is None:
+                is_assessed = False
             weight = node.get("weight") or 1
             if is_assessed:
-                score_val = node.get("score") or 0
                 ra_min = (
                     node.get("min_score") if node.get("min_score") is not None else 0
```
Mohamed-Hacene: we should fallback `ra_min` to` ca_min` instead of 0

Mohamed-Hacene (dismissed):
0k

Mohamed-Hacene (approved):
0k

PR #4257, opened 2026-06-03
fix: risk-assessment scenarios filter current level and residual level

tchoumi313:
### FIX SUP-1354

Fix the EBIOS risk-assessment scenarios filter  current level and residual level: the dropdown was polluted with levels from every visible matrix and a +1 off-by-one made selecting "Medium" return "High" scenarios, with "Low" (value 0) silently unselectable.

Now the dropdown is scoped to the assessment's own matrix, values are the matrix index (aligned with what's stored), 0 is treated as a real selection, and a numeric ?current_level=0 URL survives reload — by forwarding the query string in the SvelteKit proxy, scoping the backend risk action, and normalizing falsy/string-vs-number value comparisons in AutocompleteSelect.

Closes #

## Checklist

- [x] Library changes built via the `tools/` script

- [x] Tests added or updated and passing locally — backend `poetry run pytest`, frontend `pnpm run test` / `./tests/e2e-tests.sh` *(if relevant)*
- [x] Regression test added for bug fixes *(if relevant)*
- [x] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added *(if relevant)*
- [x] No tests or docs needed for this change — why: Because it' a fix without  needing us to have a new ui nor doc 

## Summary by CodeRabbit

* **New Features**
  * Risk assessment views now return matrices scoped to the user's accessible data and can be filtered by assessment.

* **Bug Fixes**
  * Fixed dropdown selection mismatches by normalizing option IDs and value comparisons.
  * Corrected endpoint query handling so filter requests include URL query parameters.
  * Resolved inconsistent risk option label selection when multiple sources contribute.

* **Improvements**
  * Table filters now update reactively when source or configuration changes; filter config typing made more extensible.


backend/core/views.py
```diff
@@ -2745,9 +2745,15 @@ def risk(self, request):
         viewable_matrices: list[RiskMatrix] = RoleAssignment.get_accessible_object_ids(
             Folder.get_root_folder(), request.user, RiskMatrix
         )[0]
+        matrices = RiskMatrix.objects.filter(id__in=viewable_matrices)
+
+        risk_assessment_id = request.query_params.get("risk_assessment")
+        if risk_assessment_id:
+            matrices = matrices.filter(riskassessment__id=risk_assessment_id).distinct()
```
monsieurswag: The `.dinstinct()` call SHALL be removed as it's useless.
There's only one `RiskMatrix` per `RiskAssessment` as you can see at `backend/core/models.py:5829`:

```py
risk_matrix = models.ForeignKey(
    RiskMatrix,
    on_delete=models.PROTECT,
    help_text=_("WARNING! After choosing it, you will not be able to change it"),
    verbose_name=_("Risk matrix"),
)
```

Because of it it's impossible for `matrices.filter(riskassessment__id=risk_assessment_id)` to return more than one `RiskMatrix` which makes `.dinstinct()` usage useless.

backend/core/views.py
```diff
@@ -2757,8 +2763,8 @@ def risk(self, request):
 
                 # Use the translated name if available, otherwise fall back to the default name
                 name = translated.get("name") or risk.get("name", "")
-                _choices[risk.get("id", i)] = name
-            options = options | _choices
+                _choices[i] = name
+            options = _choices | options
```
monsieurswag: Do the following:

- Delete the `_choices` variable
- Delete `options = _choices | options`
- Replace `_choices[i] = name` by `options[i] = name`.

The reasons are:

- Make the code shorter
- Reduce the number of symbols (variables).
- Make the `options` value changes easier to track.

Btw never use `options = _choices | options` to update a dict.
The idiomatic way to update a dict in python is `options.update(_choices)`.

frontend/src/lib/components/Forms/AutocompleteSelect.svelte
```diff
@@ -149,9 +149,14 @@
 
 	const { value, errors, constraints } = formFieldProxy(form, valuePath);
 
+	type SelectValue = string | number | undefined;
+
 	let selected: typeof options = $state([]);
```
monsieurswag: Replace `typeof options` with `Option[]` as using `typeof {...}` for developers to jump off to `options` to try to guess the what the `typeof options` type truly means.

frontend/src/lib/components/Forms/AutocompleteSelect.svelte
```diff
@@ -149,9 +149,14 @@
...
-	let selectedValues: (string | undefined)[] = $derived(
-		selected.map((item) => item.value || item.label || item)
+	let selectedValues: SelectValue[] = $derived(
+		selected.map((item) => {
+			const v = item?.value;
+			return v !== undefined && v !== null && v !== '' ? v : (item?.label ?? item);
+		})
 	);
```
monsieurswag: The `item` is effectively of type `Option`.

Based on this there are multiple issues.

# First issue

`item?.value` doesn't make sense as it assume `value` may not be a property of `item` while the `Option` type states that the `value` property is NEVER `undefined`.
(As you can see line 14: `value: string | number;`)

So you MUST either:
- Change `value: string | number;` to `value?: string | number;` in `interface Option`.
- Or change `item?.value` to `item.value` (and remove `v !== undefined`).

# Second issue

The `v !== null` comparison doesn't make sense as the `interface Option` contains `value: string | number;` which indicates `value` can't be `null`.

So you MUST either:
- Change `value: string | number;` to `value?: string | number;` in the `interface Option`.
- Or remove `v !== null`.

# Third issue

The `item?.label` code also doesn't make sense as the `interface Option` contains `label: string;` which can't be undefined.

So you MUST either:
- Change `item?.label` to `item.bale`
- Or change `label: string;` to `label?: string;` in `interface Option`

# Fourth issue

As said before `item` is of type `Option`.
If `selectedValues` is of type `SelectValue[]` THEN the return value of the function passed to `selected.map` MUST be of type `SelectValue`.

If both `v` and `item?.label` are evaluted to an empty string `""` the returned value will be `item`:

`return v !== undefined && v !== null && v !== '' ? v : (item?.label ?? item)`

But `item` is of type `Option`, the `Option` type isn't compatible with the `SelectValue`, so you MUST never return `item` this function.

frontend/src/lib/components/Forms/AutocompleteSelect.svelte
```diff
@@ -269,13 +274,9 @@
...
 				optionsLoaded = true;
 			}
-			// After options are loaded, set initial selection using stored initial value
-			if (initialValue) {
-				selected = options.filter((item) =>
-					Array.isArray(initialValue)
-						? initialValue.includes(item.value)
-						: item.value === initialValue
-				);
+			if (initialValue !== undefined && initialValue !== null && initialValue !== '') {
```
monsieurswag, line 277: Your condition `if (initialValue !== undefined && initialValue !== null && initialValue !== '')` is inconsistent with `if (initialValue != null && options.length > 0)` at line 171.

Replace the line 171 if, with the one you wrote at line 277.
(as yours is clearly better, using a loose comparison like `!= null` is really bad).

frontend/src/routes/(app)/(internal)/risk-assessments/[id=uuid]/+page.svelte
```diff
@@ -37,6 +39,20 @@
 	const useBubbles = data.useBubbles;
 	const risk_assessment = $derived(data.risk_assessment);
 
+	const scenarioTableFilters = $derived.by(() => {
+		const base = listViewFields['risk-scenarios'].filters;
+		const scope: [string, string][] = [['risk_assessment', risk_assessment.id]];
+		const withScope = (f: ListViewFilterConfig): ListViewFilterConfig => ({
+			...f,
+			props: { ...f.props, optionsDetailedUrlParameters: scope }
+		});
```
Comment (monsieurswag) [inline frontend/src/routes/(app)/(internal)/risk-assessments/[id=uuid]/+page.svelte]: This part is good, but please replace `f` by `filter`.
We should really avoid using one-letter variable as much as we can in the codebase, it just make things harder to read.
It kind of make sense when you have 4 times the same variable on the same line, but most of the time this is a bad practice.

monsieurswag (changes requested):
The PR works as expected, there are just some typing stuff to correct.

Adding `[key: string]: unknown;` to `ListViewFilterConfig` is smart, i should have done it when i created this interface.

# Important:

There's another bug, when selecting multiple values for the `current_level` or `residual_level` only the last one is used for filtering.
This bug isn't caused by your PR, it originates from the `main` branch.

You can either fix this bug within this PR (if it doesn't add to much extra complexity to the PR diff).
OR otherwise just create (and assign yourself) a ticket for this bug on Jira and fix this bug in a second PR once this one is merged.

monsieurswag (approved):
LGTM

PR #4268, opened 2026-06-05
fix: increase autocomplete's lazymode cap to 20

### LINKED ISSUE #4159: Limited list in search results
**Describe the bug**
It appears that the list of search results for assets (from vulnerabilities!?) is limited. Although there are more matching assets, which can positively be looked up individually, they are not shown when only typing a partial string.

**To Reproduce**
1. Go to Vulnerabilities
2. Click on Assets
3. Start typing an asset name that matches multiple assets and has more than 9 potential results.
4. See that only the first 9 are displayed, not the rest.

**Expected behavior**
I expect all matches to be displayed.

**Screenshots**

Image

Image

 - Version 3.16.3

**Additional context**

### LINKED ISSUE #4178: Inconsistent search results drop-down
**Describe the bug**
In the Assets section, when adding multiple dependencies by ways of a filtered search, the list of search results is reset back to unfiltered after adding. Consequently, I have to type the (same) search string repeatedly, when adding multiple dependencies.

**To Reproduce**
1. Go to Assets, open any asset, go to Dependencies.
2. Type a string to search for matching assets.
3. Click on any item from the filtered results list
4. See that the filter disappears and the list of results is reset back to unfiltered.

**Expected behavior**
For example, when adding assets to a vulnerability and typing a string to search for assets, a list of matching results is displayed. Now, when I click on one match, this one item is added but the other result items remain in the list (it is only getting one item shorter). This is fine, and I would expect this to be the standard mode of operations. 

**Screenshots**
Step 1 - search:

Image

Step 2 - click to add:

Image

Step 3 - list is reset:

Image

Step 4 - need to type again:

Image

 - Version 3.16.3

**Additional context**

ab-smith:
## Summary by CodeRabbit

* **Improvements**
  * Autocomplete search now displays an increased number of results when lazy loading, enabling users to see more available options during their search interactions.

### DIFF
--- a/frontend/src/lib/components/Forms/AutocompleteSelect.svelte
+++ b/frontend/src/lib/components/Forms/AutocompleteSelect.svelte
@@ -124,7 +124,7 @@
 		optionSnippet = undefined,
 		placeholder = '',
 		lazy = false,
-		lazyLimit = 10,
+		lazyLimit = 20,
 		lazyThreshold = 50,
 		maxVisibleChips: _maxVisibleChips = 3
 	}: Props = $props();


Axxiar:
The PR works fine, but the UI is misleading now. 

We end up with imbricated scrolls, meaning that when we are at the bottom of the inner scroll, we are not at the bottom of the 20 elements list and we have to scroll the outer one to reach the end 

image

Axxiar (approved):
LGTM

PR #4270, opened 2026-06-05
feat:  add a feature for an admin to reset the mfa of another user

tchoumi313:
## What & why

Enable admin to reset the MFA of their users without having to open a ticket 

Closes #

## Test plan

image

image

## Checklist

- [ ] CLA accepted (first contribution only)

- [X] Svelte 5 syntax; `pnpm run lint` and `pnpm run format` run

- [X] Tests added or updated and passing locally — backend `poetry run pytest`, frontend `pnpm run test` / `./tests/e2e-tests.sh` *(if relevant)*
- [ ] Regression test added for bug fixes *(if relevant)*
- [x] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added *(if relevant)*

## Release Notes

* **New Features**
  * Administrators can now disable another user’s MFA from the user management interface via a dedicated page with a confirmation step (type-to-confirm).
  * MFA status can now be surfaced in the user edit experience to drive when the disable option is shown.

* **Documentation**
  * Added admin recovery guidance for disabling another user’s MFA, including an audit notes warning and admin/self-disable limitations.
  * Expanded MFA local authentication documentation (TOTP, WebAuthn/passkeys, recovery codes).

* **Tests**
  * Added API tests covering success, permission checks, and error cases.

* **Localization**
  * Added MFA disable UI strings across many languages.


backend/iam/views.py
```diff
@@ -460,6 +463,52 @@ def post(self, request, *args, **kwargs):
...
+        target_user = serializer.validated_data.get("user")
+
+        # Block self-reset: admins must use the normal MFA disable flow
+        # (which requires their TOTP code) so a stolen admin session
+        # cannot silently drop the second factor.
+        if target_user.pk == request.user.pk:
+            return Response(
+                {"error": "cannotResetOwnMFA"},
+                status=status.HTTP_403_FORBIDDEN,
+            )
```
monsieurswag: # TO FIX

An admin SHALL have the permission to reset its own MFA.

At `frontend/src/routes/(app)/(third-party)/my-profile/settings/+page.server.ts:136` the sveltekit action `deactivateTOTP` shows the `${ALLAUTH_API_URL}/account/authenticators/totp` endpoint give the ability to anyone (including an admin) to reset itw own MFA.
Based on this preventing an admin from resetting its own MFA is bad.

Hiding the link to the `/users/{self_user.id}/edit/reset-mfa` frontend page if you consider this page exist for account recovery from lost TOTP secret (which wouldn't make sense for someone already authenticated apart from adding a bit more risk of misuse).
So you can let the `showResetMFA` as is.

We can't let one API endpoint allow an admin to reset its own MFA, while the other says it's forbidden.

Fix: just let the admin reset its MFA via the `/users/{self_user.id}/edit/reset-mfa` API endpoint.
Remove the `cannotResetOwnMFA` translation as it will no longer be used.
Remove the `test_admin_cannot_reset_own_mfa` test.
Remove the `ResetMFAView:482-489` code.
Remove the `product-docs/configuration/mfa.md:93` line.

Note:

An "hijacked admin session" was already enough to remove MFA before (you can do it from `/my-profile/settings`).
And even an admin couldn't remove his own MFA with his account, he could juste create another admin user and use it to remove his own MFA.

product-docs/configuration/organization/iam-model.md
```diff
@@ -40,7 +40,9 @@ Multi-Factor Authentication (MFA) is critical for access protection. CISO Assist
 #### Local Authentication <a href="#ember934" id="ember934"></a>
 
 * for local accounts, CISO Assistant includes native MFA
-* based on **TOTP** (e.g., Google/Microsoft Authenticators)
+* based on **TOTP** (e.g., Google/Microsoft Authenticators) and **WebAuthn / passkeys** (hardware security keys, platform authenticators)
+* recovery codes are issued at enrolment so the user can self-recover from a lost factor
+* if a user loses every factor and runs out of recovery codes, an **administrator can reset their MFA** from the user's edit page — see [Setting up MFA → Admin recovery](../mfa.md#admin-recovery-resetting-another-users-mfa). Self-reset is blocked: admins must use the normal MFA settings page (which requires their current second factor) to remove their own MFA.
```
monsieurswag: # TO FIX

Remove `(which requires their current second factor)` from `product-docs/configuration/organization/iam-model.md` as this is false.

Using the `/my-profile/settings` page i can do `"Disable MFA"` without entering my TOTP code.

frontend/messages/en.json
```diff
@@ -4985,5 +4985,15 @@
 	"scoreRange": "Score range",
 	"scoreScale": "Score scale",
 	"downloadExcelTemplate": "Download Excel template",
-	"frameworkActionsHelp": "The Excel template is meant for third parties or external entities to fill offline, then re-import the results through the {dataWizard}."
+	"frameworkActionsHelp": "The Excel template is meant for third parties or external entities to fill offline, then re-import the results through the {dataWizard}.",
+	"resetMFA": "Reset MFA",
+	"resetMFA1": "If this user has lost access to their second factor, you can",
+	"resetMFALinkText": "reset their MFA",
+	"resetMFA2": "All their MFA authenticators will be removed and they will be prompted to re-enroll on their next login",
+	"resetMFAExplanation": "This will remove all MFA authenticators (TOTP, WebAuthn, recovery codes) for {email}. They will be prompted to re-enroll MFA on their next login.",
```
monsieurswag: # TO FIX

When i log in after my MFA has been removed, there's no "prompt" which appears.
Remove the "prompt" mentions throughout this PR's diff.

frontend/messages/en.json
```diff
@@ -4985,5 +4985,15 @@
...
+	"resetMFAWarning": "This action cannot be undone. The user will lose access to their second factor until they re-enroll.",
```
monsieurswag: # TO FIX

I think we should change the `"resetMFAWarning"` translation.
It says `The user will lose access to their second factor until they re-enroll.`.

The `"lose access to {...}"` if a poor choice of word, they won't "lose access" to anything.
This kind of phrasing can scare off admins.

This should be replaced by something emphasizing on the security risk associated with the action.
Like `The MFA security will be disabled for this user until he re-enables it again.`.

frontend/messages/en.json
```diff
@@ -4985,5 +4985,15 @@
...
+	"mfaResetSuccessfully": "MFA has been reset for this user",
+	"errorResettingMFA": "Could not reset MFA for this user",
+	"cannotResetOwnMFA": "Admins cannot reset their own MFA from this endpoint — use the normal MFA settings page",
+	"userHasNoMFAEnabled": "This user does not have MFA enabled"
```
monsieurswag: # TO FIX

We MUST avoid inconsistent terminology, using multiple words for the same concept is the best way to confuse users.
The current terminology is inconsistent.

The original terminology was "enable MFA" and "disable MFA" (as you can see in `"enableTOTP"` and `"disableTOTP"` translations).

Your translations include 2 new terms "re-enroll" (which means enabling again).
And "reset" which means disable.

Fix:

Replace any "re-enroll" term by "enable"
Replace any "reset" term by "disable".

THIS INCLUDES EVERYTHING (Variable names, class names, function names, file names etc...), examples:

- The `reset-mfa` endpoint MUST be renamed to `disable-mfa`.
- The `mfa-reset-confirmation.png` file MUST be renamed to `mfa-disable-confirmation.png`

Etc...

You can just use full text search on your PR's diff to rename everything.

frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/disable-mfa/+page.server.ts
```diff
@@ -0,0 +1,38 @@
...
+			body: JSON.stringify({ user: event.params.id })
+		});
+
+		if (!res.ok) {
+			let payload: Record<string, unknown> = {};
+			try {
+				payload = await res.json();
+			} catch (err) {
+				console.warn('reset-mfa: backend returned a non-JSON error body', err);
+			}
```
Comment (monsieurswag) [inline frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/disable-mfa/+page.server.ts:29]: # NIT

You SHALL NOT wrap the `.json()` call into a try/catch block.

Imagine if we that each time we parse backend response to content to JSON, it would make the code annoying/harder to read.

Run `grep -rnE 'await.*res.*json'` with your cwd at `frontend/src`, you will see we never do this in the codebase.

Fix: remove the try/catch block.

backend/iam/views.py
```diff
@@ -460,6 +463,52 @@ def post(self, request, *args, **kwargs):
...
+
+        # Delete first and branch on the deleted count to avoid an
+        # exists-then-delete race where a concurrent request could empty the
+        # authenticators between the check and the delete.
+        deleted_count, _ = Authenticator.objects.filter(user=target_user).delete()
+        if not deleted_count:
```
monsieurswag: # NIT

You SHALL NOT use `not deleted_count` to check if an `int` isn't `0`.

Fix: use `if deleted_count > 0` instead as it's way easier to comprehend.

Note that `not deleted_count` would be `true` if `deleted_count` was inferior to `0`.

frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/+page.svelte
```diff
@@ -11,6 +11,10 @@
 	}
 
 	let { data }: Props = $props();
+
+	const isAdmin = $derived(Boolean(page.data.user?.is_admin));
+	const isSelf = $derived(page.data.user?.id === data.object.id);
+	const showResetMFA = $derived(isAdmin && !isSelf && data.object.has_mfa_enabled === true);
```
Comment (monsieurswag) [inline frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/+page.svelte]: # NIT

Replace `data.object.has_mfa_enabled === true`, never do `=== true` or `=== false` comparisons for strictly boolean types.

Fix: Just write `const showResetMFA = $derived(isAdmin && !isSelf && data.object.has_mfa_enabled);` instead.

I guess you want to ensure logical boolean variables remain of the `Boolean` type.
Which is why you call `Boolean` in the `isAdmin` definition, to ensure `undefined` becomes `false` is `page.data.user` is `null`/`undefined`.

You shouldn't assume `has_mfa_enabled` can be undefined, at `backend/core/serializers.py:1796` you can see `UserReadSerializer.has_mfa_enabled` is set to `serializers.BooleanField(read_only=True)`.
Which mean the `"has_mfa_enabled"` property is always defined (and will therefore never be read as `undefined` by the frontend).

frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/reset-mfa/+page.svelte
```diff
@@ -0,0 +1,58 @@
...
+
+	interface Props {
+		data: PageData;
+	}
+
+	let { data }: Props = $props();
+	let userInput = $state('');
+
+	const yes = $derived(m.yes().toLowerCase());
+	const canConfirm = $derived(!!userInput && userInput.trim().toLowerCase() === yes);
```
Comment (monsieurswag) [inline frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/reset-mfa/+page.svelte]: # NIT

The `!!userInput &&` condition is useless as it's only `true` if `userInput` is an empty string and `userInput.trim().toLowerCase() === yes` would be `false` with an empty string anyway.

Note: NEVER use `!!someVariable`, `Boolean(someVariable)` will always be more readable.
And NEVER use `!!someString` to check if a string is empty, use `someString === ""` instead.

Fix: remove `!!userInput &&`.

backend/iam/urls.py
```diff
@@ -27,6 +28,7 @@
         name="password-reset-confirm",
     ),
     path("set-password/", SetPasswordView.as_view(), name="set-password"),
+    path("reset-mfa/", ResetMFAView.as_view(), name="reset-mfa"),
```
monsieurswag: # FIW

From a pure REST API standpoint the `/api/iam/reset-mfa/` endpoint choice is bad, it should be `/api/users/{user.id}/reset-mfa/`.

But let's keep it this way for now as endpoints with similar pattern already exist (e.g. `/api/iam/change-password/`) and we don't want the routing to be inconsistent.

frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/disable-mfa/+page.svelte
```diff
@@ -0,0 +1,58 @@
...
+	import { m } from '$paraglide/messages';
+
+	interface Props {
+		data: PageData;
+	}
+
+	let { data }: Props = $props();
+	let userInput = $state('');
+
+	const yes = $derived(m.yes().toLowerCase());
```
Comment (monsieurswag) [inline frontend/src/routes/(app)/(internal)/users/[id=uuid]/edit/disable-mfa/+page.svelte:13]: # FIW

Using `$derived(m.yes().toLowerCase());` instead of `"yes"` is bad for 2 reasons.

## It's bad for accessibility.

What if the user can't/don't know how to type some korean/chinese character.
Sure he could switch his language, but there's no way the average user would understand that.

## It's confusing as hell in some languages

e.g. The chinese translation for it is `"confirmYes": "在下方输入'yes'确认：",` which indicates the string which must be typed is the english version of `yes` even though the chinese one must be typed (`"是"`).

I mean there's no benefit to localizing `"yes"` anyway, this is some kind of passcode more than a real semantic "yes".

Using "yes" instead is perfect:

- The ASCII characters composing "yes" are accessible by anyone.
- 99% of the non-english speaker world's population know the meaning of "yes" anyway.
- The remaining 1% will interpret it as some kind of "passcode" to perform some action.

We shouldn't fix this problem in this PR though as it occurs at multiple place in the codebase.

monsieurswag (dismissed):

eric-intuitem (approved):
LGTM now

PR #4284, opened 2026-06-07
feat: add link icon next to task name in control plan when link defined

eric-intuitem:
## What & why

This facilitates the work of teams that use external tools for managing the tasks, and use CA only for control.

## Test plan

Link is visible when non-empty, and clicking on it opens the corresponding page.
image

## Summary by CodeRabbit

* **New Features**
  * Tasks with external links in the review interface now show a visible external-link icon for quick identification.
  * The icon appears in both compact and standard views and opens links in a new tab for easy access to associated resources.


ab-smith:
could be misleading because the text is clickable to send to the actual task, at least a different icon

ab-smith (approved):

PR #4289, opened 2026-06-08
fix: audit fields visibility for third parties

Mohamed-Hacene:
What & why
Third-party (TPRM) respondents saw fields meant to be hidden from them (score, status, extended_result…). Two root causes:

TPRM-entity audits never populated field_visibility — EntityAssessmentWriteSerializer.create left it {}, and the frontend treats an empty map as "show everything." (The other creation paths already populated it.)
Respondent detection only matched the AUDITEE role — get_auditee_filtered_folder_ids() excluded third-party, so on the folder-based endpoints they fell through to viewer_role = "auditor" and weren't row-scoped.
Fix
Always emit a complete field_visibility: populate it at creation for TPRM audits, and default it on read in ComplianceAssessmentReadSerializer (repairs existing {} audits, no migration).
Treat third-party as a respondent everywhere: collapse the duplicate get_auditee_/get_respondent_filtered_folder_ids helpers into one covering both auditee + third-party. Every call site already uses it, so viewer_role, row-scoping, and write-guards all get fixed with no per-site changes.
Internal users unaffected; auditors unchanged. Files: core/utils.py, core/serializers.py, tprm/serializers.py (views.py untouched).

Test plan
Third-party respondent: hidden fields gone and rows scoped to assignments (list, table-mode, RA detail), incl. a previously-empty-field_visibility audit.
Auditor / reader: full view unchanged.
New TPRM audit: field_visibility populated at creation.
Backend test_field_visibility.py + test_api_requirement_assessments.py pass.


eric-intuitem:
We'll need a more comprehensive refactoring to base access on permissions instead of roles. We will add a new permission that the respondent is lacking: view_audit_full. With that new permission the decisions will be straightforward and the model much simpler to understand.

eric-intuitem (approved):
LGTM now

PR #4291, opened 2026-06-08
fix: new limit in lib import for ref_id

melinoix:
## What & why

SQLite does not enforce `VARCHAR(max_length)` at the DB level, so `ref_id` values
longer than 100 chars can be stored silently. PostgreSQL *does* enforce it, causing
cryptic failures during SQLite → PG migrations.

This adds an explicit length check in save, for all string fields.

## Test plan

- Attempted to load a library YAML containing a `ref_id` of 101+ chars → load is
  rejected with: `'ref_id' value is too long (N chars) — maximum allowed is 100.`
- Loaded a valid library (all `ref_id` ≤ 100 chars) → loads without error.
- Ran `poetry run pytest` — no regressions.

## Checklist

### Backend

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted validation behavior for risk matrix imports to streamline processing
  * Enhanced model save operations with improved field-level validation while refining unique constraint checking during persistence


melinoix:
Need a review specifically from @eric-intuitem , concerning a problem we found in deployment/migration

eric-intuitem (dismissed):
The limit is fixed in models.py. We need to augment it, and to make sure the ORM implements it everywhere, even on sqlite. 
Currently there are holes when loading libraries.

eric-intuitem (approved):
LGTM now

PR #4302, opened 2026-06-09
feat: add the assignee null option in the evidences filters list

martinzerty:
## What & why

This PR adds the ability to filter the evidences' ower and not assigned (with the "--" option) and refactors the null option in the frontend.

For the frontend, instead of manually adding the "--" option in the tables.ts, we pass `nullable:true` in the AutocompleteSelect component which adds the "--" option.

## Test plan

* Create multiple evidences, some with assignee, some without
* Check the assignee' filter options in the evidences list
* Filter the evidences with no assignee "--"

image

image

## Checklist

- [x] Tests added or updated and passing locally — backend `poetry run pytest`, frontend `pnpm run test` / `./tests/e2e-tests.sh` *(if relevant)*

## Summary by CodeRabbit

- **New Features**
  - Added a posture assessment status filter.
  - Improved multi-select dropdown filtering so “unset” (`--`) is controlled via double-dash behavior instead of hard-coded options.

- **Bug Fixes**
  - Standardized handling of unset/null, empty, and single/multi selections across filters, including evidence owner filtering and nullable choice filtering.
  - Updated scenario filtering to use a dedicated nullable status filter set.
  - Evidence `created_at`/`updated_at` range filtering is now provided via shared timestamp handling.

- **Tests**
  - Added coverage for nullable filter behavior (form-level normalization and database-backed queryset filtering).


backend/core/views.py
```diff
@@ -11054,28 +11054,51 @@ def inspect_requirement(self, request, pk):
         )
 
 
+class EvidenceFilterSet(GenericFilterSet):
+    status = NullableChoiceFilter(choices=Evidence.Status.choices)
```
monsieurswag: # TO FIX

The `status = NullableChoiceFilter(choices=Evidence.Status.choices)` filter seems useless as the `Evidence.status` field (defined at `backend/core/models.py:4504`) is non-nullable (`default=status.DRAFT`).

What was you reason to do that ?
If this is an error, remove this line from the code.

backend/core/views.py
```diff
@@ -11054,28 +11054,51 @@ def inspect_requirement(self, request, pk):
...
+        real_actors = Actor.objects.filter(
+            id__in=[v for v in owner_values if v != "--"]
+        )
+        if has_null and real_actors:
+            return queryset.filter(
+                Q(owner__isnull=True) | Q(owner__in=real_actors)
+            ).distinct()
+        elif has_null:
+            return queryset.filter(owner__isnull=True)
+        return queryset.filter(owner__in=real_actors).distinct()
```
monsieurswag: # TO FIX

Overriding `filter_queryset` is usually bad, it should only be used at a last resort.

You MUST implement a reusable filter for this, as we surely will use "nullable" M2M field filters for other objects too after doing it for `Evidence.owner`.

An example of how to implement a reusable filter like this can be found here: https://github.com/intuitem/ciso-assistant-community/commit/8327e8f7843ab8abd609028ba9c1d50ee5991819

frontend/src/lib/components/Forms/AutocompleteSelect.svelte
```diff
@@ -144,6 +144,15 @@
 		});
 	}
 
+	function prependNullOption(opts) {
+		if (!nullable || opts.some((o) => o.value === '--')) return opts;
+		return [{ label: '--', value: '--', translatedLabel: '--' }, ...opts];
+	}
```
monsieurswag: # NIT

Using an early return for an otherwise single-liend function doesn't make it more readable.

And positive conditions are usually a lot easier to understand than negative ones.
In this case your condition is a sort of negative version of the condition required for the last line be executed.

Also typing is missing, you should always add types when you can.

Writting the function like this would be better:

```ts
function prependNullOption(options: Option[]): Option[] {
  const hasNullOption = options.some((option) => option.value === '--');

  if (nullable && !hasNullOption) {
    const nullOption = { label: '--', value: '--', translatedLabel: '--' };
    options = [nullOption, ...options];
  }

  return options;
}
```

frontend/src/lib/components/Forms/AutocompleteSelect.svelte
```diff
@@ -238,7 +247,7 @@
 							// Small dataset with complete response — use eager mode
 							effectiveLazy = false;
 							if (returnedCount > 0) {
-								options = processOptions(items);
+								options = prependNullOption(processOptions(items));
```
monsieurswag: # NIT

Having `prependNullOption(processOptions(items))` is bad as "prepending null options" is a part of "processing options".
The null option prepending should either be "seemless" or directly integrated inside `processOptions`.

We may not even need a `prependNullOption` funtion.

Maybe the following one can work (i didn't test it), you can try it:

```ts
const NULL_OPTION: Option = { label: '--', value: '--', translatedLabel: '--' };

$effect(() => {
  if (nullable) {
    const isNullOptionMissing = options.every((option) => option.value !== '--');

    if (isNullOptionMissing) {
      options = [NULL_OPTION, ...options];
    }
  }
});
```

frontend/src/lib/utils/table.ts
```diff
@@ -76,10 +76,9 @@ const CONTENT_TYPE_OPTIONS = [
 	{ label: 'ENCLAVE', value: 'EN' }
 ];
 
-const YES_NO_UNSET_OPTIONS = [
+const YES_NO_CAPITAL_OPTIONS = [
```
monsieurswag: # NIT

Rename `YES_NO_CAPITAL_OPTIONS` to `YES_NO_OPTIONS`, as "capital" just make the variable name more confusing, and the case of the `label` strings in the variable are irrevelants.

This style of naming would only make sense if we had 2 different version like (`YES_NO_OPTIONS` and `YES_NO_UPPERCASE_OPTIONS`) or (`YES_NO_LOWERCASE_OPTIONS` and `YES_NO_UPPERCASE_OPTIONS`)

monsieurswag (dismissed):
The reviewed code SHALL be fixed.
And both the new tests SHALL be removed.

Tests are inefficients if they don't target abstractions.
Testing for specific fields in specific objects is inneficient.
The `backend/core/tests/test_evidence_filter.py` test is inefficient as it only test 2 fields of a single object.

Creating a test for filters is smart though, testing for a function/class reused throughout the filters could be usefull.

---

The `frontend/src/lib/utils/nullable-filter.test.ts` test however really is useless.

A dedicated 47 line test file is overkill if it's just for testing a non security sensitive, local(non reused/exported), low regression risk, low-volatility(not likely to be changed), highly simple, 2-line function.

backend/core/views.py
martinzerty: my mistake

eric-intuitem:
Is this PR still needed following #4316?

martinzerty:
> Is this PR still needed following #4316?

Yes but has to be adapted to the PR #4316

tchoumi313 (dismissed):
LGTM

tchoumi313 (dismissed):
LGTM

tchoumi313 (dismissed):
LGTM

frontend/src/lib/utils/table.ts
```diff
@@ -445,8 +439,9 @@ export const RISK_TOLERANCE_FILTER: ListViewFilterConfig = {
 	component: AutocompleteSelect,
 	props: {
 		label: 'withinTolerance',
-		options: YES_NO_UNSET_OPTIONS,
-		multiple: false
+		options: YES_NO_OPTIONS,
```
Axxiar: `RISK_TOLERANCE_FILTER` is now broken, it sends wrong values to the backend (true/false instead of previous yes/no)

backend/core/views.py
```diff
@@ -9459,47 +9505,36 @@ def inspect_requirement(self, request, pk):
...
-        "status": ["exact"],
-        "expiry_date": ["exact"],
-        "contracts": ["exact"],
-        "processings": ["exact"],
-        "data_breaches": ["exact"],
-        "created_at": ["gte", "lt"],
-        "updated_at": ["gte", "lt"],
-    }
-
-    def get_queryset(self):
```
Axxiar, line 9624: why removing the `get_queryset` override ? we lose some optimization from `prefetch_related` if doing so.

backend/core/views.py
```diff
@@ -9459,47 +9505,36 @@ def inspect_requirement(self, request, pk):
         )
 
 
+class EvidenceFilterSet(TimestampRangeFilterMixin, GenericFilterSet):
+    owner = NullableModelChoiceFilter(queryset=Actor.objects.all())
+
+    class Meta:
+        model = Evidence
+        fields = [
```
Axxiar, line 9648: missing `security_exceptions`

Axxiar (changes requested):

Axxiar (approved):
LGTM

PR #4305, opened 2026-06-09
fix(docs): correct uv usage in enterprise backend README

Axxiar:
Quick edit to add `./` to the `manage.sh` uv command, otherwise it fails to find the script


ab-smith (approved):

nas-tabchiche (approved):
LGTM

PR #4320, opened 2026-06-10
fix: revoke all knox token except current one in change password view

melinoix:
## What & why

When a user changed their password via `ChangePasswordView`, existing Knox
session tokens were not invalidated. An attacker who had obtained a valid token
(stolen session, leaked token) kept full API access even after the victim reset
their password — the password change had zero effect on active sessions.

**Root cause:** `ChangePasswordView.post()` called `user.set_password()` +
`user.save()` (lines 425–426) with no follow-up token cleanup. Knox tokens are
independent DB rows — changing the password hash does not touch them.

**Fix — `iam/views.py` lines 427–433:**

```python
AuthToken.objects.filter(user=user).exclude(
            Exists(PersonalAccessToken.objects.filter(auth_token=OuterRef("pk")))
        ).delete()
```

- `exclude(digest=current_digest)` — préserve la session courante du user pour qu'il reste connecté après avoir changé son mdp, sans avoir à se reconnecter
- `exclude(Exists(PersonalAccessToken...))` — préserve les Personal Access Tokens (tokens d'API long-lived créés intentionnellement pour CI/CD, intégrations) — les supprimer serait un effet de bord non désiré. C'est le même pattern que `RevokeOtherSessionsView` (ligne 487)
- Tous les autres tokens (sessions browser ouvertes ailleurs, token volé) sont supprimés

Closes #

## Test plan

**Vulnérabilité confirmée avant fix :**

1. Token actif confirmé — `GET /api/iam/current-user/` avec le token `82eb2e78...` → **200** avec les données du user `admin@tests.com`
2. Changement de mot de passe — `POST /api/iam/change-password/` avec `old_password: "poipoipoi"` et `new_password: "NouveauMotDePasse123!"` → **200 OK**
3. Même token retesté après le changement — `GET /api/iam/current-user/` → **200** ← token toujours valide malgré le changement de mot de passe, vulnérabilité confirmée

**Après fix (redémarrage serveur requis) :**

1. Répéter les étapes 1–2 → le token utilisé pour le changement reste valide (session courante préservée)
2. Tester un second token du même user → **401** — révoqué

## Checklist

- [x] Migrations generated with `makemigrations` and committed — none needed
- [x] New dependencies checked — none added

- [ ] Tests added or updated and passing locally — `uv run pytest`
- [ ] No docs update needed — internal security fix, no API contract change

## Summary by CodeRabbit

* **New Features**
  * Changing your password now logs out all other active sessions across devices and browsers while keeping your current session active.
  * Personal API access tokens remain functional after password changes, preserving integrations and third‑party access.
  * Password changes and session revocation are applied atomically so both succeed or neither do.
  * Session revocation runs even when the current session token is not included in the request.


melinoix:
@coderabbitai review

monsieurswag:
Should we invalidate token even if `old_password == new_password` ?
If the password remain the same it seems kind of weird to invalidate the sessions (tokens).

Did you think about it ?

monsieurswag:
Btw where the `current_token = request.META.get("HTTP_AUTHORIZATION", "").split(" ")[-1]` code you mentionned in your PR description comes from ?

I did full text search with this string in our codebase for both the `main` branch and the `revoke-all-knox-token-except-current-changepasswordview` branch (this PR's branch) and i can't find it anywhere.

It's a bit important because using the `[-1]` index is bad as it would mean `Authorization: token {token} GARBAGE` would pick `GARBAGE` as a token instead of `{token}`.
It seems like a really weird API behavior.

monsieurswag:
And shouldn't we also delete the current token used by the user ?
Most security sensitive apps i used in my life also invalid the current token during a password change.

It kind of make sense when you think about it, i know it's a bit worst for UX but for this kind of functionality we should obviously prioritize security over UX.

melinoix:
> Should we invalidate token even if `old_password == new_password` ? If the password remain the same it seems kind of weird to invalidate the sessions (tokens).
> 
> Did you think about it ?

Yes we should also invalidate just to clean for security

melinoix:
> Btw where the `current_token = request.META.get("HTTP_AUTHORIZATION", "").split(" ")[-1]` code you mentionned in your PR description comes from ?
> 
> I did full text search with this string in our codebase for both the `main` branch and the `revoke-all-knox-token-except-current-changepasswordview` branch (this PR's branch) and i can't find it anywhere.
> 
> It's a bit important because using the `[-1]` index is bad as it would mean `Authorization: token {token} GARBAGE` would pick `GARBAGE` as a token instead of `{token}`. It seems like a really weird API behavior.

Already [1] i just did not update the description

melinoix:
> And shouldn't we also delete the current token used by the user ? Most security sensitive apps i used in my life also invalid the current token during a password change.
> 
> It kind of make sense when you think about it, i know it's a bit worst for UX but for this kind of functionality we should obviously prioritize security over UX.

Yes lets also delete the current

nas-tabchiche (changes requested):
Needs tests

melinoix:
Good PR but need a little relaunch about today iam

PR #3492, opened 2026-02-18
fix(lib): Convert all v1 Excel to v2 Format

tarkadia:
One-time conversion to ease reuse/derivation of Excel libraries

## Summary by CodeRabbit

* **New Features**
  * Added bulk-conversion mode for processing multiple Excel files with directory traversal and customizable output options
  * New validation tool to verify converted Excel files against reference libraries and generate compatibility reports
  * Directory traversal utility for recursive Excel file conversion with detailed logging and error tracking
  * File-renaming utility with dry-run mode for removing "_NEW" suffixes from Excel files

PR #4327, opened 2026-06-11
fix(lib): English version of ENS Framework

tarkadia:
## Framework

Add English version of the ["ROYAL DECREE 311/2022, OF 3 MAY, REGULATING THE NATIONAL SECURITY"](https://ens.ccn.cni.es/es/docman/documentos-publicos/39-boe-a-2022-7191-national-security-framework-ens/file) Framework.

## Fix

The original Excel file containing the full content of the framework was lost because of PR #3492. The right version has been restored.

## Info

There was a misdesign in the framework, as some Ref. IDs contained text that could be translated in English. After discussion, we decided that keeping them in Spanish was the best approach. 

- [x] Library changes built via the `tools/` script


eric-intuitem (approved):
LGTM now

PR #4329, opened 2026-06-11
fix: clamp audit score progress values

Mohamed-Hacene:
Some audits created before the score initialization feature can contain requirement scores set to null or 0. With recent score calculation changes, these legacy values can produce negative donut progress values on scales whose minimum is greater than 0.

Clamp the rendered progress against the audit score range before converting it to a donut percentage so legacy data no longer breaks the audit page.


nas-tabchiche (approved):
LGTM

martinzerty (commented):
Fixes the bug LGTM

PR #4362, opened 2026-06-18
feat(integrations): support asset sync + cache ServiceNow schema with startup warming and manual refresh

nas-tabchiche:
## What & why

The ServiceNow integration settings page fetched the table/column/choice schema live from ServiceNow on every load. Those calls are slow (the table list returns thousands of rows; columns and choices recursively walk the parent-table inheritance chain), so the page was sluggish each time it opened.

This caches the schema in a DB-backed table so the page reads it locally:

- New `IntegrationSchemaCache` model (one-to-one with `IntegrationConfiguration`). DB-backed so it is shared across web workers and the Huey consumer and survives restarts. No Redis dependency.
- The `get_tables` / `get_columns` / `get_choices` RPCs read through the cache and fall back to a live fetch-and-store on a miss, so it self-heals.
- Startup warming: a Huey `on_startup` hook enqueues a populate-if-empty warm of each active ServiceNow config (tables + configured table columns + mapped choice fields). A restart with an already-populated cache costs no live calls.
- Manual "Refresh schema" button in the field mapper force re-fetches and overwrites the cache for when the schema changes ServiceNow-side. Generic and harmless for other providers (base no-op).

## Test plan

- `uv run pytest integrations/itsm/servicenow/test_schema_cache.py` → 10 tests passing (cache miss populates and stores, cache hit skips the client, columns/choices cached, force refresh overwrites, populate-if-empty skips when warm and fills when cold, choice warming driven by field_map).
- Migration applied on both SQLite and Postgres.
- Clicked through the ServiceNow settings page: schema loads from cache, and the Refresh schema button re-pulls and updates the table/column/choice dropdowns.

## Checklist

- [x] Svelte 5 syntax; `pnpm run format` run (`pnpm run lint` not run yet)
- [x] New/changed UI strings added to `frontend/messages/en.json` (`refreshSchema`, `schemaRefreshed`; fr also added)

## Summary by CodeRabbit

* **New Features**
  * Added model-aware integration mappings and sync targeting (separate asset vs control behaviors).
  * Introduced a cached schema refresh flow for ServiceNow (tables/columns/choices) with a UI action and background cache warming.
  * Added `model_key` support for remote object listing and refreshed mapping suggestions per model.
* **Bug Fixes**
  * Fixed remote field/value conversions to use the selected model’s mapping rules and prevent legacy defaults from leaking.
  * Improved conflict handling and relinking behavior; ambiguous remote matches are now safely skipped.
  * ServiceNow record labels are more consistent across different record types.
* **Tests**
  * Added/expanded schema-cache and model-aware mapping/sync coverage.


ab-smith:
bot review:
```
- #1 N+1 on the Asset list — fix before merge. Not a blocker in the "it's broken" sense, but it's a real performance regression on a hot, high-cardinality endpoint that ships silently and is annoying to walk back later. Cheap to fix (gate to_representation to non-list actions, or prefetch). I'd hold for this.
- #2 Jira "Refresh schema" always errors — fix before merge or hide the button. Every Jira user who clicks the new button gets an error toast. It contradicts the PR's own stated intent ("harmless for other providers"). Small fix. Should go in.
- #3 duplicate field keys / #4 wasted SELECT — not blockers. Ship-and-follow-up is fine. #3 is a11y/constraint-indicator polish (binding is correct); #4 mirrors pre-existing AppliedControl behavior.
```

nas-tabchiche:
> - N+1 on the Asset list — fix before merge. Not a blocker in the "it's broken" sense, but it's a real performance regression on a hot, high-cardinality endpoint that ships silently and is annoying to walk back later. Cheap to fix (gate to_representation to non-list actions, or prefetch). I'd hold for this.

This one is a false positive. `AssetListSerializer` does not include the linking mixin, therefore the per-row sync never runs on the list endpoint.

> - Jira "Refresh schema" always errors — fix before merge or hide the button. Every Jira user who clicks the new button gets an error toast. It contradicts the PR's own stated intent ("harmless for other providers"). Small fix. Should go in.

Addressed, `execute_action` now routes `refresh_schema` to the base noop before building a client. Added to `get_interactive_actions()`

> - duplicate field keys / wasted SELECT — not blockers. Ship-and-follow-up is fine. 3 is a11y/constraint-indicator polish (binding is correct); 4 mirrors pre-existing AppliedControl behavior.

Both addressed. `FieldMapper.svelte` now namespaces the `field` attribute with a prefix.
`IntegrationLinkSerializerMixin.to_representation` now gates the `SyncMapping` query on retrieve.

Mohamed-Hacene (approved):
0k

PR #4381, opened 2026-06-22
Fix Python 3 syntax error in convert_library_v2.py exception handlers

smakarim:
## Fix Python 3 syntax error in `convert_library_v2.py`

The library converter (`backend/scripts/convert_library_v2.py`, also exposed via the `tools/convert_library_v2.py` symlink) uses the **Python 2** exception syntax `except TypeError, ValueError:` in four places. Under Python 3 this is a hard `SyntaxError`, so the module fails to even import:

```
$ python3 backend/scripts/convert_library_v2.py library.xlsx
  File "backend/scripts/convert_library_v2.py", line 859
    except TypeError, ValueError:
           ^^^^^^^^^^^^^^^^^^^^^
SyntaxError: multiple exception types must be parenthesized
```

Because it is a parse-time error, the converter cannot run at all once execution reaches code that exercises these handlers — answer scoring (`add_score`), framework `weight` / integer fields, and metric `default_target`. In practice this breaks conversion for any library that uses implementation groups or scoring.

### Fix
Wrap the exception tuples so they are valid Python 3:

```python
except (TypeError, ValueError):
```

Four occurrences (lines 859, 1138, 1153, 1294). No behavior change otherwise.

### Verification
- `python3 -c "import ast; ast.parse(open('backend/scripts/convert_library_v2.py').read())"` now parses cleanly (previously raised `SyntaxError`).
- Ran the converter end to end on a library that uses implementation groups; it completes and writes valid YAML where it previously aborted at import.

## Summary by CodeRabbit

* **Chores**
  * Updated data conversion script syntax for improved code consistency and maintainability.

### DIFF
--- a/backend/scripts/convert_library_v2.py
+++ b/backend/scripts/convert_library_v2.py
@@ -856,7 +856,7 @@ def _handle_framework(obj, library, object_blocks, prefix_to_urn, compat_mode, v
                             try:
                                 score_to_add = int(val)
                                 choices[i]["add_score"] = score_to_add
-                            except TypeError, ValueError:
+                            except (TypeError, ValueError):
                                 raise ValueError(
                                     f"(answers_definition) Invalid add_score value '{val}' "
                                     f"for answer ID '{answer_id}', choice #{i + 1}. Must be an integer"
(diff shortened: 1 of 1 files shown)


ab-smith:
Hello,
This a false positive since the syntax is now the default format of python 3.14 with latest ruff formatting.

For the record: PEP 758 (Python 3.14) allows except A, B: without parentheses (as long as no as clause is used), and Ruff v0.15.0+ enforces this concise style when target-version = py314.

Regards

PR #4382, opened 2026-06-22
Fix Python 3 syntax errors: parenthesize multi-type except clauses (backend does not import on Py3)

smakarim:
### Problem
The backend currently **does not import on Python 3**. A prior formatting pass (#4335, "align to ruff v0.15.x and code quality recovery") removed the parentheses from every multi-exception `except` clause, leaving Python-2 syntax:

```python
except TypeError, ValueError:
```

On Python 3 this is a hard `SyntaxError`:

```
  File "backend/core/views.py", line 368
    except TypeError, ValueError:
           ^^^^^^^^^^^^^^^^^^^^^
SyntaxError: multiple exception types must be parenthesized
```

This affects **87 clauses across 37 files**, including core modules that are imported on every request — `core/views.py`, `core/serializers.py`, `core/models.py`, `core/focus_middleware.py` — so the application cannot start. (`backend/scripts/convert_library_v2.py` is affected too, which is what surfaced this for me while contributing framework libraries.)

### Fix
Wrap each multi-type `except` in a tuple:

```python
except (TypeError, ValueError):
```

This was applied mechanically and only to multi-type clauses (single-type `except X:` and `except X as e:` are untouched). No behavioral change.

### Verification
- `ast.parse` now succeeds for every `backend/**/*.py` (0 files failing; previously 37 failed).
- `python manage.py check` → **System check identified no issues (0 silenced)**.
- `manage.py migrate` + dev server boot succeed and the REST API serves requests (verified on a fresh SQLite instance).

### Note
Supersedes #4381, which fixed only `convert_library_v2.py`. This PR resolves the same class of issue across the whole backend. Happy to close #4381 once this is reviewed.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal code syntax for Python 3 compatibility across multiple backend modules. No user-facing changes or functional updates.

### DIFF
--- a/backend/chat/management/commands/prune_index.py
+++ b/backend/chat/management/commands/prune_index.py
@@ -76,7 +76,7 @@ class Command(BaseCommand):
             try:
                 app_label, model_name = model_path.split(".")
                 model_class = apps.get_model(app_label, model_name)
-            except LookupError, ValueError:
+            except (LookupError, ValueError):
                 continue
             object_type = _normalize_model_name(model_name)
             try:
--- a/backend/chat/memory.py
+++ b/backend/chat/memory.py
@@ -197,7 +197,7 @@ def inject_tool_replays(
             args = obs.get("args", {})
             try:
                 args_str = json.dumps(args, ensure_ascii=False, default=str)
-            except TypeError, ValueError:
+            except (TypeError, ValueError):
                 args_str = str(args)
             result_text = truncate_to_tokens(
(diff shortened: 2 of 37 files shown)


ab-smith:
Hello,
This a false positive since the syntax is now the default format of python 3.14 with latest ruff formatting.

For the record: PEP 758 (Python 3.14) allows except A, B: without parentheses (as long as no as clause is used), and Ruff v0.15.0+ enforces this concise style when target-version = py314.

Regards

PR #4427, opened 2026-06-30
fix: normalize urls to avoid comparison issue on default ports

### LINKED ISSUE #4422: Helm chart: backend service port 80 breaks frontend SSR auth (handleFetch token never attached → empty tables)
## Summary

With the Helm chart's **default `backend.service.port: 80`**, the SvelteKit frontend never attaches the auth token to **server-side (SSR) backend calls**, so every `+page.server.ts` `load()` gets `401` and all data tables render empty — even though the DB is populated, the user is authenticated, and RBAC is correct.

## Root cause

`frontend/src/hooks.server.ts` `handleFetch` only attaches the token when:

```js
if (request.url.startsWith(BASE_API_URL)) { ... request.headers.append('Authorization', `Token ${token}`) }
```

The chart builds `PUBLIC_BACKEND_API_URL` (= `BASE_API_URL`) from the helper:

```
http://<fullname>-backend:<backend.service.port>/api
```

With the chart default `backend.service.port: 80`, this is `http://<fullname>-backend:80/api`. But `fetch()` / `new Request()` **normalises the default HTTP port away**, so the actual `request.url` has no `:80`, and the `startsWith` check fails:

```js
new Request('http://host:80/api/folders/').url
// => 'http://host/api/folders/'        (":80" stripped)
'http://host/api/folders/'.startsWith('http://host:80/api')
// => false   → token never appended → backend returns 401
```

So with port 80, **no SSR backend call is ever authenticated**.

## Impact

Any Helm deployment using the default backend service port (80) behind a reverse proxy / node adapter: login works, the app shell loads, but every data list/table is empty (each page `load()` 401s). Very confusing because the DB and permissions are fine.

## Workarounds / fixes

- **Operator workaround:** set `backend.service.port` to a non-default port (e.g. `8080`). The port is then preserved in `request.url`, `startsWith` matches, and the token is attached. (`targetPort` stays the named `http` container port, so the container is unaffected.)
- **Proper fix (either):**
  1. `backendApiUrl` helper should omit the default port (`80`), or
  2. `handleFetch` should compare **normalised origins/paths** instead of a raw `startsWith` on a URL string that includes a default port.

## Environment

- Chart `ciso-assistant` 0.11.0, app `v3.18.3`.
- Frontend served via the node adapter behind an Ingress (`/` → frontend, `/api` → backend).

ab-smith:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backend URL handling by normalizing the configured API address and removing trailing slashes when possible.
  * Added a safe fallback so the app continues using the provided API URL if parsing fails.
  * Helps reduce connection issues caused by inconsistent URL formatting.

### DIFF
--- a/frontend/src/lib/utils/constants.ts
+++ b/frontend/src/lib/utils/constants.ts
@@ -1,11 +1,18 @@
 import { env } from '$env/dynamic/public';
 import { m } from '$paraglide/messages';
 
-export const BASE_API_URL = `${
-	Object.hasOwn(env, 'PUBLIC_BACKEND_API_URL')
-		? env.PUBLIC_BACKEND_API_URL
-		: 'http://localhost:8000/api'
-}`;
+const rawBackendApiUrl = Object.hasOwn(env, 'PUBLIC_BACKEND_API_URL')
+	? env.PUBLIC_BACKEND_API_URL
+	: 'http://localhost:8000/api';
+
+// Strip default ports so this matches request.url in handleFetch's startsWith check (issue #4422).
+export const BASE_API_URL = (() => {
+	try {
+		return new URL(rawBackendApiUrl).href.replace(/\/$/, '');
+	} catch {
+		return rawBackendApiUrl;
+	}
+})();
 
 export const DEFAULT_LANGUAGE = `${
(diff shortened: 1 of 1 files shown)


Mohamed-Hacene (approved):
0k

PR #4429, opened 2026-06-30
fix: align content-type for two cases outside the hook

### LINKED ISSUE #4423: Password reset request omits Content-Type: application/json → backend 415, no email sent
## Summary

The "forgot password" flow never sends an email. The frontend's server action POSTs to `/api/iam/password-reset/` **without a `Content-Type` header**, so the DRF backend rejects it with **415 Unsupported Media Type** before the email logic runs. The reset-confirm action has the same issue.

## Details

`frontend/src/routes/(authentication)/password-reset/+page.server.ts`:

```js
const endpoint = `${BASE_API_URL}/iam/password-reset/`;
const requestInitOptions = { method: 'POST', body: JSON.stringify(form.data) };
const res = await event.fetch(endpoint, requestInitOptions);
```

There's no `headers` object, so `event.fetch` defaults the body's `Content-Type` to `text/plain;charset=UTF-8`. The backend (DRF, JSON parser only) responds:

```
415  {"detail": "Unsupported media type \"text/plain;charset=UTF-8\" in request."}
```

Same pattern in `password-reset/confirm/+page.server.ts` (POST to `/api/iam/password-reset/confirm/`), so the confirm step would 415 too.

Confirmed: POSTing the same payload with `Content-Type: application/json` returns `202` and the backend logs `Email sent successfully` — so the backend and SMTP are fine; only the missing header is the problem.

## Fix

Add the JSON content type to both reset actions:

```js
const requestInitOptions = {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify(form.data)
};
```

(Still present on `main` as of 2026-06.)

## Environment

- app `v3.18.3`, chart `ciso-assistant` 0.11.0, node-adapter frontend behind an Ingress.

ab-smith:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved password reset and password reset confirmation requests by sending them with the correct JSON content type.
  * This helps ensure the forms are processed reliably by the server.

### DIFF
--- a/frontend/src/routes/(authentication)/password-reset/+page.server.ts
+++ b/frontend/src/routes/(authentication)/password-reset/+page.server.ts
@@ -57,6 +57,7 @@ export const actions: Actions = {
 
 		const requestInitOptions: RequestInit = {
 			method: 'POST',
+			headers: { 'Content-Type': 'application/json' },
 			body: JSON.stringify(form.data)
 		};
 
--- a/frontend/src/routes/(authentication)/password-reset/confirm/+page.server.ts
+++ b/frontend/src/routes/(authentication)/password-reset/confirm/+page.server.ts
@@ -26,6 +26,7 @@ export const actions: Actions = {
 		form.data.uidb64 = event.url.searchParams.get('uidb64');
 		const requestInitOptions: RequestInit = {
 			method: 'POST',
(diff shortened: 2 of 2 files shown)


Mohamed-Hacene (approved):
0k

PR #4434, opened 2026-06-30
feat: multiple import improvements

ab-smith:
- support upgrading composite objects (eg, assessments like audit, risk analysis)
- require domain and make perimeter optional for consistency 
- Switch to domain picker

## Summary by CodeRabbit

* **New Features**
  * The data import wizard can now reconcile imported records into a user-selected existing “target” assessment (via a new target selector) for supported composite assessment types, instead of always creating a new parent container.
  * The import conflict strategy applies to the target’s child items during reconciliation.
* **Bug Fixes**
  * Improved fallback behavior when perimeter is omitted and tightened matching rules to avoid duplicate containers/records.
  * Invalid or inaccessible target selections now fail safely without altering existing containers.
* **Documentation**
  * Updated the conflict/reconciliation guidance with new English and French text for selecting “update existing” vs “create new.”


tchoumi313 (commented):
The “Update existing assessment” selector loads ~1000 items in a native `<select>`, making it hard to use. It should be replaced with a searchable select ( similar to the model selector or ideally AutocompleteSelect) showing the full path.
The domain selector has the same issue, so both could be handled together in a dedicated PR.

tchoumi313 (dismissed):
LGTM

enterprise/frontend/src/routes/(app)/(internal)/extra/data-wizard/+page.server.ts
```diff
@@ -48,6 +49,7 @@ export const actions: Actions = {
 					'X-Perimeter-Id': perimeter,
 					'X-Framework-Id': framework,
 					'X-Matrix-Id': matrix,
+					'X-Target-Id': target,
```
tchoumi313: Optional scope selectors omit their hidden input when empty, so formData.get() returns null. Coerce to '', sending null would serialize to the string "null" and the backend's `or None` guard would treat it as a real (invalid) id. 


```suggestion
					'X-Folder-Id': folder ?? '',
					'X-Perimeter-Id': perimeter ?? '',
					'X-Framework-Id': framework ?? '',
					'X-Matrix-Id': matrix ?? '',
					'X-Target-Id': target ?? '',
```

tchoumi313 (approved):
LGTM can be merge when green

PR #3844, opened 2026-04-03
build: move from poetry to uv

Axxiar:
- [x] venv caching
- [x] update doc
  - [x] product doc
  - [x] READMEs
  - [X] notification-system.md
  - [X] rhel checklist.md
  - [x] offline-deployment.md
- [x] update CE backend
- [x] update dockerfiles
- [x] update docker compose files
- [x] update tests
- [x] helm charts
- [x] Jinja2 templates for make_config.py

## Summary by CodeRabbit

* **Chores**
  * Replaced Poetry with uv across the project; migrated packaging metadata to PEP 621, standardized Python requirement to >=3.14,<4, and updated dependency/dev-dependency declarations, CI workflows, Docker builds, and startup/service run commands.

* **New Features**
  * Added uv tooling configuration, local package source mapping, and automated build metadata generation.

* **Documentation**
  * Updated READMEs and docs to reflect uv/.venv usage and revised command examples.

PR #4442, opened 2026-06-30
fix: Harmonization & Improvement of Docker Scripts & Files

tarkadia:
# Why?

This PR harmonizes the Docker Compose workflows used to initialize and build the Community and Enterprise editions. It fixes several inconsistencies between shell scripts, PowerShell scripts, Jinja2 templates, and the reference Compose files, making local startup more consistent, more reliable, and closer to the expected behavior for each edition.

# Fixes & Improvements

## > fix: Compose Build version of Enterprise Edition

The scripts used to create the `docker-compose-build` version at the root of the project simply could not generate the Enterprise version of CISO Assistant, because they could not take input parameters.
For the Enterprise version, according to the documentation in `enterprise/README.md`, the following command should be used if we want to quickly test the Enterprise version:
```sh
./docker-compose-build.sh -f enterprise/docker-compose-build.yml
```
But this command was building the Community version instead.

**Affected scripts:**
* `docker-compose-build.sh` 
* `docker-compose-build.ps1` 

## > fix: Make `docker-compose-build.sh` `createsuperuser` phase interactive again

Regression introduced by PR #3844. A `-T` had been added, which meant that the console could not ask the user for the e-mail and password.

**Affected scripts:**
* `docker-compose-build.sh` 

## > fix & enhancement: More precise Docker context for Enterprise Edition

This change was made to avoid using the global `.dockerignore` file, which does not contain some specific folders or files to ignore in `/frontend` and `/backend`.

The `.dockerignore` files located in the `/frontend` and `/backend` directories will now be the only truth source to avoid unexpected bugs during local testing, or copying unwanted files (such as `.env` in `/frontend` and `.venv` in `/backend`). 
I'm suggesting this fix because my backend wouldn't start due to a `.venv` directory in `/backend`, and it took me a while to figure out where the problem was coming from.

A combination of the `context` and `additional_contexts` parameters was used in `enterprise/docker-compose-build.yml`. The `Dockerfile` files for the frontend and backend of the Enterprise edition were modified accordingly, while making sure to keep identical behavior.

_Note: This can also, in some cases, considerably reduce the size of Docker images generated locally._

**Affected files:**
* `enterprise/docker-compose-build.yml` 
* `enterprise/backend/Dockerfile` 
* `enterprise/frontend/Dockerfile` 

## > fix & enhancement: Harmonize Health Check Values with root `docker-compose.yml` for Jinja2 Templates

The Jinja templates had verification times for `healthcheck` that did not match the backend `healthcheck` in the root `docker-compose.yml` file (the healthcheck stopped too early, which made docker compose fail).

**Affected files:**
* `config/templates/docker-compose-postgresql-bunkerweb.yml.j2` 
* `config/templates/docker-compose-postgresql-caddy.yml.j2` 
* `config/templates/docker-compose-postgresql-traefik.yml.j2` 
* `enterprise/config/templates/docker-compose-postgresql-caddy.yml.j2` 
* `enterprise/config/templates/docker-compose-postgresql-traefik.yml.j2` 
* `enterprise/config/templates/docker-compose-sqlite-caddy.yml.j2` 
* `enterprise/config/templates/docker-compose-sqlite-traefik.yml.j2` 

## > enhancement: More PowerShell versions of Docker scripts

There are now PowerShell versions of some Docker scripts.

**Affected scripts:**
* `config/docker-compose.ps1` 
* `enterprise/config/docker-compose.ps1` 

## > enhancement: Harmonize Migration waiting Method

The scripts did not all have the same way of waiting for the backend to be ready before creating the `superuser`. In order to harmonize everything and use only one way of doing it for all scripts, I used the `docker-compose-build` scripts at the root of the project as a reference.
The method is as follows:
> 1. Check if the backend is ready.
> 2. If this is not the case, wait 10s.
> 3. Start again from step 1 (60 times max.)

The maximum wait time is equal to 10 minutes so that the backend can run its migration and so people know there is a problem if it exceeds this limit.

**Affected scripts:**
* `docker-compose-build.sh`  (simple rewrite of the method)
* `docker-compose.sh` 
* `docker-compose.ps1` 
* `config/docker-compose.sh` 
* `enterprise/config/docker-compose.sh` 

## > enhancement: Remove useless `uv` calls

In the PowerShell scripts, specifying `uv run` was not necessary. This is due to the fact that the `Dockerfile` files already add the `.venv` to the path.

**Affected scripts:**
* `docker-compose.ps1` 
* `docker-compose-build.ps1` 

## > enhancement: Enforce global `.dockerignore`

Technically, the root `.dockerignore` is not very useful anymore, because the Enterprise edition no longer uses the project root as its context. This file was left here as a safeguard, in case something else that I did not see uses it.
`**/.venv/` and `**/__pycache__/` were added as elements to exclude.

**Affected file:**
* `.dockerignore` 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docker startup gating by waiting for the backend health check to become ready before creating the superuser.
  * Increased backend container healthcheck timing (timeout, retries, and start period) across Docker Compose templates to better handle slower environments.
* **Chores**
  * Standardized Docker Compose startup/build scripts with selectable compose files and more reliable readiness waiting; failures now show recent backend logs.
  * Expanded Docker ignore rules to cover common virtual environments and cache directories.
* **Documentation**
  * Updated enterprise Docker setup instructions to reference the correct Linux/macOS and Windows startup scripts.


eric-intuitem (approved):
LGTM.

PR #4449, opened 2026-07-01
fix(jira): field mapping broken on self-hosted Jira 9+

Mohamed-Hacene:
## What & why

Configuring the Jira field mapping fails on self-hosted Jira Server/Data Center 9.0+: the user can pick a project, but the column list and the suggested mapping never load. Regression introduced with the 3.17.4 field-mapping feature. Jira Cloud is unaffected.

`JiraClient.get_table_columns()` and `_get_allowed_values_from_createmeta()` call `jira.createmeta()`. Atlassian removed that global endpoint in Jira Server/DC 9.0, so the `jira` client raises `Unsupported JIRA version` before any request, which surfaces to the `get_columns` / `suggest_mapping` RPC actions as an error.

Added `_fetch_field_meta()` which keeps `createmeta()` for Cloud and DC < 9.0 and switches to the per-issue-type endpoints (`project_issue_types` / `project_issue_fields`) on DC ≥ 9.0, normalized to the same field shape. No behaviour change on Cloud or DC < 9.0.

## Test plan

- [ ] On a Jira Data Center 9+/10.x instance: open the Jira integration field mapping, select a project + issue type, confirm the columns load and the suggested field/value mapping populates.
- [ ] On Jira Cloud: same screen, confirm mapping still loads (unchanged path).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Jira field and column retrieval across different Jira Server/Data Center versions.
  * Added more reliable fallback handling when metadata lookups fail, reducing broken table-column generation.
  * Standardized how selectable field values are read, helping ensure choice lists display correctly.


nas-tabchiche (approved):
LGTM. Can be merged once CI passes

PR #4453, opened 2026-07-01
feat: autologin page

### LINKED ISSUE #4287: Add automatic redirection to SSO
**Problem statement**
I have configured my CISO Assistant instance with SSO (OpenID Connect). When arriving on the login page, users have to manually click on the "Login with SSO" button despite this being the only possible way of connecting to the app. This is not a huge deal but it is a small integration hiccup that can be smoothed easily.

**Expected behavior**
Automatically go to SSO when it is configured and no valid session is currently found on the user's browser.

**Additional context**
This is a more and more common way to integrate an app with a SSO service. There is, AFAIK, two main philosophies at play :
* Implicit automation : it is possible for the app to somehow deduct that the SSO login is the only way to log in (eg. Jenkins OIC plugin, Rallly...), so there is no point in showing the login page at all.
* Explicit setup : it is still possible/desirable to be able to log in with another method, but SSO should be the preferred way for most users, so in that case the application can provide a config option (eg `SSO_AUTO_REDIRECT=true`) for the admin to enable the automatic SSO redirection.

In both cases, the admin is responsible to document, when relevant, the necessary steps to avoid login with SSO. For example, Jenkins OIC plugin implements a so-called "escape hatch" where requesting explicitly for the `/login` page (instead of the root URL) will not redirect the user to SSO. Other apps account for a specific URL parameter to avoid SSO redirection, like requesting the page `/login?no-sso`.

Based on the way CISO Assistant handle users (some being purely local users with/without MFA, some being SSO users, some users being subjected to the "Force SSO" parameters and some being excluded from it...), I think the explicit setup would be better suited.

Image

It could also be possible to have this behavior as a side effect of checking the "Force SSO" parameter : arguably, if logging in with SSO is supposed to be forced, then non-SSO logins are probably few/non-existent, so bypassing the login page unless explicitly invoked does still make sense.

eric-intuitem:
## What & why

For users authenticating with SSO, it is unfortunate to have to click on the SSO login button each time they want to access the tool. An URL providing direct access would be more comfortable.

/login?sso URL in complement to /login

## Test plan

Test auto login manually.

## Checklist

- [x] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added _(if relevant)_

## Summary by CodeRabbit

* **New Features**
  * Added support for starting SSO login directly from the login page using a special login URL parameter.
  * The login screen now shows a loading state and SSO message while redirecting to the identity provider.
  * Documented how to use direct SSO links and optional post-login destination parameters.

* **Documentation**
  * Expanded guidance for Force SSO Login, local login exceptions, and the impact on existing passwords.


eric-intuitem:
Claude security review:
>drill down into the current PR for security risk


Ran an agent, used a tool
Ran skill
/security-review
Identified security vulns in PR
I'll wait for the identification sub-task to complete.


Ran 2 commands
I have enough context to complete the analysis while the identification sub-task confirms. Let me wait for it to return.


Used a tool
That monitor approach won't signal correctly. Let me stop it and just await the agent's completion notification.


Used 2 tools
I'll await the identification sub-agent's completion notification before finalizing the report.


Background shells
1 completed, 1 running
Security Review — PR autologin-page (?sso direct SSO login)
Scope reviewed: frontend/src/routes/(authentication)/login/FormCard.svelte (code) and product-docs/configuration/sso/README.md (documentation, excluded).

Result: No HIGH or MEDIUM security vulnerabilities found. No findings met the ≥8 confidence threshold, so none are reported.

Analysis of the newly introduced attack surface
The PR adds a ?sso query param that auto-invokes, on mount, the exact SSO redirect the existing Log in with SSO button already performed (triggerSSO() → redirectToProvider(sp_entity_id, getSSOCallbackURL(callback_url), 'login')). It introduces no new data sink — it only automates a pre-existing, user-initiated code path.

The one user-controlled input in the flow is the next query parameter. It is not a new vector (the unchanged getSSOCallbackURL already consumed it), and it is constrained at multiple layers:

Open redirect — not present. next is passed through getSecureRedirect() (helpers.ts:112), whose regex /^\/\w+/ requires a leading / followed by a word character. This rejects protocol-relative (//evil.com), absolute (https://evil.com), and backslash/@ tricks, and coerces failures to '' → /. Validation is applied both when building the callback URL (FormCard.svelte:57) and again server-side before the post-auth redirect (sso/authenticate/+page.server.ts:43). Result: next can only ever be a same-origin relative path.
CSRF — not present. redirectToProvider posts to /iam/sso/redirect/ with a CSRF token from the cookie (allauth.ts:36), unchanged by this PR; the backend validates it.
Authentication bypass — not present. ?sso only starts the redirect; the full IdP flow (OIDC state/nonce validation, email matching) and any MFA still execute server-side. autoSSO is gated on data.SSOInfo?.is_enabled, and the backend remains the authority.
XSS — not present. No user input is interpolated into markup; the conditional UI renders only static i18n strings. next is used solely in URL construction, never in the DOM.
Per review guidance, client-side gating is not itself a vulnerability (the backend validates), and the open-redirect path is affirmatively mitigated rather than merely theoretical.

Conclusion
The change reuses an existing, validated SSO flow and adds no new injection, redirect, CSRF, or auth-bypass surface. No security issues block this PR.

ab-smith (approved):

PR #4464, opened 2026-07-03
fix: better management of non-ascii character on filename during export

### LINKED ISSUE #4370: PDF Export Fails When Risk Assessment Title Contains Non-ASCII Characters (v3.18.2)
### Environment

* CISO Assistant Community Edition
* Version: v3.18.2
* Build: 5cee5f45f
* Deployment: Docker containers on Synology NAS
* Reverse proxy: Caddy
* Browser: Google Chrome 149
* OS: Windows 11

### Description

Exporting a Risk Assessment to PDF via the web interface returns **HTTP 500 Internal Server Error**.

The backend PDF generation endpoint works correctly and successfully generates the PDF, however the frontend export route fails.

### Steps to Reproduce

1. Create or open a Risk Assessment.
2. Use a title containing Slovak diacritic characters, for example:

   * "Štúdia analýzy rizík"
   * any title containing characters such as Š, Č, Ž, Ť, Ľ, etc.
3. Click **Export PDF** from the Risk Assessment page.

### Actual Result

The browser receives:

```text
HTTP 500 Internal Server Error
```

Frontend log:

```text
[500] GET /risk-assessments/<uuid>/export/pdf

TypeError: Cannot convert argument to a ByteString because the character at index 29 has a value of 352 which is greater than 255.

    at webidl.converters.ByteString
    at Object.record<ByteString, ByteString>
    at webidl.converters.HeadersInit
    at Object.ResponseInit
    at new Response
```

### Expected Result

The PDF should be downloaded successfully regardless of whether the Risk Assessment title contains Unicode characters.

### Additional Findings

The backend endpoint works correctly:

```text
GET /api/risk-assessments/<uuid>/risk_assessment_pdf/
```

The PDF is generated and downloaded successfully when called directly with API authentication.

This suggests that the issue is located in the frontend export route:

```text
/risk-assessments/<uuid>/export/pdf
```

Most likely while constructing the HTTP response headers (probably Content-Disposition) using a filename containing non-Latin-1 characters.

### Workaround

Rename the Risk Assessment title to remove diacritics (ASCII-only filename).

For example:

```text
Štúdia analýzy rizík
```

→

```text
Studia analyzy rizik
```

After removing diacritics, PDF export works correctly.

### Suggested Fix

Use RFC 5987 / UTF-8 compatible filename encoding in the Content-Disposition header instead of assuming a Latin-1 ByteString.

ab-smith:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `Content-Disposition` headers for exports, PDFs, and attachments to handle special characters and non‑ASCII filenames more reliably.
  * Strengthened filename sanitization to avoid malformed download responses and header injection edge cases.
  * Standardized attachment/download header generation across compliance exports, risk assessment PDF downloads, and evidence attachment downloads.


eric-intuitem (approved):
LGTM

PR #4478, opened 2026-07-08
fix: exceptions adjustments

Mohamed-Hacene:
## What & why

- Security exception email templates can now be customized from Settings (they were missing from the customization screen).
- The Approver field on security exceptions is now deprecated and read-only, users should use validation flows instead.

## Test plan

- Checked the four security exception templates show up in Settings > email templates.
- Checked the Approver field is disabled and marked as deprecated on the exception form.

## Summary by CodeRabbit

- **Updates**
  - The Security Exception approver field is now read-only and clearly marked as deprecated, with guidance to use validation flows instead.
  - Status-change notifications are sent to security exception owners.
  - Added notification template support for expiring, expired, assigned, and status-changed security exceptions.
  - Added English and French help text explaining the deprecated approver field.


ab-smith:
new templates on the registry needs the text content as well

Mohamed-Hacene:
> new templates on the registry needs the text content as well

Yes, they're already created. This PR exposes them so they can be modified.

ab-smith (approved):

PR #4493, opened 2026-07-10
fix(pro): edge cases with doc report templates override

ab-smith:
## Summary by CodeRabbit

* **New Features**
  * Added Dutch language support for audit reports, including localized compliance labels and domain terminology.
  * Reports now use the language selected in user preferences when choosing templates and translations.

* **Bug Fixes**
  * Improved language fallback behavior so reports remain usable when a translation or language-specific template is unavailable.
  * English templates and labels are now used automatically as a fallback.


Mohamed-Hacene (approved):
Ok

PR #4494, opened 2026-07-10
fix(backend): handle some removed Requirement Fields and IGs on Framework update

tarkadia:
## Why?

Framework updates previously preserved data that had been removed from newer Library versions:

- Nullable **Requirement Node Fields** omitted from the new version kept their previous values.
- **Implementation Groups (IGs)** removed from the Framework definition remained selected in existing Compliance Assessments.

This update ensures that framework updates correctly remove stale data.

> [!NOTE]
> I recommend looking at the commits rather than the entire diff to make it easier to understand.

## Fixes 

- Explicitly define the nullable **Requirement Node Fields** that must be cleared to `null/None` when omitted from a newer framework version, including:
  - `ref_id`
  - `name`
  - `description`
  - `annotation`
  - `typical_evidence`
  - `visibility_expression`
  - `implementation_groups`
- Compare imported values with existing Requirement Nodes and update only fields whose values actually changed.
- Treat an omitted `implementation_groups_definition` as a framework without IGs.
- Remove stale IGs from existing Compliance Assessment selections while preserving valid selections.

## Tests

Added regression coverage for:

- Clearing Requirement Node Fields omitted from a newer framework version.
- Requirement Nodes reorganization.
- Removing one IG while keeping valid selections.
- Clearing all selected IGs when the framework no longer defines any.

## Summary by CodeRabbit

* **Bug Fixes**
  * Framework updates now reconcile changes to implementation-group definitions and clamp compliance assessment selections to only valid groups.
  * If implementation-group definitions are removed in an updated library, affected assessment selections are cleared.
  * Imported requirement node updates now clear fields omitted by newer payloads while preserving updated ordering.
  * Requirement node bulk updates now apply only the specific changed fields to reduce unnecessary updates.
* **Tests**
  * Added regression coverage for requirement-node “clear-on-omission” behavior and implementation-group clamping on framework updates.


eric-intuitem (approved):
LGTM

PR #4502, opened 2026-07-12
feat: allow changing the domain of an EBIOS-RM study

eric-intuitem:
## What & why

It was not possible to move an EBIOS-RM study from one domain to the another. This PR provides this missing capability.

image

## Test plan

unit test / functional test / manual test

## Checklist

## Summary

- **New Features**
  - You can now move an EBIOS RM study to a different domain, with its workshop objects updated accordingly.
  - Added a folder/domain selector to the study form.
  - Added a bulk “change folder/domain” action for EBIOS RM studies.
  - Updated the study update tool to accept an optional folder change.
- **Documentation**
  - Added localized help text explaining what is moved when changing the domain.
- **Bug Fixes**
  - Improved consistency of domain updates while keeping shared deliverables (e.g., elementary actions and risk assessments) in place.
- **Tests**
  - Added domain-move test coverage, including which objects cascade vs. remain exempt.


backend/ebios_rm/serializers.py
```diff
@@ -40,16 +59,8 @@ def update(self, instance, validated_data):
...
-                    StrategicScenario,
-                    AttackPath,
-                    OperationalScenario,
-                ]
-                for model in child_models:
-                    model.objects.filter(ebios_rm_study=updated_instance).update(
+                for model, study_path in STUDY_FOLDER_CASCADE_MODELS.items():
+                    model.objects.filter(**{study_path: updated_instance}).update(
                         folder=updated_instance.folder
                     )
```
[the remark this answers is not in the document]
eric-intuitem: This is on-purpose.

frontend/messages/cs.json
```diff
@@ -3518,6 +3518,7 @@
 	"campaigns": "Kampaně",
 	"addCampaign": "Zahájit kampaň",
 	"campaignDomainHelpText": "Doména, ke které bude kampaň připojena.",
+	"ebiosRmStudyDomainHelpText": "Změna subjektu přesune studii a všechny objekty jejích workshopů do vybraného subjektu.",
```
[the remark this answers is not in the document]
eric-intuitem: Will fix

ab-smith (dismissed):

backend/ebios_rm/serializers.py
```diff
@@ -20,6 +20,25 @@
 from rest_framework import serializers
 
 
+# Models whose folder mirrors the study's folder, mapped to the queryset path
+# leading back to the study. Folder-scoped ebios_rm models must appear here or
+# in STUDY_FOLDER_EXEMPT_MODELS so they are not left behind on a domain move.
+STUDY_FOLDER_CASCADE_MODELS = {
```
monsieurswag: # NIT:

Set the `STUDY_FOLDER_CASCADE_MODELS` type to `Final[dict[type[models.Model], str]]` for better typing.
Otherwise LSPs will fail to infer the dictionary key type which breaks autocomplete/type hinting.

backend/ebios_rm/tests/test_ebios_rm_study.py
```diff
@@ -43,3 +60,104 @@ def test_create_ebios_rm_study_with_assets(self):
...
+            model
+            for model in apps.get_app_config("ebios_rm").get_models()
+            if issubclass(model, FolderMixin)
+        }
+        unclassified = (
+            folder_scoped_models
+            - set(STUDY_FOLDER_CASCADE_MODELS)
+            - STUDY_FOLDER_EXEMPT_MODELS
+        )
+        assert not unclassified, (
```
monsieurswag, line 82: # NIT:

Replace `not unclassified` by `len(unclassified) == 0` (using `not` is only okay on booleans, otherwise it just makes the code less expressive/analyzable).

backend/ebios_rm/serializers.py
```diff
@@ -20,6 +20,25 @@
...
+    StrategicScenario: "ebios_rm_study",
+    AttackPath: "ebios_rm_study",
+    OperationalScenario: "ebios_rm_study",
+    OperatingMode: "operational_scenario__ebios_rm_study",
+    KillChain: "operating_mode__operational_scenario__ebios_rm_study",
+}
+
+# Folder-scoped models that deliberately do not follow the study: they are
+# shared catalog objects (or the study itself) with a user-managed folder.
+STUDY_FOLDER_EXEMPT_MODELS = {EbiosRMStudy, ElementaryAction}
```
monsieurswag: # NIT:

Write a docstring instead of single-line comments when describing the meaning of global constants.

Docstrings may be uglier (because they are placed underneath the variable definition), but using docstrings make it so the comment will be displayed by the IDE when hovering over the constant (which is helpful for devs).

monsieurswag: # NIT:

The `STUDY_FOLDER_EXEMPT_MODELS` constant should be moved to the `test_ebios_rm_study.py` file as it's a test-specific constant (it's dead code within the context of the codebase).

backend/ebios_rm/tests/test_ebios_rm_study.py
```diff
@@ -43,3 +60,104 @@ def test_create_ebios_rm_study_with_assets(self):
...
+        root = Folder.get_root_folder()
+        domain_a = Folder.objects.create(
+            name="Domain A", parent_folder=root, content_type=Folder.ContentType.DOMAIN
+        )
+        domain_b = Folder.objects.create(
+            name="Domain B", parent_folder=root, content_type=Folder.ContentType.DOMAIN
+        )
+        matrix = RiskMatrix.objects.get(
+            urn="urn:intuitem:risk:matrix:risk-matrix-4x4-ebios-rm"
+        )
```
monsieurswag: # NIT:

Why do we even use the `@pytest.mark.usefixtures("ebios_rm_matrix_fixture")` + the manual Risk `RiskMatrix.objects.get` DB select when we can just pass the `ebios_rm_matrix_fixture` as an argument instead ?
(Like we do at `backend/ebios_rm/tests/fixtures.py:31`)

I get that the newly introduced test just copied this logic from the other tests in the `test_ebios_rm_study.py` file, but it seems to be the wrong way to use fixtures.

monsieurswag (dismissed):
# To fix

If the models in `STUDY_FOLDER_CASCADE_MODELS` MUST always keep the same folder as their ebios rm study we SHALL move the `EbiosRMStudyWriteSerializer.update` folder update code to the `EbiosRMStudy.save` method.

Otherwise we may have corectness issues later as this invariant can be violated by changing the `EbiosRMStudy.folder` field directly (without using the `EbiosRMStudyWriteSerializer`).
Example:

```py
>>> ebios_rm_study = EbiosRMStudy.objects.get(name="Test")
>>> ebios_rm_study.folder
<Folder: Test Domain>
>>> ebios_rm_study.stakeholders.first().folder
<Folder: Test Domain>
>>> ebios_rm_study.folder = Folder.objects.get(name="AAA")
>>> ebios_rm_study.save()
>>> ebios_rm_study.folder
<Folder: AAA>
>>> ebios_rm_study.stakeholders.first().folder
<Folder: Test Domain>
```

# Question

If all the models in `STUDY_FOLDER_CASCADE_MODELS` MUST have their folder being the same as the one for their `EbiosRMStudy` why do they even inherit from the `FolderMixin` ?

Is this futureproofing ? (in case we ever need the folder of one of these models to be different from their ebios rm study folder)

eric-intuitem:
Using FolderMixin for secondary objects makes the IAM algorithm simpler and faster. It's a tradeoff, we still have a few objects not (yet) following this pattern in the code.

backend/ebios_rm/tests/test_ebios_rm_study.py
eric-intuitem: if not my_set is perfectly idiomatic, please read https://peps.python.org/pep-0008/#programming-recommendations

backend/ebios_rm/serializers.py
eric-intuitem: ok

backend/ebios_rm/tests/test_ebios_rm_study.py
eric-intuitem: Fair, now use fixture as an argument.

backend/ebios_rm/serializers.py
eric-intuitem: ok

backend/ebios_rm/serializers.py
eric-intuitem: https://peps.python.org/pep-0224/ has been rejected. So we should not use it.

ab-smith (approved):

PR #4506, opened 2026-07-13
fix(ci/cd): step 3: use mirrors for 3rd party images

ab-smith:
## Summary by CodeRabbit

* **Chores**
  * Updated automated test environments to use mirrored PostgreSQL and MailHog service images.
  * Added configurable container image sources for Python, Node.js, Caddy, and Qdrant.
  * Preserved existing default images while enabling alternative image sources for builds and test runs.


eric-intuitem (approved):
LGTM

PR #4519, opened 2026-07-16
chore: rationalize all markdown fields to fix discrepancies

eric-intuitem:
## What & why

Some fields are not providing markdown while they should. Some are providing markdown at rendering but not at edition.
The root cause is the lack of a central Source of Truth for markdown formatting.

Keep the SoT in the frontend (doing in the backend would be very difficult and overkill). The SoT is cruds.ts.

image

## Test plan

Manual test in Ebios-RM workshop 4.

## Checklist

- [x] No tests or docs needed for this change — why: existing tests + manual tests

## Summary by CodeRabbit

* **New Features**
  * Expanded Markdown-capable fields throughout model forms (e.g., observation, annotation, justification, operating modes description).
  * Generalized Markdown rendering across detail/list views so configured fields (including justification and operating-modes text) display formatted content instead of plain text.
* **Tests**
  * Updated functional tests to use the Markdown editor toggle for operating modes.
  * Updated form test helpers to enable Markdown editing based on field type rather than specific field names.


frontend/src/lib/utils/crud.ts
```diff
@@ -3256,12 +3265,33 @@ const FIELD_COMPONENT_MAP = {
 	}
 };
 
+/**
+ * Fields rendered and edited as Markdown on every model.
+ * Together with each model's optional `markdownFields`, this is the single source
+ * of truth consumed by the detail view (render), the list view (render) and, by
+ * convention, the form editors (MarkdownField / TableMarkdownField).
+ */
```
monsieurswag, line 3289: # NIT:

Either keep the first sentence of the docstring or replace it with something close like this (written by ChatGPT btw):

```ts
/**
 * Default fields rendered and edited as Markdown.
 * Model-specific fields can be added through `markdownFields`.
 */
```

We shouldn't let Claude write docstrings/comments because of how horrible they are.
Claude surely was fine tuned to sound as smart as possible.
That's i guess the reason why it likes to write long comments/docstrings with tons of useless hard-to-understand jargon + useless long list of positive details about the commented code.

Example of problems in this docstring:

What source of truth "truth" is it the "single" of ?
Because if we talk about the truth of "should we render a field of a specific model" with markdown then `FIELD_COMPONENT_MAP` is also a source for this truth.

And if we talk about the field names which are enforced to be interepted by markdown then obviously there will be a single source of truth, how could it not be ?
Who's mad enough to write a second source of truth + write the code to reconciliate both source of truth for such a specific thing ?

This is also wrong to say it's consumed by `MarkdownField` and `TableMarkdownField` (obviously it's NOT why would the markdown renderer care about what field should use the markdown renderer itself).

And no one reading a docstring ever cares about who "consumes" the thing the docstring describes anyway.
Imagine if you had to specify the consumers of every single field/function in a codebase.

You never read a docstring hoping to know which part of the codebase consumes it.
And updating docstrings each time a new part of a codebase consumes it would be a maintenance nightmare obviously.

frontend/src/lib/utils/crud.ts
```diff
@@ -169,6 +169,12 @@ export interface ModelMapEntry {
 	path?: string;
 	endpointUrl?: string;
 	customNameDescription?: boolean;
+	/**
+	 * Fields (in addition to DEFAULT_MARKDOWN_FIELDS) whose content is authored and
+	 * rendered as Markdown for this model. Single source of truth consumed by the
+	 * detail view, the list view, and (by convention) the form editors.
+	 */
+	markdownFields?: string[];
```
monsieurswag, line 185: # NIT:

Replace "authored" by "edited" in this docstring as it's the current term we use in our MarkdownField component (As you can see at `frontend/src/lib/components/Forms/MarkdownField.svelte:23` in the `Props["defaultMode"]` type).

Only keep the first sentence of the docstring as the rest is just claude yapping.

Adding docstrings to "abstraction skeletons" (typescript interface fields, abstract classes (python `abstractmethod`/typescript `abstract`)) with an unclear meaning is really important, we didn't do it enough on the frontend.

Don't forget using markdown for JS docstrings is a good practice (you can wrap variable names by backticks like we tend to do in our python docstrings).

frontend/src/lib/components/DetailView/DetailView.svelte
```diff
@@ -12,7 +12,7 @@
 	import { booleanDisplay } from '$lib/utils/boolean-display';
 	import { ISO_8601_REGEX } from '$lib/utils/constants';
 	import { type ModelMapEntry, type ReverseForeignKeyField } from '$lib/utils/crud';
-	import { getModelInfo } from '$lib/utils/crud.js';
+	import { getModelInfo, getMarkdownFields } from '$lib/utils/crud.js';
```
monsieurswag: # NIT:

We should remove ".js" from the import path.
It's bad for traceability (it makes the import path more complex with no benefit) + it's confusing (devs can assume this is pure JS code).

frontend/src/lib/components/Forms/ModelForm/AccreditationForm.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
 	import AutocompleteSelect from '../AutocompleteSelect.svelte';
 	import TextField from '../TextField.svelte';
-	import TextArea from '../TextArea.svelte';
+	import MarkdownField from '../MarkdownField.svelte';
```
monsieurswag: # FYI:

Mixing 2 different import styles is bad:

```ts
import MarkdownField from '../MarkdownField.svelte';
// Mixed up with (in other files):
import MarkdownField from '$lib/components/Forms/MarkdownField.svelte';
```

It's easier to list the places where a specific module/component has been imported if all import paths are the exact same.
That's why we should always use absolute import paths.

frontend/tests/utils/form-content.ts
```diff
@@ -62,12 +62,10 @@ export class FormContent {
...
-			// Check if this is a markdown field (description, observation, or justification) and handle it
-			if (
-				(key === 'description' || key === 'observation' || key === 'justification') &&
-				field?.type === FormFieldType.TEXT
-			) {
-				// Try to click the markdown edit button if it exists
+			// Markdown fields render preview-first; switch to edit mode before filling.
+			// Detected generically via the edit-toggle button, so there is no per-field
+			// list to keep in sync with the app's markdown fields.
+			if (field?.type === FormFieldType.TEXT) {
```
monsieurswag: # NIT:

Another useless justification comment by claude.
The code itself is great and easy to understand, we don't need any comment.

frontend/src/lib/utils/crud.ts
```diff
@@ -3256,12 +3265,33 @@ const FIELD_COMPONENT_MAP = {
...
+ * Together with each model's optional `markdownFields`, this is the single source
+ * of truth consumed by the detail view (render), the list view (render) and, by
+ * convention, the form editors (MarkdownField / TableMarkdownField).
+ */
+export const DEFAULT_MARKDOWN_FIELDS = [
+	'description',
+	'observation',
+	'annotation',
+	'justification'
+];
```
monsieurswag: # NIT:

We don't need to `export` this constant as it's only meant to be used by `crud.ts`.

monsieurswag (changes requested):
# TO FIX:

The `justification` field isn't rendered as markdown in the `operational-scenarios` detail view (frontend page: `/operational-scenarios/{ID}?activity=one`).

Image

---

The `justification` field isn't rendered as markdown in the `ro-to` detail view (frontend page: `/ro-to/{ID}?activity=two`).

Image

---

The `annotation` field isn't rendered as markdown in the `risk-matrices` detail view (frontend page: `/risk-matrices/{ID}`)

Image

---

The `AttackPath.justification` can be modified in the frontend BUT isn't displayed in the `attack-paths` detail view (frontend page: `/attack-paths/{ID}`).

This clearly seems to be a "bug" (something we forgot to do at least), if you agree this is an error use this PR to fix this issue.

The `ComplianceAssessment.observation` also have the exact same issue.

# FYI:

The `RiskAcceptance.justification` field can't be modified in the frontend (from what i've seen), i wonder if that's normal.

---

The `Project.observation` field can't be modified in the frontend, i wonder if that's normal.

frontend/src/routes/(app)/(internal)/risk-matrices/[id=uuid]/+page.svelte
```diff
@@ -1,10 +1,12 @@
 <script lang="ts">
 	import Anchor from '$lib/components/Anchor/Anchor.svelte';
 	import RiskMatrix from '$lib/components/RiskMatrix/RiskMatrix.svelte';
-	import { URL_MODEL_MAP } from '$lib/utils/crud.js';
+	import MarkdownRenderer from '$lib/components/MarkdownRenderer.svelte';
+	import { URL_MODEL_MAP, getMarkdownFields } from '$lib/utils/crud';
 	const showRisks = true;
 	let { data } = $props();
 	const riskMatrix = data.data;
+	const markdownFieldSet = getMarkdownFields(data.urlModel);
```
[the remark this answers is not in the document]
Comment (monsieurswag) [reply frontend/src/routes/(app)/(internal)/risk-matrices/[id=uuid]/+page.svelte:9]: That's right it's always better to use $derived when transforming a value coming from $props.

monsieurswag (approved):
LGTM

PR #4562, opened 2026-07-23
fix: Restore builtin behavior

eric-intuitem:
## What & why

Builtin objects should not be created/modified/deleted via the API.

## Test plan

Enrich the existing test on groups.

## Checklist

## Summary

* **Bug Fixes**
  * Built-in records are now protected from destructive changes: DELETE requests are denied (HTTP 403).
  * Built-in fields are enforced as read-only; only `is_visible` can be edited for applicable models, while other attributes remain immutable.
  * Built-in user group renames and attempts to change the builtin status are prevented; membership additions still work.
  * Batch mutation flows now consistently respect built-in immutability rules.
  * Note: the built-in “Main” entity remains fully editable.
* **Tests**
  * Added API tests covering built-in immutability across PATCH, batch updates, and deletion outcomes.


monsieurswag (dismissed):
LGTM

ab-smith:
bot feedback:

The blanket 403 on PUT/PATCH breaks two shipped flows. Some builtin objects are meant to be partially editable:

1. Builtin terminologies / classifications — you can't delete them, but you're supposed to be able to hide them. The edit form for builtin rows shows only the is_visible checkbox (TerminologyForm.svelte:42-74). Saving it now returns 403.
2. The "Main" entity — created with builtin=True at startup (startup.py:1809), and the UI explicitly allows editing it (DetailView.svelte:418-426). Renaming it to your org name now returns 403.

Suggested fix: keep the new guard for DELETE only. For updates, enforce immutability in the serializer instead: when the instance is builtin, make all fields read-only except a per-model allowlist (is_visible for terminologies/classifications, the normal editable fields for Entity). The serializer + batch changes in the PR are good as-is.

Also: UserGroupViewSet.destroy (views.py:7715) is now dead code — the permission guard fires before it. It can be removed.

ab-smith (approved):

PR #4582, opened 2026-07-28
feat(data-wizard): add iam group column support for folder creation in data wizard

tchoumi313:
## What & why

Add the column iam_group to the domain import to enables IAM group creation for the folder
 
Closes #

## Checklist

- [ ] CLA accepted (first contribution only)

- [ ] Svelte 5 syntax; `pnpm run lint` and `pnpm run format` run
- [ ] New/changed UI strings added to `frontend/messages/en.json` (keep keys, preserve `{tokens}`)

- [ ] Regression test added for bug fixes _(if relevant)_
- [X] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added _(if relevant)_

## Summary by CodeRabbit

* **New Features**
  * Folders created or updated with IAM group creation enabled now receive default IAM groups and role assignments automatically.
  * Folder imports support additional true/false values, including `x`, `oui`, `vrai`, `non`, and `faux`, plus numeric values; matching is case-insensitive.
  * Blank or unrecognized values leave the existing setting unchanged.
  * Disabling IAM groups removes default groups and role assignments when permitted; updates are rejected when users or identity-provider mappings prevent removal.

* **Documentation**
  * Updated folder import guidance and command-line help for supported values and behavior.


product-docs/configuration/data-import.md
```diff
@@ -823,6 +818,7 @@ Folders (domains) are the top-level organisational units in CISO Assistant. Impo
 * Conflict detection is performed by `name` + parent folder.
 * When `domain` is left blank the folder is attached to the root of the tenant.
 * An error is returned if `domain` matches more than one folder name.
+* `iam_group` only needs a value to trigger the flag; leaving it blank does not disable IAM group creation on an existing folder.
```
monsieurswag: This behavior isn't consistent with the frontend Folder edit(update) form.
When disabling the `"Create IAM groups"` checkbox in the folder edit form (frontend URL: `/folders/{ID}/edit`) the "IAM groups" (the `UserGroup` objects related to this domain) get removed (after submitting the form).

The data-wizard should have the same behavior.

tchoumi313: adjusted

backend/data_wizard/views.py
```diff
@@ -728,7 +732,8 @@ def process_records(self, records: list[dict]) -> Result:
             )
             if serializer.is_valid():
                 try:
-                    serializer.save()
+                    instance = serializer.save()
+                    self.after_create(instance)
```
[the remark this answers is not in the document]
tchoumi313: fix

backend/data_wizard/views.py
```diff
@@ -927,6 +927,10 @@ def prepare_create(
     ) -> tuple[dict, Optional[Error]]:
         pass
 
+    def after_create(self, instance) -> None:
+        """Hook called after a new record is created. No-op by default."""
+        pass
```
monsieurswag, line 932: # TO FIX:

Remove the `RecordConsumer.after_create` and `FolderRecordConsumer.after_create` functions.

You know we must keep `RecordConsumer` as simple as possible otherwise the abstraction will become bloated and become a pain to work with.
So obviously i would like us to avoid adding new "hooks" to it as much as possible.

In this case i don't even see why we need a `after_create` hook.

I commented the `self.after_create(instance)` line at `backend/data_wizard/views.py:1091` and the `create_iam_groups` column of the folder import still works.

To me it seems logic as the the `FolderWriteSerializer` shoud handle the `"create_iam_groups"` "field" by itself.
Observe it by yourself, run the following code in your django shell:

```py
>>> from core.serializers import FolderWriteSerializer as S
>>> a = S(data={"name": "XXXXX", "create_iam_groups": True})
>>> a.is_valid()
True
>>> a.save()
<Folder: XXXXX>
>>> b = S(data={"name": "YYYYY", "create_iam_groups": False})
>>> b.is_valid()
True
>>> b.save()
<Folder: YYYYY>
```

You can see the `XXX` Folder has the "IAM Groups" column set to true, while the `YYY` doesn't.

Image

Btw after commenting the `self.after_create(instance)` call, i ran the tests and everything was green, so removing `after_create` isn't wrong based on your tests.
(Command used to run the tests: `uv run pytest data_wizard/tests/test_consumers.py -vv --reuse-db`)

monsieurswag (changes requested):

PR #4594, opened 2026-07-29
fix!: prune old invalid Implementation Groups consistently when updating a framework

ab-smith:
## Summary by CodeRabbit

* **Bug Fixes**
  * Removed selections that reference implementation groups no longer present after framework updates, keeping compliance assessments, posture assessments, and campaigns aligned with current framework definitions.
  * Added a library-conversion integrity check that detects and fails fast when requirement nodes reference “orphan” implementation groups not declared by the framework.

### DIFF
--- a/backend/core/models.py
+++ b/backend/core/models.py
@@ -983,6 +983,69 @@ class LibraryUpdater:
                 },
             )
 
+    @staticmethod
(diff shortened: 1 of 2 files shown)


eric-intuitem (approved):
LGTM

PR #4603, opened 2026-08-01
chores(ci): fix release versions now that we have multiple releases streams

ab-smith:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved release version detection during automated builds.
  * Tag-triggered builds now use the pushed version tag directly.
  * Non-tag builds consistently derive versions from matching version tags.


eric-intuitem (approved):
LGTM

PR #4630, opened 2026-08-06
fix: allow embedded tables to save filters in the localstorage

martinzerty:
## What & why

Currently the embedded tables can be filtered but the filters are not saved in the LocalStorage. It implies the lose of the filter when going back to the page. For instance, when going on a finding tracking view at `/findings-assessments/uuid`, we have the findings and the evidences tables. These two are embedded tables in the findings-assessments view. However, the filters in those tables are not saved and the page resetting (reloading with no url parameters) removes the filters.

Now, all the embedded tables also save their filter in the localstorage. So, we now have a unic key for the filters in `/findings-assessments/uuid1` and `/findings-assessments/uuid2`.

Closes #

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Filter selections now persist consistently in standalone and embedded tables.
  * Saved filters are scoped to the current page and table, preventing unintended overlap.
  * Reused tables now refresh filters correctly when their context changes.
  * Resetting filters now reliably clears saved filter state across all table types.
  * Filter values and form data now refresh correctly when a table’s context changes.


martinzerty:
@coderabbit full review

frontend/src/lib/components/ModelTable/ModelTable.svelte
```diff
@@ -503,6 +508,27 @@
...
+			}
+			_form.form.update((data) => {
+				for (const field of filteredFields) {
+					data[field] = (fresh[field] ?? []).map((v: any) => v.value);
+				}
+				return data;
+			});
+		}
+		previousFilterStoreKey = key;
+	});
```
nas-tabchiche: Are you sure this re-seeds?
Svelte 5 runs effects in declaration order. You have another effect affecting filters and depending on `page.url` at line 447, that one will run first.
Let's say we have 2 applied controls: A and B. If you go from A's detail view to B's, this happens:
1. The effect at line 447 runs first. It puts A's filters in the new URL (:457-460) and saves A's filters in `localStorage` under B's key (:476-478).
2. Then this new `$effect` runs and calls `seedFilterValues`, which reads the URL and `localStorage`, but both contain A's filters.

nas-tabchiche (changes requested):
See linked comment. Also, `localStorage` growth is unbounded.

PR #4635, opened 2026-08-07
feat(frontend): rework the audit table-mode

Mohamed-Hacene:
## What & why

Refreshes and reworks the compliance audit table-mode.

- The assessment/questionnaire toggle is now an auditor comfort switch, decoupled from third parties: assessment shows everything with questionnaire answers read-only, questionnaire ("focus") shows only the questions and surfaces the computed result/score.
- A collapsible table of contents keeps section context on scroll and its result filter now also filters the list, and the header gains shortcuts (edit audit, action plan, evidences). 
- Editing respects reader-vs-editor permissions and the focus-mode domain filter, the header scores refresh live after a score edit, and the page is lighter than before (no per-requirement superForms, no full reload on every field edit).

## Test plan

- [ ] As an editor, set result / status / score / extended-result inline; header scores update without a page reload.
- [ ] Switched to questionnaire ("focus") mode: only questions are editable and the rest is hidden, computed result/score shown; assessment mode shows the answers read-only.
- [ ] As a reader (no change permission), every control is disabled and create/select is hidden.
- [ ] With focus mode active, evidence/measure selection lists only items from the focused domain.
- [ ] TOC result filter hides non-matching requirements while keeping their section headers.
- [ ] Flat audit (no sections) and a deeply nested one: sticky section bar and TOC navigation land correctly.

## Summary by CodeRabbit

* **New Features**
  * Added a redesigned compliance assessment table with collapsible sections, sticky progress header, filtering, section navigation, and streamlined requirement cards.
  * Added score controls with sliders, progress rings, score descriptions, and empty-state handling.
  * Added a global score endpoint and parent requirement information in assessment responses.
  * Added segmented controls for status and result selection.

* **Improvements**
  * Assessments now respect editing permissions and clearly indicate read-only access.
  * Updated extended-result labeling and localized permission messaging.
  * Removed the former standalone table-of-contents component and utilities.


melinoix (dismissed):
extended_result_options is built from extendedResultColorMap keys (not_set, observation, ...), but the backend's RequirementAssessment.ExtendedResult enum only accepts observation_sensitive_point (not observation) and has no not_set choice — it uses null/blank instead. Selecting "Observation" or clearing a badge sends an invalid value and gets silently rejected (the page just reverts via invalidateAll()). auditee-assessments/[id=uuid]/+page.svelte:65-71 has the correct mapping (observation_sensitive_point + null on clear) — worth reusing that here.


ScoreControl supports a scored/onScoredChange toggle to clear is_scored, but none of the call sites here pass it — is_scored is only ever set to true (line 681/703), never back to false. Once a score slider is touched, there's no way left in table-mode to un-score a requirement, so it stays permanently counted in get_global_score(). The old Checkbox/switch this PR removed handled that.

Image

You can not put back to ' i do not want this to be scored'

Mohamed-Hacene:
> extended_result_options is built from extendedResultColorMap keys (not_set, observation, ...), but the backend's RequirementAssessment.ExtendedResult enum only accepts observation_sensitive_point (not observation) and has no not_set choice — it uses null/blank instead. Selecting "Observation" or clearing a badge sends an invalid value and gets silently rejected (the page just reverts via invalidateAll()). auditee-assessments/[id=uuid]/+page.svelte:65-71 has the correct mapping (observation_sensitive_point + null on clear) — worth reusing that here.
> 
> ScoreControl supports a scored/onScoredChange toggle to clear is_scored, but none of the call sites here pass it — is_scored is only ever set to true (line 681/703), never back to false. Once a score slider is touched, there's no way left in table-mode to un-score a requirement, so it stays permanently counted in get_global_score(). The old Checkbox/switch this PR removed handled that.
> 
> Image
> You can not put back to ' i do not want this to be scored'

I agree with the first point, but not with the second. This is a design choice, since we decided to activate or deactivate the score globally at the audit level, rather than requirement by requirement.

nas-tabchiche (dismissed):
LGTM. To merge once CI passes

nas-tabchiche (approved):

PR #4654, opened 2026-08-12
fix(workflows): fail the send_email node on delivery errors instead of reporting success

nas-tabchiche:
## What & why

The send_email action enqueued mail through the fire-and-forget huey notification task and always reported success: no consumer running, SMTP down, or mailer unconfigured all produced a clean ACTION_EXECUTED log while nothing was sent.

The action now validates its config inside the run, then dispatches delivery to a dedicated huey task: the token parks WAITING, the task sends outside the engine transaction (no instance-tree locks held during SMTP I/O, one connection per batch) and hands the token back. Success advances the run; a delivery failure fails the node and feeds the per-node retry policy. Static config errors (missing settings, no recipients, invalid recipient) are non-retryable and fail the node immediately instead of burning the retry schedule.

Also:

- send_email is not gated on notifications_enable_mailing: that toggle governs digest notifications and the old path never checked it
- the console backend (MAIL_DEBUG) no longer requires EMAIL_HOST/EMAIL_PORT, so workflows are testable in dev/demo setups
- display-name recipients (`Jane <jane@x>`) keep working; validation runs on the addr-spec
- delivery is best-effort across recipients: one dead address no longer starves the rest. A node retry re-sends to already-served recipients; accepted trade-off

Delivery requires the huey worker, same as retries and scheduled triggers. If the worker dies mid-delivery the token stays WAITING until the TTL reaper collects it, same exposure as async subprocess waits.

## Test plan

`automation/workflows/tests/test_send_email_action.py` rewritten for the dispatch flow (14 tests): token parks and dispatches after commit, per-recipient delivery, duplicate task delivery ignored, toggle off still delivers, console backend, missing settings, display-name recipient, invalid recipient, transport failure, partial failure still delivers to the rest, retry policy re-dispatches, non-retryable config errors, zero-sent backend report, no recipients. Full workflows package green locally (305 passed).

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Workflow email actions now support reliable deferred delivery after workflow transactions complete.
  - Multiple recipients can be validated and processed in a single workflow.
  - Optional HTML email content remains supported.

- **Bug Fixes**
  - Email actions prevent sending when mailing is disabled or required settings are missing.
  - Invalid or missing recipients are rejected before delivery.
  - Delivery failures provide clearer details and retry behavior, including for partially completed deliveries.


backend/automation/workflows/actions.py
```diff
@@ -36,7 +38,26 @@
...
 
 class ActionError(Exception):
-    pass
+    """Deliberate action failure. retryable=False marks permanent problems
+    (static config, validation) that no retry can change: the engine fails
+    the node immediately instead of burning the retry schedule."""
+
+    def __init__(self, message, retryable=True):
+        super().__init__(message)
+        self.retryable = retryable
```
monsieurswag: # NIT

It seems weird to me to use a "statefull" Exception (with a `retryable` attribute) instead of have 2 different Exception types.

Having 2 different exception types (`ActionError` and `FatalActionError` for example) would a lot more descriptive.
If 2 exceptions represent 2 significantly different things there should just be separated into 2 different classes.

There's a reason why modern programming languages separate the "error" and "panic" concepts.

backend/automation/workflows/engine.py
```diff
@@ -551,17 +603,20 @@ def _process(token):
                 _advance(token)
             except (ActionError, EngineError) as e:
                 failure = str(e)
+                failure_retryable = getattr(e, "retryable", True)
```
monsieurswag: # TO FIX:

At `backend/automation/workflows/engine.py:604-606`:

Generally we must avoid using `getattr` as much as we can, it makes the code less traceable (harder to debug/refactor) + it makes it less predictable (more error prone).

Something like this would be a lot cleaner:

```py
except EngineError as e:
  failure = str(e)
except ActionError as e:
  failure = str(e)
  failure_retryable = e.retryable
```

backend/automation/workflows/engine.py
```diff
@@ -621,6 +676,25 @@ def _handle_failure(token, message):
...
+    (a new event type would cost a migration): NODE_ENTERED is already
+    written, and the ACTION_EXECUTED/ERROR row lands when the task reports.
+    """
+    token.status = WorkflowToken.Status.WAITING
+    token.save(update_fields=["status", "updated_at"])
+    # on_commit: the WAITING row must be visible before the consumer runs, or
+    # a fast worker finds an ACTIVE token and drops the dispatch.
+    kwargs = {"token_id": str(token.id), **dispatch.kwargs}
+    task = dispatch.task
+    transaction.on_commit(lambda: task(**kwargs))
```
monsieurswag: # TO FIX:

The whole code flow is very hard to follow.

An `execute` method return a `DeferredDispatch` with a function (the `task`) + its `kwargs` (declared as a dict).
Then the engine pass the `DeferredDispatch` to `_defer_action` alongside the `token`.
Then `_defer_action` mutate the `token` + patch up the `kwargs` dict and pass it to the `task`

There's a lot of indirection: storing a function in a variable returned, captured by the caller + passing the variable to function + patching `kwargs` + calling the stored function inside the iniyally returnd variable with the patched `**kwargs`.

---

The `DeferredDispatch` class is future-proofing in case we add some more deferred action later on (which is okay to do as it's likely to happen).

Instead we should make `_defer_action` a method of `DeferredDispatch` (like `DeferredDispatch.dispatch(self, token)`).
Doing so would increase association and reduce code indirection.
(Maybe we should find a better name for the `DeferredDispatch` class too).

We should strictly typed "deferrable tasks" instead, we can document the tasks with docstrings, we have clear well-defined arguments for each task.

Example:

```py
class DeferredTask:
  def __init__(self, defered_task: Callable[..., None], **kwargs):
    """Docstring describing how to create a new defered task."""
    self.defered_task = task
    self.kwargs = kwargs

  def dispatch(self, token: WorkflowToken):
    token.status = WorkflowToken.Status.WAITING
    token.save(update_fields=["status", "updated_at"])

    task = dispatch.defered_task
    transaction.on_commit(lambda: task(self.kwargs, token=token))

class DeferredSendEmailTask(DeferredTask):
  def __init__(self, subject: str, body: str, recipients: list[str]):
    """Docstring describing the meaning of the task args."""
    super().__init__(send_email_task, subject=subject, body=body, recipients=recipients)

# ---------------------

if isinstance(output, DeferredTask):
  output.dispatch(token)
```

backend/automation/workflows/tasks.py
```diff
@@ -31,6 +31,77 @@ def retry_token_task(token_id):
...
+    policy applies — a retry re-sends to every recipient, including the ones
+    already served, accepted over losing the failed ones."""
+    import structlog
+    from django.conf import settings
+    from django.core.mail import get_connection
+
+    from core.tasks import send_email_now
+
+    from .engine import complete_deferred_action, fail_deferred_action
+    from .models import WorkflowToken
```
monsieurswag: # NIT:

Move all imports at the top of the file (except if function-scoped imports are used to avoid circular imports).

backend/automation/workflows/actions.py
```diff
@@ -535,8 +556,23 @@ class SendEmailAction(BaseAction):
...
     def execute(self, config, instance):
-        from core.tasks import send_notification_email
-
+        # Config errors fail the node here; delivery happens in a huey task
+        # (DeferredDispatch) so SMTP I/O never runs while the engine
+        # transaction holds the instance-tree locks. The task resumes or
+        # fails the node, so delivery errors still feed the retry policy.
+        from core.tasks import missing_email_settings
+
+        from .tasks import send_email_task
```
monsieurswag: # NIT:

Move all imports at the top of the file (except if function-scoped imports are used to avoid circular imports).

backend/automation/workflows/engine.py
```diff
@@ -311,6 +312,53 @@ def resume_token(token):
...
+            token.refresh_from_db()
+            if token.status != WorkflowToken.Status.WAITING:
+                return
+            token.instance = instance
+            token.status = WorkflowToken.Status.ACTIVE
+            token.save(update_fields=["status", "updated_at"])
+            _handle_failure(token, message)
+            _run(instance)
+    finally:
+        current_trigger_depth.reset(depth_token)
```
monsieurswag: # NIT:

Is there anyway we create some kind of abstraction to avoid this duplicated code ? the 2 functions have a very similar structure.
The duplicated logic seems kind of important so it could be good idea to create a little abstraction for it.

Also why `complete_deferred_action` and `fail_deferred_action` aren't method in the `WorkflowToken` class anyway ?
After all they're clearly token-centric functions which may purpose is to modify the token and to do some stuff with it.

And always add annotations to functions (both functions mentioned above don't have any).

backend/core/tasks.py
```diff
@@ -710,6 +727,19 @@ def send_notification_email(subject, message, owner_email, html_message=None):
         )
 
 
+def missing_email_settings():
+    # The console backend (MAIL_DEBUG) never opens an SMTP connection, so
+    # EMAIL_HOST/EMAIL_PORT only bind when the backend would use them.
```
monsieurswag: # NIT:

We should rename the function to `get_missing_email_settings` to make it clear it's a function (and a "pure" one).
The `missing_email_settings` variable name make it seem like it's a `list` instead of a function, even if you know this is a function you may wonder if it perform any action on the email settings.

Also always add annotations like `def missing_email_settings() -> list[str]:`.

The comments are really hard to understand, if they've been written in order to make the code clearer (which is a good idea for this function) you should rewrite them.

backend/core/tasks.py
```diff
@@ -710,6 +727,19 @@ def send_notification_email(subject, message, owner_email, html_message=None):
...
+    required_settings = ["EMAIL_HOST", "EMAIL_PORT", "DEFAULT_FROM_EMAIL"]
+    if getattr(settings, "EMAIL_BACKEND", "").endswith("console.EmailBackend"):
+        required_settings = ["DEFAULT_FROM_EMAIL"]
+    return [
+        setting
+        for setting in required_settings
+        if not hasattr(settings, setting) or not getattr(settings, setting)
+    ]
```
monsieurswag, line 748: # TO FIX:

The `getattr(settings, setting)` condition doesn't make sense as it's verifying if string has itself as an existing attribute name.

For example if `"DEFAULT_FROM_EMAIL"` is a requirement setting it will check if the `"DEFAULT_FROM_EMAIL"` string has an `DEFAULT_FROM_EMAIL` string (which obviously doesn't make sense).

backend/core/tasks.py
```diff
@@ -675,6 +675,35 @@ def send_notification_email_expired_eta(owner_email, controls):
         )
 
 
+def send_email_now(subject, message, recipient, html_message=None, connection=None):
+    """Synchronous send that propagates failures to the caller, unlike the
+    fire-and-forget send_notification_email task. Pass `connection` to batch
+    several sends over one SMTP session (the caller owns its lifecycle)."""
```
monsieurswag: # NIT:

Add annotations for argument types.

It's good to specify that "failures are propagated to the caller", but then, you must specify what exception may be raised by the function (`RuntimeError`), otherwise the dev will be forced to jump to the function's code to guess it.

Also a docstring should only document what it actually does (documenting what it doesn't do is useless it just fills the docstring with junk).
So remove `, unlike the fire-and-forget send_notification_email task` from the docstring.

backend/core/tasks.py
```diff
@@ -675,6 +675,35 @@ def send_notification_email_expired_eta(owner_email, controls):
...
+    from contextlib import nullcontext
```
monsieurswag: # NIT:

Move the import to the top of the file.

backend/automation/workflows/tasks.py
```diff
@@ -31,6 +31,77 @@ def retry_token_task(token_id):
     run_instance(token.instance)
 
 
+@db_task()
+def send_email_task(token_id, subject, body, recipients):
```
monsieurswag: # NIT:

Add annotations to function (argument types).

backend/automation/workflows/actions.py
```diff
@@ -545,12 +581,24 @@ def execute(self, config, instance):
             if email.strip()
         ]
         if not recipients:
-            raise ActionError("send_email: no recipients configured")
+            raise ActionError("send_email: no recipients configured", retryable=False)
+        for email in recipients:
+            # Validate the addr-spec only: display-name recipients
+            # ('Jane Doe <jane@x>') delivered before the delivery rework and
+            # must keep working. (Commas inside quoted display names were
+            # already broken by the comma-split above; unchanged.)
```
monsieurswag: # NIT:

This kind of comment belongs to a PR description, or a github comment.
But it shouldn't be in the codebase itself.

backend/core/tasks.py
```diff
@@ -710,6 +727,19 @@ def send_notification_email(subject, message, owner_email, html_message=None):
...
+    if getattr(settings, "EMAIL_BACKEND", "").endswith("console.EmailBackend"):
+        required_settings = ["DEFAULT_FROM_EMAIL"]
+    return [
+        setting
+        for setting in required_settings
+        if not hasattr(settings, setting) or not getattr(settings, setting)
+    ]
+
+
 def check_email_configuration(owner_email, controls):
```
monsieurswag: # NIT:

It may be better to make `check_email_configuration` return an `Optional[str]` (an optional error message) instead of a `bool`.
This would avoid creating a new `missing_email_settings` function (which feels kind of misplaced, i wouldn't expect a "get missing settings" function inside a `task` module).

Then you could something like:

```py
error_msg = check_email_configuration()
if error_msg is not None:
    raise ActionError(
        f"send_email: Bad email configuration ({error_msg!r})",
        retryable=False,
    )
```

monsieurswag (dismissed):

backend/automation/workflows/actions.py
nas-tabchiche: fair, good pattern to generalize throughout the codebase

backend/automation/workflows/engine.py
nas-tabchiche: `DeferredDispatch.dispatch` makes sense indeed

ab-smith (approved):

PR #4655, opened 2026-08-12
feat(workflows): widen the read_objects registry to requirements, risk and validations

nas-tabchiche:
## What & why

Workflows could read 12 models but nothing from the requirements, risk or validation families, which blocks the recurring sweep use cases (overdue requirements, expiring acceptances, stale validations).

Adds requirement_assessment, risk_scenario, risk_acceptance and validation_flow to READABLE_MODELS, and the missing expiry_date on applied_control. RequirementAssessment and ValidationFlow have no name column, which BASE_READ_FIELDS assumed: the base field list is now trimmed to columns the model actually has. A registry integrity test pins that every listed field is a concrete column and every model has a folder FK.

Rows read like API responses, so users familiar with the API don't have to learn a second shape: every key a workflow row emits exists on the model's ReadSerializer with the same shape. Enum fields render display labels, risk levels serialize as the get_*_risk() matrix cells, scoping FKs as {str, id, name} dicts under their API names, and requirement assessments expose name (__str__) plus a requirement subset dict. Filters and order_by stay on the raw columns, with operators gated by column type so SQLite and PostgreSQL always return the same rows.

Sweeps can target one audit or one risk assessment via the compliance_assessment / risk_assessment fields instead of matching every assessment in the folder subtree. Non-assessable header rows are excluded like every other platform aggregate, and range filters on risk levels skip the -1 "not rated" sentinel.

The builder UI is data-driven off the readable-models endpoint, so no frontend changes beyond one i18n key (isScored).

## Test plan

test_read_actions.py grew from 23 to 33 tests: registry integrity, nameless-model reads, deadline sweeps, audit/assessment scoping, unrated sentinels, operator type gating, non-assessable exclusion and API shape parity. Full module green.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Automation workflows can now read requirement assessments, risk scenarios, risk acceptances, and validation flows.
  - Applied control records now include expiry dates.
  - Workflow outputs provide additional requirement metadata and related record names.
  - Added the “Is scored” label to English interface text.

- **Bug Fixes**
  - Improved field validation and filtering for workflow reads, including records without names and expired risk acceptances.


backend/automation/workflows/actions.py
```diff
@@ -498,11 +661,15 @@ def execute(self, config, instance):
 
         queryset = (
             entry["model"]
-            .objects.filter(folder_id__in=_read_scope_folder_ids(instance.folder))
+            .objects.filter(entry.get("base_filter") or Q())
+            .filter(folder_id__in=_read_scope_folder_ids(instance.folder))
             .filter(id__in=authz.viewable_ids(run_identity(instance), entry["model"]))
```
monsieurswag: # To fix:

The `authz.viewable_ids` function isn't scalable for no good reason.
It should return a `QuerySet` instead of a `set`.
If we have 1 million ID it may completely break SQL queries.

Change `authz.viewable_ids` to make it return a `QuerySet` instead.

backend/automation/workflows/actions.py
```diff
@@ -498,11 +661,15 @@ def execute(self, config, instance):
 
         queryset = (
             entry["model"]
-            .objects.filter(folder_id__in=_read_scope_folder_ids(instance.folder))
+            .objects.filter(entry.get("base_filter") or Q())
+            .filter(folder_id__in=_read_scope_folder_ids(instance.folder))
```
monsieurswag, line 707: # To check:

The `_read_scope_folder_ids` function return the list of subfolders (+ the ID of the provided folder itself).
I am kind of afraid this may be a security hole (what is the user doesn't have a `is_recursive=False` `RoleAssignment`) obviously as of now no one should have a non-recursive role assignment.

But out model allow for non-recursive assigment so we should never assume a user always has read permissions over descendant domains.
Check if this doesn't let a user with a non-recursive role assignment to read object from descendant domains it shouldn't have access to.

# NIT:

At `backend/automation/workflows/actions.py:665`

The `_read_scope_folder_ids` function can be simplified by using the `include_self` argument of the `Folder.get_sub_folders` method.

backend/automation/workflows/actions.py
```diff
@@ -935,14 +1102,27 @@ def validate_read_config(node):
...
-            if condition.get("field") not in fields:
+            field = condition.get("field")
+            op = condition.get("op", "eq")
+            if field not in fields:
                 errors.append(
                     (
                         "action_read_invalid_filters",
-                        f"'{condition.get('field')}' is not a filterable field of "
+                        f"'{field}' is not a filterable field of "
                         f"'{config.get('model')}'",
```
monsieurswag, line 1160: # NIT:

Passing a django model (`model.Model`) to string interpolation (like this: `f"'{config.get('model')}'",`) will produce a "weird" string like this:

```py
>>> config = { "model": RequirementAssessment }
>>> print(f"'{config.get('model')}'")
'<class 'core.models.RequirementAssessment'>'
```
We should use something like `config.get("model").__name__` instead (i am not saying this is good looking code, i am just saying we should take the `__name__` of the model instead).

backend/automation/workflows/actions.py
```diff
@@ -397,14 +531,41 @@ def _read_scope_folder_ids(folder):
...
     if field not in allowed_fields:
         raise ActionError(f"read_objects: '{field}' is not a filterable field")
     op = condition.get("op", "eq")
     lookup = _READ_OP_LOOKUPS.get(op)
     if lookup is None:
         raise ActionError(f"read_objects: unknown operator '{op}'")
+    if op not in _allowed_ops(_concrete_field(entry["model"], field)):
+        raise ActionError(
+            f"read_objects: operator '{op}' is not valid for field '{field}'"
+        )
```
monsieurswag, line 603: # FIW:

You can use `{op!r}` instead of `'{op}'` to get the `repr(op)` stringification of `op`, this doesn't change much thing for strings (although it make the f-string a bit cleaner), but using `!r` it a good practice overall for prettier formatting.

backend/automation/workflows/actions.py
```diff
@@ -397,14 +531,41 @@ def _read_scope_folder_ids(folder):
...
+        if field.name == name or field.attname == name:
+            return field
+    return None
+
+
+def _allowed_ops(field):
+    """Operators that make sense for the column's type. An untyped op either
+    crashes at query time or — worse — compiles on both databases with
+    different rows: 'contains' on a boolean LIKEs against 'true'/'false' on
+    PostgreSQL (casts to text) but against 0/1 on SQLite."""
```
monsieurswag: # NIT:

Add annotations + there's the same docstring problem as the one mentionned for `_read_fields`.

backend/automation/workflows/actions.py
```diff
@@ -397,14 +531,41 @@ def _read_scope_folder_ids(folder):
...
 
 
-def _read_condition_to_q(condition, allowed_fields, context):
+def _concrete_field(model, name):
+    """Concrete column by name or attname (FK columns are whitelisted by
+    attname, e.g. compliance_assessment_id)."""
+    for field in model._meta.concrete_fields:
+        if field.name == name or field.attname == name:
+            return field
+    return None
```
monsieurswag, line 550: # NIT:

At `backend/automation/workflows/actions.py:540`

Remove the useless return.

backend/automation/workflows/actions.py
```diff
@@ -397,14 +531,41 @@ def _read_scope_folder_ids(folder):
 }
 
 
-def _read_condition_to_q(condition, allowed_fields, context):
+def _concrete_field(model, name):
+    """Concrete column by name or attname (FK columns are whitelisted by
+    attname, e.g. compliance_assessment_id)."""
```
monsieurswag: # NIT:

At `backend/automation/workflows/actions.py:534`

Use annotations (for both the arguments and the return type).
Like `(model: type[models.Model], field_name: str) -> Optional[models.fields.Field]`

Also choose a better function name `_concrete_field`, bu its name it seems like a private module-level global variable which represent a single "concrete field".

Rename it like `get_field_from_model`, `get_model_concrete_field`, `get_model_field_from_name` or whatever.

backend/automation/workflows/actions.py
```diff
@@ -370,6 +419,91 @@ def _requirements_breakdown(assessment):
...
+            "current_level": lambda s: s.get_current_risk(),
+            "residual_level": lambda s: s.get_residual_risk(),
+            # Subset of the API's FieldsRelatedField dict.
+            "risk_assessment": lambda s: {
+                "str": str(s.risk_assessment),
+                "id": str(s.risk_assessment_id),
+                "name": s.risk_assessment.name,
+            },
+        },
+        "select_related": ("risk_assessment__risk_matrix",),
```
monsieurswag: # NIT:

At `backend/automation/workflows/actions.py:494`

Why making `"select_related"` a tuple ? everything else in the configs are lists so why this exception.
Variable tuples (tuples of different lengths) should only be used when we need hashing.
Replace the `"select_related"` tuples by lists instead.

backend/automation/workflows/actions.py
```diff
@@ -370,6 +419,91 @@ def _requirements_breakdown(assessment):
...
+        # Matrix-index levels store -1 for "not rated" (RiskScenario.save
+        # resets them while proba/impact are unset). Range filters must not
+        # sweep un-assessed scenarios in — same >= 0 guard as
+        # within_tolerance. eq -1 stays available to target unrated rows.
+        "unrated_sentinels": {"inherent_level", "current_level", "residual_level"},
+        # Same JSON the API serializers expose (current_level =
+        # get_current_risk(): the matrix cell dict, "name" for display,
+        # "value" keeping the comparable index). Shadows the raw column in
+        # the output row; the column names stay the filter/order surface,
+        # comparing on the integer index.
```
monsieurswag: # NIT:

At `backend/automation/workflows/actions.py:473-482`

Describing what `"unrated_sentinels"` is a really good idea as it doesn't seem to be trivial and seem very specific.

But the comment is really really hard to understand still.

backend/automation/workflows/actions.py
```diff
@@ -370,6 +419,91 @@ def _requirements_breakdown(assessment):
...
+    "risk_scenario": {
+        "model": RiskScenario,
+        "fields": [
+            "description",
+            "ref_id",
+            "treatment",
+            "inherent_level",
+            "current_level",
+            "residual_level",
+            # ref_id ("R.1") collides across assessments; scope by id instead.
```
monsieurswag: # NIT:

At `backend/automation/workflows/actions.py:470`

This comment seems to be obsolete, delete it.

backend/automation/workflows/actions.py
```diff
@@ -370,6 +419,91 @@ def _requirements_breakdown(assessment):
...
+            "eta",
+            "due_date",
+            # Lets sweeps target one audit instead of every audit in scope.
+            "compliance_assessment",
+        ],
+        # Output-only identifiers: a bare row is unusable without knowing
+        # which requirement of which audit it assesses. Same keys as
+        # RequirementAssessmentReadSerializer: "name" is __str__ (falls back
+        # to the requirement's ref_id — most shipped framework nodes are
+        # ref_id-only), "requirement" a subset of its nested node dict.
```
monsieurswag: At `backend/automation/workflows/actions.py:440-444`

It's a good idea to expose the stringified version of a requirement assessment, the comment is "over-engineered" and hard to understand though.
Also, describing how RequirementAssessment.__str__ is out of scope of any possible comment at this place of the code.

backend/automation/workflows/actions.py
```diff
@@ -370,6 +419,91 @@ def _requirements_breakdown(assessment):
...
+        "fields": [
+            "status",
+            "result",
+            "extended_result",
+            "score",
+            "is_scored",
+            "documentation_score",
+            "eta",
+            "due_date",
+            # Lets sweeps target one audit instead of every audit in scope.
```
monsieurswag: # NIT:

This is a dead comment, remove it.

backend/automation/workflows/actions.py
```diff
@@ -370,6 +419,91 @@ def _requirements_breakdown(assessment):
         "model": EntityAssessment,
         "fields": ["description", "status", "eta", "due_date"],
     },
+    "requirement_assessment": {
+        "model": RequirementAssessment,
+        # create_requirement_assessments seeds an RA row for EVERY node of
+        # the framework, section headers included; every platform aggregate
+        # excludes non-assessable rows, and so do reads.
```
monsieurswag: # NIT:

The comment speaks about `create_requirement_assessments` which is nowhere to be found, it seems to be a dead comment.

If the comment really is a description of the code i don't get it what "seeds" means, what are " section headers", what is a "platform aggregate", and what "and so do reads" means in this context, and what's a non-assessable "row" (i guess it's a RequirementNode/RequirementAssessment) ?

The `"base_filter"` exclude non-assessable requirements, the code is clear, we may need a comment to explain why if the reason isn't obvious, but nothing more.

backend/automation/workflows/actions.py
```diff
@@ -308,27 +336,44 @@ def _requirements_breakdown(assessment):
 READABLE_MODELS = {
     "applied_control": {
         "model": AppliedControl,
-        "fields": ["description", "ref_id", "status", "eta", "priority", "link"],
```
monsieurswag: # NIT:

The `READABLE_MODELS` constant is a dict with (model_name => "workflow model read config").
The "workflow model read config" should be a @dataclass for strong typing, otherwise manipulating the values of `READABLE_MODELS` is confusing (you don't know what key may exist, you don't know the type each value in the config may have, you don't have docstrings to describe what each field of the config mean etc...).

backend/automation/workflows/actions.py
```diff
@@ -290,10 +302,26 @@ def execute(self, config, instance):
...
+# to the API read serializer's shape (display labels, matrix cells, nested
+# FK dicts) so workflow rows read like API responses; the column name stays
+# the filter/order surface, comparing on the raw stored value.
 BASE_READ_FIELDS = ["id", "name", "created_at", "updated_at"]
 
 
+def _read_fields(entry):
+    """BASE_READ_FIELDS trimmed to columns the model actually has (e.g.
+    RequirementAssessment and ValidationFlow have no name), plus the entry's
+    own fields."""
```
monsieurswag: # NIT:

The `_read_fields` function name implies the "reponsability"(action) of this function is to "read fields", this function does NOT do this.
This functions returns the exhaustive list of field names the workflow engine is allowed to read on the `entry["model"]` model.
The function name is very confusing/misleading.
It should be named something like `get_readable_fields` or whatever.

Also annotations SHALL be added, entry isn't typed and there's no return type either.

The docstring isn't gramatically correct english i don't get what `BASE_READ_FIELDS` "trimmed", it's a constant anyway so it cannot "trim" anything.

Also functions SHALL follow PEP8/PEP257 docstring conventions (imperative-style docstrings starting with a description of what the function does).
Here the docstring just say bizarre stuff about `BASE_READ_FIELDS`.

backend/automation/workflows/actions.py
```diff
@@ -290,10 +302,26 @@ def execute(self, config, instance):
...
 # both the serialized output AND the filter/order whitelist — no "__" paths,
-# no relations, so filters cannot tunnel into other objects.
+# no relations, so filters cannot tunnel into other objects. FK fields are
+# listed under their API name and filter on the id value — still no join.
+#
+# Computed output entries may shadow a listed column to reshape its output
+# to the API read serializer's shape (display labels, matrix cells, nested
+# FK dicts) so workflow rows read like API responses; the column name stays
+# the filter/order surface, comparing on the raw stored value.
 BASE_READ_FIELDS = ["id", "name", "created_at", "updated_at"]
```
monsieurswag: # NIT:

9 lines of comments for such a simple constant is wayyy too much.
Plus the comment is false, it says "Explicit registry of models workflows may read" while this content is a list of field names, that's not the same thing at all.

monsieurswag (dismissed):

backend/automation/workflows/actions.py
```diff
@@ -397,14 +531,41 @@ def _read_scope_folder_ids(folder):
...
+    crashes at query time or — worse — compiles on both databases with
+    different rows: 'contains' on a boolean LIKEs against 'true'/'false' on
+    PostgreSQL (casts to text) but against 0/1 on SQLite."""
+    if isinstance(field, BooleanField):
+        return {"eq", "neq", "is_null"}
+    if isinstance(field, (ForeignKey, UUIDField)):
+        return {"eq", "neq", "in", "not_in", "is_null"}
+    if isinstance(field, (DateField, IntegerField, FloatField, DecimalField)):
+        return set(_READ_OP_LOOKUPS) - {"contains"}
+    return set(_READ_OP_LOOKUPS)
```
monsieurswag: # To fix:

As the function currently exists, `field` can be anything (at least this is what the lack of annotation describes).

If so we SHALL only return `return set(_READ_OP_LOOKUPS)` is `field` is of type `models.fields.Field`.

backend/automation/workflows/actions.py
monsieurswag: # To fix:

As the code is right now `_allowed_ops(_concrete_field(entry["model"], field))`, the `_concrete_field` function may return `None`.
But the `_allowed_ops` seems to expect a valid django field (`models.fields.Field`).

If the `_concrete_field` return value is `None` it shouldn't be passed to `_allowed_ops`.

monsieurswag: # To fix:

As the code is right now `_allowed_ops(_concrete_field(entry["model"], field))`, the `_concrete_field` function may return `None`.
But the `_allowed_ops` seems to expect a valid django field (`models.fields.Field`).

If the `_concrete_field` return value is `None` it shouldn't be passed to `_allowed_ops`.

backend/automation/workflows/actions.py
nas-tabchiche: `config["model"]` here is a string

ab-smith (approved):

PR #4667, opened 2026-08-17
fix(api): answer the metadata action so OPTIONS stops returning 500

### LINKED ISSUE #4653: OPTIONS` on any `BaseModelViewSet` endpoint returns 500 (`SerializerFactory` returns `None` for the `metadata` action)
## Describe the bug

Any `OPTIONS` request to an endpoint served by `BaseModelViewSet` returns
**500 Internal Server Error** with `TypeError: 'NoneType' object is not callable`.

Root cause: `SerializerFactory.get_serializer()` in `backend/core/serializers.py`
maps a serializer only for a fixed set of actions and returns `None` for
anything else:

```python
def get_serializer(self, base_name: str, action: str):
    if action in ["list", "retrieve"]:
        serializer_name = f"{base_name}ReadSerializer"
    elif action in ["create", "update", "partial_update", "destroy"]:
        serializer_name = f"{base_name}WriteSerializer"
    else:
        return None
    return self._get_serializer_class(serializer_name)
```

When DRF handles `OPTIONS`, `SimpleMetadata.determine_actions()` calls
`view.get_serializer()` while `view.action == "metadata"`. That action is not in
either branch, so `get_serializer_class()` returns `None` and DRF then calls
`None(...)`.

Because the affected code is in the shared base viewset, **every** resource is
affected, not one in particular.

## To Reproduce

1. Authenticate against the API.
2. Send `OPTIONS` to any collection or detail endpoint, e.g.
   `OPTIONS /api/folders/`, `OPTIONS /api/perimeters/`, `OPTIONS /api/assets/`.
3. The response is `500`. `GET` on the very same URL returns `200`.

Backend traceback (abridged):

```
rest_framework/metadata.py  determine_actions()
  locals: actions={}, method='PUT', self=<SimpleMetadata>, view=<FolderViewSet>
rest_framework/generics.py:114 in get_serializer
  return serializer_class(*args, **kwargs)
  locals: serializer_class = None, context={'action': 'metadata'}
TypeError: 'NoneType' object is not callable
```

## Expected behavior

`OPTIONS` should return `200` with the resource metadata, as DRF does by
default. At minimum it should not raise: an unmapped action should fall back to
a sensible serializer rather than `None`.

## Actual impact on the UI

This is not only an API-level annoyance. In our deployment the web UI could not
delete a domain: clicking the delete (trash) control produced **no request at
all** — verified in the backend logs, where no `DELETE` was ever received. The
deletion succeeded immediately when issued directly against the API
(`DELETE /api/folders/<id>/` → `204`), which shows the write path itself is
healthy and only the metadata step is broken.

## Environment

- CISO Assistant Community Edition **v3.19.2** (backend and frontend images)
- PostgreSQL 16, Docker Compose deployment, self-hosted
- Browser: Firefox (also reproducible headlessly, straight against the API)
- **Also present in `main`** at the time of writing: the `get_serializer` body
  quoted above is unchanged, so upgrading to v3.21.x does not fix it.

## Additional context

We could not find an existing issue mentioning `SerializerFactory`,
`get_serializer_class` or a 500 on the metadata endpoint.

A possible fix, keeping the current design: return the read serializer (or a
plain `ModelSerializer` on the viewset's model) for unmapped actions, instead of
`None`. Alternatively, override `metadata_class` so `determine_actions()` does
not need a serializer for actions the factory does not know about.

luantaraschi:
## What & why

`OPTIONS` on any endpoint served by `BaseModelViewSet` answers 500. `GET` on the same URL answers 200.

DRF handles `OPTIONS` through `SimpleMetadata.determine_actions()`, which calls `view.get_serializer()` while `view.action` is `"metadata"`. `BaseModelViewSet.get_serializer_class` resolves through `SerializerFactory`, and that factory maps a fixed list of actions and returns `None` for anything else, so DRF ends up calling `None(...)`:

```
File "rest_framework/metadata.py", line 92, in determine_actions
    serializer = view.get_serializer()
File "rest_framework/generics.py", line 114, in get_serializer
    return serializer_class(*args, **kwargs)
TypeError: 'NoneType' object is not callable
```

Because the factory lives in the shared base viewset, every resource is affected rather than one in particular.

The change adds `"metadata"` to the branch that resolves the write serializer. `OPTIONS` describes the shape a client may send, so that is the serializer DRF should be describing.

Read only viewsets never reach the factory at all. `determine_actions()` only asks for a serializer for `PUT` and `POST`, so `ActorViewSet`, which limits `http_method_names` to `get`, `head` and `options`, does not go through it. I checked that one because `Actor` has no write serializer: `OPTIONS /api/actors/` already answers 200 on `main`, so there is no case here where the new branch would look for a serializer that does not exist.

Closes #4653.

## Test plan

New file `backend/app_tests/api/test_api_options_metadata.py`, one parametrised test sending `OPTIONS` to the folders, assets and perimeters collections as an authenticated admin.

All three fail on `main` with the `TypeError` above and pass with the change.

```
uv run pytest app_tests/api/test_api_options_metadata.py
```

Ran on Python 3.14 in a Linux container, matching the version in `pyproject.toml`. `ruff format` reports both touched files already formatted, and `ruff check` reports the same 284 pre-existing findings in `core/serializers.py` before and after, with none added by the new test file.

The whole `app_tests/api` folder was run as well: 1741 passed, 9 skipped, 1 xfailed and 1 failed. The failure is `test_api_posture.py::TestPosturePermissions::test_results_scoped_to_viewable_assets`, which ends in `django.db.utils.OperationalError: parser stack overflow` on SQLite. It fails the same way on `main` without this change, so it is not related to it.

## Checklist

### Backend

- [x] Migrations generated with `makemigrations` and committed (or none needed) — no model change, so none needed

### Tests & documentation

- [x] Tests added or updated and passing locally

## Summary by CodeRabbit

* **New Features**
  * Added support for metadata requests.
  * Added score-override status to requirement assessment imports and exports.
  * Limited document attachment endpoints to supported read-only methods: GET, HEAD, and OPTIONS.

* **Bug Fixes**
  * Improved related-field handling for serialized data.
  * Improved requirement assessment score updates when manual score overrides are enabled or disabled.

* **Tests**
  * Added coverage confirming authenticated OPTIONS requests return successfully across collection endpoints.


ab-smith:
@luantaraschi thank you, 
in order to start the review please add the following comment for the CLA bot:

luantaraschi:

luantaraschi:
recheck

ab-smith (approved):

PR #4670, opened 2026-08-17
fix(data-wizard): imported audit answers ignored when template exported in non-native locale

Axxiar:
## What & why

Importing an audit through the Data Wizard silently dropped questionnaire answers whenever the file's language didn't match the framework's authoring language.

What this PR does:
- Match imported answers against both the framework's base-language labels and the current UI language ones (question texts and choice values).
- Normalize the comparison (whitespace collapse + casefold); stored answer values are untouched.
- Log a warning whenever an answer line matches no question or no choice, instead of dropping it silently.

## Test plan

- Load a French-native questionnaire framework with EN translations
- Download its Excel template under an English UI
- Fill an answers cell (<question> >> <choice>)
- Import -> answers must appear on the requirement in the audit (not skipped)

- Same import with French labels (or framework base language) -> still matches
- Answer with extra spaces / different case -> still matches

## Checklist

- [x] Regression test added for bug fixes

## Summary by CodeRabbit

* **Improvements**
  * Compliance questionnaire imports now recognize question and answer labels regardless of capitalization or extra spacing.
  * Imports support both translated and base-language labels.
  * Multiple-choice and text answers are handled more reliably.
  * Unrecognized questions or answer choices now generate warnings instead of being silently skipped.
  * Template hints are excluded from imported answers.


Axxiar:
Will fix in another PR. We need to take into account every translation of a question in every locale of the framework.

PR #4671, opened 2026-08-18
fix(iam): scim user migration to keep local login

melinoix:
## What & why

Follow-up to #4658, addressing review feedback from Abder on three gaps left by the SCIM/SSO hardening:

- **Lockout risk**: existing SCIM-managed accounts without `keep_local_login` would lose local login/password reset on deploy with no fallback if SSO wasn't set up. Added a migration to grandfather them (`keep_local_login=True`).
- **Bypassable guard**: the check blocking SSO disablement while SCIM users depend on it only matched `is_enabled: false` explicitly — a request simply omitting the field slipped through. Now treats a missing field the same as `false`.
- **Password wipe bypass**: enabling Force SSO Login wiped passwords via `bulk_update`, which skips `save()` and therefore skips the exemption that protects SCIM-managed accounts. Excluded them explicitly from that query.

## Test plan

- `uv run pytest app_tests/api/test_api_scim.py` — 21/21 passed
- Migration applied cleanly on a fresh DB; `makemigrations --check --dry-run` → no changes detected
- `ruff format --check` clean

## Checklist

### Backend

- [x] Migrations generated with `makemigrations` and committed
- [x] No new dependencies

### Tests & documentation

- [x] Tests added or updated and passing locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabling SSO now works when the setting is missing or left unset.
  * Existing protections for SCIM-managed users remain unchanged.

* **Improvements**
  * Existing SCIM-managed users who previously could not log in locally can now retain local login access.


backend/iam/sso/views.py
```diff
@@ -75,7 +75,9 @@ def update(self, request, *args, **kwargs):
         force_sso = serializer.validated_data.get("force_sso", False)
 
         if is_enabled and force_sso:
-            users = list(User.objects.filter(keep_local_login=False))
+            users = list(
+                User.objects.filter(keep_local_login=False, is_scim_managed=False)
+            )
```
monsieurswag: When the "force_sso" setting is enabled, every user should be forced to log in with SSO.
The only way to break this rule is to set to `keep_local_login` to `True`.

Why would let a SCIM managed user with `keep_local_login=False` be able to bypass SSO if "force_sso" is enabled ?

To me it seems like a mistake.

monsieurswag (changes requested):

backend/iam/sso/views.py
melinoix: you are right, check new commit

eric-intuitem:
will be included in a more comprehensive PR

PR #4676, opened 2026-08-20
fix: validation issue on metric sample

melinoix:
## What & why

`CustomMetricSample.value` (a JSONField) was never validated on write. A sample created via
the API with a bare number (e.g. `1.0`) instead of the expected `{"result": 1.0}` /
`{"choice_index": 1}` shape crashed `display_value()`/`raw_value()` with an `AttributeError`
on read,  and since the Metric Instances list serializes every row, one bad sample took down
the *entire* list (500) for every user, silently rendered as an empty table by the frontend.

- `metrology/models.py`: `raw_value()`/`display_value()` now guard against a non-dict `value`,
  return `"N/A"` (consistent with the rest of the app), and log a warning instead of crashing.
- `metrology/serializers.py`: `CustomMetricSampleWriteSerializer.validate_value()` rejects a
  non-dict `value` at write time, so new malformed samples can't get created going forward.

## Test plan

Reproduced the exact crash locally (`CustomMetricSample.value = 1.0`), confirmed
`GET /api/metrology/metric-instances/` returned 500 before the fix and 200 after, with
`current_value`/`raw_value` as `"N/A"` for the affected row. Confirmed `POST` with a bare
number now returns 400 instead of 201.

## Checklist

- [x] Regression test added for bug fixes _(if relevant)_ — TODO: add one before merging
- [ ] `product-docs/` updated — not needed, internal defensive fix

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed metric sample values to prevent errors when viewing or processing results.
  * Displays “N/A” when a sample value is not in the expected format.
  * Rejects invalid qualitative and quantitative values, including non-positive indexes, non-numeric results, booleans, unexpected fields, and incorrect value types.
  * Validation now applies consistently when creating or updating samples, including when existing values are retained.
  * Ensures required sample values are provided and match the selected metric category.

* **Tests**
  * Expanded coverage for valid and invalid metric values, malformed payloads, and category-specific validation.


backend/metrology/serializers.py
```diff
@@ -143,6 +143,13 @@ def validate_timestamp(self, value):
             )
         return value
 
+    def validate_value(self, value):
+        if not isinstance(value, dict):
+            raise serializers.ValidationError(
+                'value must be an object, e.g. {"result": 1.0} for a quantitative '
+            )
+        return value
```
monsieurswag: # To fix

We SHALL use a "json schema" for validation (using the `jsonschema` library) instead of a simple `isinstance(value, dict)` check.

As of now, the serializer still allow for an invalid `value`.
For example the `{"invalid_key": "invalid_value"}` invalid value would still be accepted by the serializer.

Here's an example of a json schema that we could use for a strict-ebough validation.
(it was generated by chatGPT so if you end up using it ensure it's correct first).

```py
QUALITATIVE_VALUE_SCHEMA = {
    "type": "object",
    "properties": {
        "choice_index": {
            "type": "integer",
            "minimum": 1,
        },
    },
    "required": ["choice_index"],
    "additionalProperties": False,
}

QUANTITATIVE_VALUE_SCHEMA = {
    "type": "object",
    "properties": {
        "result": {
            "type": "number",
        },
    },
    "required": ["result"],
    "additionalProperties": False,
}

VALUE_SCHEMA = {
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "oneOf": [
        QUALITATIVE_VALUE_SCHEMA,
        QUANTITATIVE_VALUE_SCHEMA,
    ],
}
```

backend/metrology/models.py
```diff
@@ -291,6 +294,16 @@ def raw_value(self):
         else:
             value_dict = self.value
 
+        if not isinstance(value_dict, dict):
+            logger.warning(
+                "CustomMetricSample.value is not an object, expected "
```
monsieurswag: # NIT:

Replace the term `"object"` with `"dict"`, the `isinstance(value_dict, dict)` check if `value_dict` is a "dict", not if it's an "object".

`CustomMetricSample.value` will always be an "object" (because everything in a python is an object, for example `isinstance({anything}, object)` will always return `true`).

monsieurswag (dismissed):
Ideally we would add a JSON validator to the `JSONField` directly, but i get the idea to avoid stacking up too much migrations.
And putting the logic directly in the serializer isn't so bad.

If we want to avoid this kind of problem this PR fixed in the future we should apply what's been said in PR#4322

Image

martinzerty (approved):
LGTM

PR #4681, opened 2026-08-22
fix(ui): search filter misbehaving for non-latin characters 

Punkoivan:
## What & why
Fixed regexp, used in filtering entities while searching. Previous one didn't filter non-Latin correctly.

## Test plan

Withouts changes it works for Latin like this:
Screenshot_20260808_160701-1

And fix makes it for Cyrillic like Latin (filtering correctly):
Screenshot_20260808_160733

## Checklist

- [х] PR title follows Conventional Commits (`!` set if this is a breaking change)

lint failed due to unresolved dev dependency though, expect it to run correctly via CI.

- [x] No tests or docs needed for this change — why:
no docs needed since it's small bug fix for non-latin elements.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search normalization to support Cyrillic and other non-Latin letters and numbers.
  * Search now handles punctuation and diacritics more consistently.
  * Cyrillic option labels and search terms correctly return matching results.

* **Tests**
  * Added coverage for multilingual search normalization and filtering.


monsieurswag (approved):
LGTM, thanks for the contribution that's a good fix.

PR #4682, opened 2026-08-23
fix: regression on flash mode processed items counter

ab-smith:
## Summary by CodeRabbit

* **Bug Fixes**
  * Flash mode now immediately reflects answer, result, and observation updates while navigating between assessments.


eric-intuitem (approved):
LGTM

PR #4491, opened 2026-07-10
feat(epic): workflow engine with visual builder

nas-tabchiche:
## Summary by CodeRabbit

* **New Features**
  * Added workflow automation with a visual builder for triggers, actions, branching, loops, variables, secrets, schedules, and subprocesses.
  * Added workflow versioning, publishing, activation, execution monitoring, retries, logs, and run replay.
  * Added webhook, scheduled, and internal-event triggers with validation and rate limiting.
  * Added YAML import/export and workflow library preview and instantiation.
  * Added workflow permissions, localized interface text, graph previews, and filtering tools.
* **Security**
  * Added scoped authorization, secret protection, webhook signing, and SSRF safeguards.
* **Documentation**
  * Added a versioned workflow schema for portable imports and exports.

PR #4683, opened 2026-08-24
feat(workflows): make the engine able to write, schedule by date, and see bulk edits

ab-smith:
## What & why

The visual workflow engine (#4491) shipped write-once and date-blind. This closes the four enablers that block the sample libraries: a workflow could create objects and send mail, but could not change anything that already existed, could not express a deadline, and could not see bulk edits at all.

- **`update_object` + relation writes** — a new `UPDATABLE_MODELS` registry over 14 of the 16 readable models. Two whitelists per entry: `fields`, and `allowed_values` for fields that are mostly off limits. Relations take `add | remove | set`, matching `batch_action`'s vocabulary. Targets must be inside the workflow's subtree **and** changeable by the run identity (`authz.changeable_ids`, the write counterpart of `viewable_ids`), so a run can never write a row the API would refuse that user.
- **Date context** — `{{now}}` / `{{today}}` seeded once at run start, in the schedule's own timezone when a schedule fired the run. A new `date_offset` action (`base`, `days`, `weeks`, `output`) covers arbitrary windows. No grammar change: the template language stays frozen, as `schema/workflow-v1.schema.json` promises.
- **Bulk writes emit events** — `bulk_update` skips `post_save`, so auditlog logged nothing and no workflow ever fired. `core.utils.bulk_update_with_log` writes the entries a per-object `save()` would, applied to the two paths that change an audit a user already had (`map_from`, and the `documentation_score` seeding in `perform_update`).
- **Coalescing** — that fix makes one merge able to start hundreds of runs, so a trigger now starts at most one run per (trigger, auditlog correlation id) inside a five-minute window.

**The integrity line, enforced by the registry rather than by convention:** automation may record that time passed, and may attach work, but may not render the judgment. Never `RequirementAssessment.result`/`score`, never `RiskScenario.treatment`, `Evidence` and `SecurityException` statuses fenced to expiry, `Finding` to everything but `dismissed`, and no entry may write `name`. `RiskAcceptance` and `ValidationFlow` are absent entirely: their state moves through `set_state()` and the write serializer's transition table plus `FlowEvent`, not through `save()`, so a column write would skip `revoked_at`, the scenario treatments a revocation reverts, and the flow's own history.

Deliberately out of scope: a delete action, error routing, and per-trigger throttle windows.

Closes #

## Test plan

- `pytest backend/automation/workflows/tests` — 392 passing, of which 64 are new across three modules: `test_date_context.py`, `test_update_actions.py`, `test_event_coalescing.py`.
- Run together with `backend/core/tests/test_requirement_score_override.py` (the paths now routed through `bulk_update_with_log`): 412 passing. Worth running them together — a correlation-id assertion that passed in isolation failed in the wider run, because auditlog's cid is a `ContextVar` an earlier API test leaves set.
- Coverage worth calling out: the guardrail refusals (result, score, treatment, approvals, values outside a column's choices), scope refusals (row outside the subtree, relation target outside scope), a create-then-link graph proving controls are no longer orphaned, a scheduled sweep filtering a date column against a computed horizon, and a reused correlation id past the window still earning a new run.
- Frontend: `pnpm run check` clean; builder exercised for the `update_object` and `date_offset` config panels.

## Checklist

### Backend

- [x] Migration generated and committed — `automation/0005`: one column, one index, no data migration
- [x] No new dependencies

### Frontend

- [x] Svelte 5 syntax; formatted
- [x] New UI strings added to `frontend/messages/en.json`
- [x] Clicked through the builder for both new action types

### Tests & documentation

- [x] Tests added and passing locally
- [ ] `product-docs/` — the module has no page yet; documenting it is the first item of the samples branch that follows this one

## Summary by CodeRabbit

- **New Features**
  - Added workflow actions for updating records, managing relationships, and calculating date offsets.
  - Added configurable field permissions, value validation, updateable-model metadata, and protected workflow variables.
  - Added run timestamps, timezone-aware date context, and correlation tracking.
- **Bug Fixes**
  - Duplicate events are now coalesced within five minutes and clearly marked.
  - Bulk updates now record changes consistently in audit logs.
- **UI Improvements**
  - Expanded workflow editor configuration, previews, translations, and trigger-result indicators.


nas-tabchiche (approved):
LGTM

PR #4698, opened 2026-08-26
fix: splash screen node invariants and auditee dashboard scoping

nas-tabchiche:
## What & why

Splash screen nodes must never be assessable. The framework builder left `assessable: true` behind when an existing requirement node was switched to splash mode, and hid the checkbox, so the stale flag was invisible. Frameworks published this way show splash nodes as regular requirements (red dot instead of purple) in the respondent view. Found while troubleshooting a customer framework built entirely with the library builder.

- builder: switching a node to splash now clears `assessable`
- loader: forces `assessable: false` on splash nodes at import; draft validation warns the author
- data migration heals already-loaded frameworks
- assignments page: splash nodes were filtered out with their whole subtree, so on frameworks nesting requirements under splash screens the tree lost all requirements and "Select all" cleared the selection instead. Splash children are now hoisted in place of the splash node
- auditee dashboard: only lists assignments the user can open. It selected by actor match alone, so a user with audit access but without `view_requirementassignment` got a card whose "continue assessment" click 404'd

## Test plan

- `uv run pytest core/tests/test_splash_assessable_invariant.py core/tests/test_auditee_dashboard_scoping.py library/tests/test_library_builder.py`
- `pnpm vitest run src/lib/components/FrameworkBuilder/builder-state.test.ts` (75 passed)
- Reproduced all three symptoms with the customer framework before the fix; assignments tree hoisting still needs a click-through against it

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Framework editing automatically clears assessability when a requirement changes to splash mode.
  * Splash-mode sections now display their child requirements in assessment views.
  * Validation warns when splash requirements are marked assessable.
* **Bug Fixes**
  * Dashboard results exclude assignments the viewer cannot open.
  * Existing splash requirements are corrected during data updates.
  * Switching display modes no longer incorrectly restores assessability.


eric-intuitem (approved):
LGTM

PR #4713, opened 2026-08-27
fix: adjustments to allow multiple journeys on the same domain

ab-smith:
## Summary by CodeRabbit

* **New Features**
  * Presets can be applied multiple times to the same folder, creating separately numbered journeys and objects.
  * Journey upgrades preserve previously tracked objects and renamed journeys.
  * Preset versions now reflect the latest available version.

* **Bug Fixes**
  * Applying presets no longer unexpectedly recreates tracked objects.

* **Improvements**
  * “Create objects” and “Apply feature flags” are now unchecked by default when applying a preset.


eric-intuitem (approved):
LGTM

PR #4722, opened 2026-08-30
fix(api): allow PATCH on requirement nodes with blank nullable fields

### LINKED ISSUE #4715: Requirement node PATCH always fails when implementation_groups is null
A PATCH to `/api/requirement-nodes/{id}/` is rejected for any node whose `implementation_groups` is null, even when that field is not part of the update.

```
PATCH /api/requirement-nodes/{id}/  {"name": "..."}
400  {"implementation_groups": ["This field cannot be blank."]}
```

**Scope: this is not specific to one library.** `implementation_groups` is declared as `models.JSONField(null=True, ...)` with no `blank=True`, so DRF treats it as required and rejects an empty value on partial update. Every requirement node in every framework that does not define implementation groups is therefore un-PATCHable, which is a large share of the stock catalogue rather than an edge case.

`OPTIONS` on the same endpoint advertises `GET, PUT, PATCH, DELETE, HEAD, OPTIONS`, so the operation is offered but unreachable for those nodes.

Sending the field explicitly does not help. All three are refused identically:

| payload | result |
|---|---|
| `{"name": "...", "implementation_groups": []}` | 400, same error |
| `{"name": "...", "implementation_groups": null}` | 400, same error |
| `{"name": "...", "implementation_groups": [""]}` | 400, same error |

**Reproducing:** load `urn:intuitem:risk:library:nist-csf-2.0` and PATCH any of its requirement nodes. All 134 have `implementation_groups: null`, so all 134 are affected. Any other library without implementation groups should behave the same way.

**Suggested fix:** treat `implementation_groups` as optional on partial update, either by adding `blank=True` to the model field, or by declaring it `required=False, allow_null=True` on `RequirementNodeWriteSerializer`, which is currently `pass`.

**Version:** v3.21.2, build 97e68c700.

---

For context on how we hit it: the stock CSF 2.0 library leaves `name` blank on all 106 subcategories, so a requirement list shows ref_ids and descriptions only. We maintain short titles and re-apply them after each upgrade, since an upgrade re-seeds stock libraries. With the API closed we write to the database directly, which we would rather not keep doing. That is our use case; the defect above is general.

AmirF194:
## What & why

`RequirementNodeWriteSerializer.update()` calls `instance.full_clean(exclude=m2m_field_names)` on every PATCH, to run `RequirementNode.clean()` for the min/max score override checks added in #4173. `full_clean()` also runs `clean_fields()` on every field on the instance, not just the ones in the request, so any node with an untouched field that is `null=True` without `blank=True` (`implementation_groups`, `order_id`, and `ReferentialObjectMixin.name`, which `RequirementNode` inherits) rejects the PATCH with `"This field cannot be blank."`, even when the request never mentions that field. Since most stock framework nodes ship with `implementation_groups: null` (all 134 nodes in NIST CSF 2.0, per the issue), this makes those nodes un-PATCHable entirely.

The fix excludes fields absent from `validated_data` (plus the already-excluded M2M fields) from `clean_fields()`, so a PATCH only validates what it actually sends. `clean()` still runs against the full instance, so the score-override cross-field checks from #4173 are unaffected.

Closes #4715

## Test plan

- Added `test_patch_survives_when_untouched_nullable_fields_are_blank`: creates a `RequirementNode` with `name`, `order_id`, and `implementation_groups` all null (as stock framework nodes are), PATCHes only `ref_id`, and asserts it succeeds. Fails on `main` with the `"This field cannot be blank"` error from all three fields, passes on this branch.
- Added a second test confirming the serializer still rejects a PATCH that sets an invalid `scores_definition_ref`, so `clean()`'s cross-field validation still runs.
- `core/tests/test_implementation_groups.py` (11 tests) and `core/tests/test_requirement_score_override.py` (20 tests) still pass.
- `ruff format --check` clean.
- Not verified: the frontend requirement-node edit form, which I did not exercise end to end.

## Summary by CodeRabbit

* **Bug Fixes**
  * Partial updates no longer fail because of validation issues in unrelated, untouched fields.
  * Modified fields continue to be validated, with invalid values rejected.
  * Optional fields can be explicitly set to `None` during updates.
  * Untouched values remain unchanged.

* **Tests**
  * Added regression coverage for partial updates with untouched nullable fields.
  * Added coverage confirming modified fields are validated and invalid changes are rejected.
  * Added coverage for explicitly clearing optional fields.


AmirF194:

AmirF194:

AmirF194:
recheck

ab-smith (approved):

PR #4724, opened 2026-08-31
fix(audit-export): make questions and answers round-trip correctly

nas-tabchiche:
## What & why

Questions and answers were only usable in two of the audit export formats, and the XLSX round-trip lost data.

The CSV dropped questions entirely. The XLSX carried them and the data wizard could read them back, but that round-trip matched on translated question text and split answer blocks on newlines. So a multi-line free-text answer lost everything after its first line, a workbook exported in one language would not re-import under another, and anything unmatched was dropped silently.

Answer lines now carry the question URN:

```
[urn:...:question:1] Do you encrypt backups? >> Yes
```

Blocks are delimited by that URN instead of by newlines, and choices are matched against every translation. Skipped lines come back as warnings instead of vanishing. Questions hidden by an unsatisfied `depends_on` are no longer exported as unanswered gaps. The HTML report now orders questions by `Question.order` rather than by an unordered `Answer` queryset. The CSV gained an `answers` column.

Render and parse live together in `core/utils.py`, so the CSV, the XLSX and the importer cannot drift apart.

Two behaviour changes worth a look:

- the CSV gains a column, which is breaking for anything scripted against that export
- an answer cell emptied by hand now clears the stored answer, where before it was ignored

Workbooks exported before this change still import, since the old text matching is kept as a fallback.

## Test plan

- 23 new tests in `core/tests/test_answers_cell.py` and `core/tests/test_audit_export_answers.py`, covering the multi-line truncation, an FR export re-imported under EN, clearing, hint skipping, both warning paths, legacy cells and hidden questions
- `uv run pytest core/tests/` 403 passed, 1 skipped
- `uv run pytest data_wizard/tests/` 327 passed, 2 skipped
- `uv run python manage.py check` clean
- `./reformat.sh` produced no changes

`test_rendering_does_not_scale_with_requirement_count` pins the query count for the new CSV column, so the export cannot regress into an N+1.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Requirements exports now include answers in CSV and spreadsheet formats.
  - Answer data can be imported from exported spreadsheet cells, including text and choice responses.
  - Import warnings identify unknown or unmatched answers.
  - Reports hide questions whose dependency conditions are not met.
  - Export descriptions now mention included answers.

- **Bug Fixes**
  - Answers are displayed in the defined question order.
  - Improved handling of unanswered, translated, multiline, and legacy answer data.


melinoix (approved):
LGTM will probably need in future some data wizard work about separators, character escaping etc. For the scope of the PR, this is good

PR #4728, opened 2026-09-01
feat: context menu action to quickly add custom-metric-sample

Axxiar:
## What & why

Add an action to the context menu on `/metric-instances/` to quickly change its value instead of having to navigate on that metric instance and click "Add new custom metric sample"

image

## Test plan

1. Go to `/metric-instances/` and create one
2. In the List view, right click on it
3. There should be an "Add new custom metric sample" action
4. Clicking it should open the Custom Metric Sample CreateModal and allow you to create one. 

## Checklist

- [X] Svelte 5 syntax; `pnpm run lint` and `pnpm run format` run
- [X] New/changed UI strings added to `frontend/messages/en.json` (keep keys, preserve `{tokens}`)
- [X] Clicked through the change in the dev server; screenshots attached for UI changes

- [X] No tests or docs needed for this change — why: small existing action that just got a new quicker way to trigger it.

## Summary by CodeRabbit

* **New Features**
  * Added a context-menu option for entering custom metric samples.
  * Expanded table filters, columns, and batch actions across commitments, findings, campaigns, assessments, entities, and scores.
  * Added findings context-menu actions and bulk updates for supported assessment, label, status, conclusion, and commitment fields.

* **Bug Fixes**
  * Improved handling of qualitative and quantitative metric values.
  * Prevented invalid qualitative selections from being assigned.
  * Improved feature-flag handling for available table fields and batch actions.


frontend/src/lib/components/Forms/ModelForm/CustomMetricSampleForm.svelte
```diff
@@ -108,7 +109,7 @@
 		const value = (event.target as HTMLInputElement).value;
 		const numValue = parseFloat(value);
 		if (!isNaN(numValue)) {
-			$valueFieldProxy = JSON.stringify({ result: numValue });
+			$valueFieldProxy = { result: numValue };
```
monsieurswag, line 112: # NIT:

There is an inconsistency, replace `isNaN` by `Number.isNaN` in this file (for the code to remain consistent).

frontend/src/lib/components/ContextMenu/metric-instances/EditValue.svelte
```diff
@@ -0,0 +1,60 @@
...
+	let { row }: Props = $props();
+
+	const modalStore: ModalStore = getModalStore();
+
+	async function openAddSampleModal() {
+		const id = row?.meta?.id;
+		if (!id) return;
+		const initialData = {
+			metric_instance: id,
+			folder: row.meta.folder?.id ?? row.meta.folder,
```
monsieurswag: Why it fallback to `row.meta.folder` ?
Wouldn't `row.meta.folder` be an object, if folder.meta.folder.id may exist then i guess `folder.meta.`folder would be something like `{id: ..., ...}` right ?
Am i wrong ?
Check if this is not a mistake.

frontend/src/lib/components/ContextMenu/metric-instances/EditValue.svelte
```diff
@@ -0,0 +1,60 @@
...
+	import CreateModal from '$lib/components/Modals/CreateModal.svelte';
+	import {
+		getModalStore,
+		type ModalStore,
+		type ModalComponent,
+		type ModalSettings
+	} from '$lib/components/Modals/stores';
+
+	interface Props {
+		row: any;
```
monsieurswag, line 17: # FIW:

Using `any` is very bad, but this is the pattern we use in a lot of existing `ContextMenu` components.
So let's keep it (it's better to have very consistent bad code than inconsistent mostly-bad code (20% good, 80% bad)).

But don't forget that `any` is usually very bad.

monsieurswag (changes requested):
There are 2 problems (not caused by this PR), but these should be fixed by this PR.

Right now when we go to the `/custom-metric-samples/{id}` frontend page (detail view) there are 2 problems:
- The `raw_value` is displayed as `[object Object]`
- The `displayTitle` in `frontend/src/routes/(app)/+layout.svelte`/`enterprise/frontend/src/routes/(app)/+layout.svelte` is set to the ID(UUID) of the custom-metric-instance object.

The former one (the second problem) may be okay so it's not mandatory for you to fix it, but the first one MUST be fixed before merging this PR.

Image

frontend/src/lib/components/ContextMenu/metric-instances/EditValue.svelte
Axxiar: Any == bad, noted :saluting_face:

frontend/src/lib/components/ContextMenu/metric-instances/EditValue.svelte
Axxiar: even looks like folder is not useful here. it was a mimicry from elsewhere. removing it

Axxiar:
>   The `displayTitle` in `frontend/src/routes/(app)/+layout.svelte`/`enterprise/frontend/src/routes/(app)/+layout.svelte` is set to the ID(UUID) of the custom-metric-instance object.

What would be a better fit ? Custom metric instance don't have name or any user friendly identifier. 

*Edit : After discussion, staying with UUID for this PR.

monsieurswag (dismissed):

ab-smith (approved):

PR #4731, opened 2026-09-01
fix(custom-fields): stale values from another domain blocked object creation

nas-tabchiche:
## What & why

Creating an object failed with an opaque `custom_fields` error when the selected domain was not the one holding the custom field definitions (support ticket, reproduced on security exceptions). Three things combined:

- the create modal loads definitions before a domain is chosen, so `/custom-fields/` is queried without `for_folder` and returns every definition of the model
- rendering a choice field writes a `null` placeholder into the payload (Svelte select binding on mount)
- switching domain hides the fields but keeps the keys, and the API rejects them as unknown

Fix: the payload only ever carries keys of currently rendered definitions. The section always resolves definitions against the selected folder (none selected = global definitions only, new `for_folder=` empty semantics) and prunes values whose definition no longer applies, untainted and skipped when a fetch fails (an inline notice shows instead of another folder's fields). The API tolerates empty values on keys that exist for the model but not for the target folder; real values there and typo'd keys are still rejected.

Behavior changes:
- moving an object to another folder now discards stored values whose definition no longer applies there (previously they lingered: invisible in forms, still matching `cf__` filters), and the destination folder's required custom fields are enforced on the move even when the payload carries no `custom_fields` key
- the applied-control form passes the folder through a field proxy instead of the whole form store, removing a refetch on every form edit

## Test plan

- 9 API tests: stale empty values (`null`/`""`/`[]`/`false`) from another scope tolerated, real value out of scope rejected, typo'd clear rejected, folder move discards out-of-scope values and enforces the destination's required fields, empty `for_folder` returns global definitions only
- unit test pinning `_is_empty`: `0`/`0.0` stay values (`0 == False` in Python), `null`/`false`/`""`/`[]`/`{}` are empty
- `uv run pytest app_tests/api/test_api_custom_fields.py custom_fields/tests/test_custom_fields.py`: 64 passed
- prettier clean; svelte-check adds no new error classes (CustomFieldsSection: 0 errors; AppliedControlPolicyForm keeps its pre-existing `SuperValidated` Props-typing noise, same pattern as the sibling forms); no live click-through (feature is EE-gated), behavior covered by the API tests

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Enforced required custom fields when moving projects into folders.
  - Preserved provided custom-field values after folder moves.
  - Prevented empty, out-of-scope values from blocking submissions while continuing to reject invalid populated fields.
  - Removed outdated custom-field values when switching folders.
  - Prevented stale field-definition responses from overwriting newer selections.
  - Improved custom-field section expansion based on required fields and meaningful values.
  - Ensured empty folder selections show only global definitions.
  - Added localized error messages when custom fields fail to load.


backend/custom_fields/serializers.py
```diff
@@ -266,6 +297,17 @@ def create(self, validated_data):
...
+        # discards values whose definition no longer applies, so they don't
+        # linger unreadable in forms while still matching cf__ filters.
+        if old_folder_id is not None and instance.folder_id != old_folder_id:
+            allowed = {
+                Folder.get_root_folder_id()
+            } | CustomFieldDefinition._ancestor_or_self_ids(instance.folder)
+            instance.custom_field_values.exclude(
+                definition__folder_id__in=allowed
+            ).delete()
         self._apply(instance)
```
martinzerty, line 312: When the folder is updated, there are no checks on the possible new mandatory custom_fields on the new folder.
In case of a folder modification, the old custom_fields are remove but the new one may be mandatory but not present.

frontend/messages/en.json
```diff
@@ -1080,6 +1080,7 @@
 	"secret": "Secret",
 	"key": "Key",
 	"customFields": "Custom fields",
+	"customFieldsLoadError": "Custom fields could not be loaded. Reselect the domain or reopen the form to retry.",
```
martinzerty, line 1083: May be good to add the translation in the (or at least some of) other languages.

martinzerty (changes requested):

backend/custom_fields/serializers.py
```diff
@@ -266,6 +297,17 @@ def create(self, validated_data):
         return instance
 
     def update(self, instance, validated_data):
+        old_folder_id = getattr(instance, "folder_id", None)
         instance = super().update(instance, validated_data)
+        # Values follow definition scope: moving the object to another folder
+        # discards values whose definition no longer applies, so they don't
+        # linger unreadable in forms while still matching cf__ filters.
+        if old_folder_id is not None and instance.folder_id != old_folder_id:
```
monsieurswag, line 305: # TO FIX:

Use something like this instead (the `Folder.get_iam_folder_id` function exists exactly for this purpose):

```py
old_folder_id = Folder.get_iam_folder_id()
instance = super().update(instance, validated_data)
new_folder_id = Folder.get_iam_folder_id(instance)
```

If we don't use `Folder.get_iam_folder_id` this may break in the future as some models don't have a `folder` field/some model "IAM folder field" may change in the future.

frontend/messages/en.json
monsieurswag, line 1083: # TO FIX:

At a translation for at least french as ~50% of our clients seem to be native french speakers.

backend/custom_fields/serializers.py
```diff
@@ -266,6 +297,17 @@ def create(self, validated_data):
...
+            allowed = {
+                Folder.get_root_folder_id()
+            } | CustomFieldDefinition._ancestor_or_self_ids(instance.folder)
```
monsieurswag, line 308: # NIT:

Replace `allowed = { Folder.get_root_folder_id() } | CustomFieldDefinition._ancestor_or_self_ids(instance.folder)` by `allowed = folder.get_parent_folders(include_self=True)`

backend/custom_fields/views.py
```diff
@@ -41,15 +41,17 @@ def get_queryset(self):
...
             folder_ids = {Folder.get_root_folder_id()}
-            if folder is not None:
-                folder_ids |= CustomFieldDefinition._ancestor_or_self_ids(folder)
+            if for_folder:
+                try:
+                    folder = Folder.objects.filter(pk=for_folder).first()
+                except ValueError, ValidationError:
+                    return queryset.none()
+                if folder is not None:
+                    folder_ids |= CustomFieldDefinition._ancestor_or_self_ids(folder)
```
monsieurswag, line 54: # NIT:

Replace `folder_ids |= CustomFieldDefinition._ancestor_or_self_ids(folder)` by `folder_ids = Folder.get_parent_folders(include_self=True)`

frontend/src/lib/components/Forms/CustomFieldsSection.svelte
```diff
@@ -35,21 +36,70 @@
...
 		}
+		if (seq !== loadSeq) return;
+		// On failure render nothing rather than another folder's fields: the
+		// payload must only ever carry keys of currently rendered definitions.
+		definitions = loaded ?? [];
+		loadFailed = loaded === null;
+		pruneStaleValues(new Set(definitions.map((d) => d.key)));
+		startOpen =
+			definitions.some((d) => d.required) ||
+			Object.values(get(formData)?.custom_fields ?? {}).some(hasValue);
```
monsieurswag, line 102: # NIT:

The `?? {}` fallback seems useless as `CustomFieldsMixin.custom_fields` return type is a `dict`.

frontend/src/lib/components/Forms/CustomFieldsSection.svelte
```diff
@@ -35,21 +36,70 @@
...
+	let startOpen = $state(false);
 
-	// Ensure the nested container exists before any field proxy writes into it.
-	if (form.data && !form.data.custom_fields) form.data.custom_fields = {};
+	const formData = form.form;
+
+	// Drop values whose definition no longer applies (e.g. after a domain change),
+	// otherwise they linger in the payload and the API rejects them as unknown
+	// keys. Payload-only: absent keys never touch values stored server-side.
+	function pruneStaleValues(validKeys: Set<string>) {
```
monsieurswag, line 50: # NIT:

Use a JSDoc-style docstring instead for `pruneStaleValues`, using `//` single-lined comments doesn't display the comments when hovering over the function's name.

frontend/src/lib/components/Forms/CustomFieldsSection.svelte
```diff
@@ -35,21 +36,70 @@
...
+		const current = get(formData)?.custom_fields;
+		if (!current || !Object.keys(current).some((key) => !validKeys.has(key))) return;
+		formData.update(
+			(data: any) => {
```
monsieurswag, line 54: # NIT:

Avoid using `any` in this kind of situation, `unknown` should be the default.
The `any` type completely bypass typescript type permissions so it's often bad to use it.

In this case the type should be `{ custom_fields?: Record<string, unknown> }`.

frontend/src/lib/components/Forms/CustomFieldsSection.svelte
```diff
@@ -35,21 +36,70 @@
...
+				}
+				return data;
+			},
+			{ taint: false }
+		);
+	}
+
+	// `false` excluded: the checkbox binding fabricates it on mount.
+	const hasValue = (v: unknown) =>
+		v != null && v !== false && v !== '' && !(Array.isArray(v) && v.length === 0);
```
monsieurswag, line 68: # NIT:

Name this `isEmpty`/`isEmptyValue` or whatever to be consistent with the `_is_empty` backend naming.

frontend/src/lib/components/Forms/CustomFieldsSection.svelte
```diff
@@ -35,21 +36,70 @@
...
-	// Ensure the nested container exists before any field proxy writes into it.
-	if (form.data && !form.data.custom_fields) form.data.custom_fields = {};
+	const formData = form.form;
+
+	// Drop values whose definition no longer applies (e.g. after a domain change),
+	// otherwise they linger in the payload and the API rejects them as unknown
+	// keys. Payload-only: absent keys never touch values stored server-side.
+	function pruneStaleValues(validKeys: Set<string>) {
+		const current = get(formData)?.custom_fields;
+		if (!current || !Object.keys(current).some((key) => !validKeys.has(key))) return;
```
monsieurswag, line 52: # NIT:

Wouldn't this function's signature be better ?:

```ts
function pruneStaleValues(definitions: Definition[]) {
  const validKeys = new Set(definitions.map((definition) => definition.key));
  // ...
}
```

backend/custom_fields/serializers.py
```diff
@@ -232,6 +254,15 @@ def _clean(self, raw: dict, folder: Folder) -> list:
             raise serializers.ValidationError({"custom_fields": errors})
         return cleaned
 
+    @staticmethod
+    def _is_empty(value) -> bool:
+        # Identity check for False so numeric 0 stays a real value.
```
monsieurswag, line 259: # NIT:

You should just docstring a comment if it describes the function.

I don't get what the comment mean though, is "identify check" a kind of "check" or is "identiy" something that "check".
And how "checking for false" is related to "numeric 0 staying a real value" what even is a real value, a non-empty value i guess, but "non-empty" and "real" aren't obvious opposite linguisycally.

You can still remove the comment anyway, no comment is better than a confusing comment after all.

backend/custom_fields/serializers.py
```diff
@@ -232,6 +254,15 @@ def _clean(self, raw: dict, folder: Folder) -> list:
...
+        return (
+            value is None
+            or value is False
+            or (isinstance(value, (str, list, dict)) and not value)
+        )
```
monsieurswag, line 264: This would be simpler instead of an or-ed chain of conditions + isinstance + not on dict/list/str.

```py
EMPTY_VALUES = ["", [], {}, False, None]

is_empty = value in EMPTY_VALUES
return is_empty
```

frontend/src/lib/components/Forms/CustomFieldsSection.svelte
```diff
@@ -35,21 +36,70 @@
 	const enabled = $derived(page.data?.featureflags?.custom_fields === true);
 
 	let definitions: Definition[] = $state([]);
+	let loadFailed = $state(false);
+	// Expanded by default only when there's something the user must see: a
+	// required field, or values already set (edit mode). Computed once per
+	// definitions load so a manual toggle isn't overridden by form edits.
+	let startOpen = $state(false);
```
monsieurswag, line 43: # FIW:

You can also add JSDoc-style docstrings to local variables like this one.

monsieurswag (dismissed):
Apart from these the PR seems to work as expected.

backend/custom_fields/serializers.py
nas-tabchiche: `["", [], {}, False, None]` would not work here, as `0 == False`

martinzerty (dismissed):
LGTM when the CI passes

ab-smith (approved):

PR #4743, opened 2026-09-03
fix(frontend): show loading state for tables and CVE pulls

Claquetteuuuh:
## What & why

Two loading states were missing on list views, both of which made the app look like it had no data when it was actually working.

**Pulling a CVE catalog gave no feedback at all.** The KEV/EUVD pull runs synchronously in the backend request (external fetch + bulk upsert) and can take a while, but nothing on screen changed — and the buttons stayed enabled, so clicking again started a second concurrent import, each popping its own success toast. Both buttons are now disabled for the duration of a pull (spinner on the running one, plus a guard in the handler), with an activity bar above the table.

**A slow first table load was indistinguishable from an empty table.** On a mediocre connection you could land on a list view, see an empty table, and conclude there was no data while rows were still on the way.

`ModelTable` now renders placeholder rows whenever it has nothing to show and data is on the way — during a pull, via the new opt-in `loading` prop, and on its own first fetch, tracked around its data handler. They are real `<tr>`/`<td>` so they line up with the columns the incoming rows will fill.

The two near-identical 30-line inline click handlers are factored into `runSync`/`confirmSync`, which is what makes a shared in-flight state possible (net −55 lines there). The `loading` prop defaults to `false`, so the ~50 other views that render `ModelTable` are unaffected. No new UI strings: `loading`, `syncKev`, `syncEuvd`, `pullCatalog` and the failure keys already existed.

Closes CA-1845

## Test plan

- Two Playwright tests added to `security-advisories.test.ts`:
  - `CVE sync locks both buttons and shows loading feedback while pulling` — stubs `/security-advisories/sync-kev` (so it never reaches CISA) and holds the response open: asserts both buttons disabled, activity bar and placeholder rows visible, that spamming either button starts no second request, and exactly one toast once it settles.
  - `table shows placeholder rows while its first page is loading` — holds the paginated data request: asserts placeholder rows appear and then disappear, and that the sync bar stays hidden (this is not a pull).
- Verified the first test genuinely catches the bug: with the fix reverted it fails on `expect(sync-kev-button).toBeDisabled()` — *Received: enabled*. The second test was verified by hand rather than against pre-fix code (its `data-testid` did not exist before).
- Checked in the dev server with the requests held open: empty table → bar + placeholder rows; populated table (3397 rows) → bar only, rows kept on screen; a single request in both cases. Slow first load simulated with a 6s delay on the data request → 5 placeholder rows during the fetch, real rows after.
- `./tests/e2e-tests.sh detailed/security-advisories.test.ts` 6 passed. Also ran with `detailed/compliance-assessments.test.ts` (heavy `ModelTable` user) to check the shared component: 6 passed + 1 pre-existing flaky (`sync MITRE CWE catalog`, hits MITRE live, passes on retry).
- `pnpm run test` 181 passed; backend `uv run pytest` 4242 passed (nothing touched in `backend/`).

## Checklist

- [x] CLA accepted (first contribution only) — n/a if already signed

- [x] Svelte 5 syntax; `pnpm run format` run (`prettier --check` clean). `pnpm run lint` (eslint) could not be run — pre-existing environment issue in this checkout (`@eslint/eslintrc` missing from `node_modules`), unrelated to this change
- [x] No new/changed UI strings — reused existing message keys
- [x] Clicked through the change in the dev server (see test plan)

- [x] Tests added or updated and passing locally — `./tests/e2e-tests.sh detailed/security-advisories.test.ts` (6 passed), `pnpm run test` (181 passed), backend `uv run pytest` (4242 passed)
- [x] Regression test added for bug fixes — verified red without the fix, green with it
- [ ] No tests or docs needed for this change — n/a, tests added above

## Summary by CodeRabbit

* **New Features**
  * Added skeleton rows and loading indicators while security advisory data loads or synchronizes.
  * Sync controls now show model-specific progress and disable all sync buttons while any sync is active.
  * Added completion feedback when synchronization finishes.
  * Tables now refresh and display updated rows after a successful synchronization.

* **Bug Fixes**
  * Prevented repeated clicks from triggering duplicate sync requests.
  * Improved loading states so tables no longer appear empty while data is being fetched.


Claquetteuuuh:

nas-tabchiche (approved):
LGTM

PR #4749, opened 2026-09-04
fix: task export as excel

ab-smith:
- [x] fix Excel report
- [ ] better way to access previous tasks
- [x] improve importer to cover daily granularity
- [ ] be able to create expected evidence on the fly
- [ ] importer to autocreate/link evidences if set
- [x] analytics page for all tasks or filtered (from table)
- [ ] include evidences importer
- [ ] missing strings on data wizard import options
- [ ] visual inconsistency / glitches for markdown fields

## Bug Fixes

- Improved Excel exports so filtering labels are retrieved correctly from task templates.
- Exports now show a blank value when no filtering labels are available, rather than an empty separator.
- Improved export performance by preloading filtering label data.

### DIFF
--- a/backend/core/views.py
+++ b/backend/core/views.py
@@ -17459,10 +17459,12 @@ class TaskTemplateViewSet(CommitmentActionsMixin, ExportMixin, BaseModelViewSet)
                 ),
             },
             "labels": {
-                "source": "filtering_labels",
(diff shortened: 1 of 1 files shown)


martinzerty (approved):
LGTM

nas-tabchiche (approved):
LGTM

PR #4751, opened 2026-09-04
fix: delete a folder having an elementaryaction in killchain yield a 500

Spaghetto784:
## What & why

Deleting a Folder that (via cascade) contains an ElementaryAction still referenced by a KillChain (`on_delete=PROTECT`) raised an uncaught `ProtectedError`, surfacing as a raw 500 instead of a user-friendly error. `FolderViewSet.destroy` now catches `ProtectedError` and returns a structured 409 with a clear message.

Closes #CA-1797

## Test plan

- Added `backend/core/tests/test_folder_destroy_protected.py`:
  - deleting a folder that cascades into a KillChain-protected ElementaryAction → 409 with a `detail` message, folder not deleted
  - deleting an empty folder still succeeds normally → 204 (no regression)
- Verified the fix is load-bearing: temporarily reverted it and confirmed the original `ProtectedError`/500 reproduces exactly as in the bug report, then restored it
- `uv run pytest backend/core/tests/test_folder_destroy_protected.py` → 2 passed- Added `backend/core/tests/test_folder_destroy_protected.py`:
  - deleting a folder that cascades into a KillChain-protected ElementaryAction → 409 with a `detail` message, folder not deleted
  - deleting an empty folder still succeeds normally → 204 (no regression)
- Verified the fix is load-bearing: temporarily reverted it and confirmed the original `ProtectedError`/500 reproduces exactly as in the bug report, then restored it
- `uv run pytest backend/core/tests/test_folder_destroy_protected.py` → 2 passed

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Deleting items with protected references now returns a clear HTTP 409 response.
  - Error details identify visible items preventing deletion and indicate when additional blockers are hidden.
  - Deleting empty folders continues to complete successfully with a 204 response.
  - Improved consistency for protected deletion errors involving domains, elementary actions, and entities.
- **Tests**
  - Added regression coverage for protected-reference deletion failures and successful empty-folder deletion.


backend/core/views.py
```diff
@@ -8308,6 +8309,21 @@ def perform_create(self, serializer):
...
         Folder.create_default_ug_and_ra(folder)
 
+    def destroy(self, request, *args, **kwargs):
+        try:
+            return super().destroy(request, *args, **kwargs)
+        except ProtectedError:
+            return Response(
+                {
+                    "detail": (
+                        "Cannot delete this domain: it still contains elements "
```
martinzerty: The message could be way more specific to the object itself. Now, the message only takes as example the elementary action but it may be something else.

in `ebios_rm/views.py` :
```python
"detail": (
        f"Cannot delete elementary action '{instance.name}' — it is "
        f"used in {len(operating_modes)} operating mode kill chain(s)"
        + (f": {names}" if names else "")
        + ". Remove it from those kill chains first."
    )
```

backend/core/views.py
```diff
@@ -8308,6 +8309,21 @@ def perform_create(self, serializer):
         folder = serializer.save()
         Folder.create_default_ug_and_ra(folder)
 
+    def destroy(self, request, *args, **kwargs):
+        try:
```
martinzerty: It may be considered to move this try/except block to BaseModelViewSet so that each child class inherit.

martinzerty (changes requested):

backend/core/views.py
Spaghetto784: Good point. The generic message now uses Django's own `error.protected_objects` to list the actual blocking objects instead of hardcoding the elementary-action case. Since a Folder can be blocked by many different kinds of protected relations, this stays accurate whatever the actual blocker is.

backend/core/views.py
Spaghetto784: Done on the new commit that is coming after ! moved the try/exception into `BaseModelViewSet.destroy`, with a `get_protected_error_response_data(instance,error)` hook that subclasses can override for a more specific message. `FolderViewSet.destroy` was removed entirely since it now inherits the generic behavior

martinzerty (approved):
LGTM

PR #4770, opened 2026-09-07
fix: put autocompleteselect dropdown on top of the table header

martinzerty:
## What & why

The header of the table of the control plan was over the dropdown of the autocompleteselect of the selection of applied controls in the filter selection section.

### After
Screenshot_2026-09-07_16-50-42

Closes #

## Test plan

Create tasks. Go to the the control plan's page. See the drop-down options clearly appearing over the table's header, not under it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the applied-controls filter dropdown on the tasks review page so it displays correctly within the page layout.


nas-tabchiche (approved):
LGTM. One off-scope remark though: `portalDropdown` is a badly named prop.

PR #4785, opened 2026-09-09
fix(backend/frontend): hash not shown in detailed view of evidences

Spaghetto784:
## What & why

The `attachment_hash` field on Evidence Revisions showed up two ways in the detail view:
- the label rendered as the raw `attachment_hash` key instead of a translated label (no i18n message key existed for it)
- the value showed `--` for revisions created before the field was introduced (migration `0124` added the column but never backfilled existing rows)

This adds the missing `attachmentHash` translation key (en/fr) and a data migration that computes the SHA256 hash for existing attachments that don't have one yet, using the same chunked-hashing logic as `EvidenceRevision.save()`.

Closes CA-1849

## Test plan

- Created an `EvidenceRevision` with a real attachment via the Django shell, confirmed the hash is computed and the detail view now shows "Attachment hash" / "Empreinte de la pièce jointe" with the correct value (verified manually in the dev server, both community and enterprise editions — enterprise reuses the same `core` app/migrations).
- `python manage.py makemigrations --check --dry-run core` → no schema drift.
- `pytest core/tests/test_models.py::TestEvidence` → passes.

## Checklist

- [x] `pnpm run lint` and `pnpm run format` run (prettier check passes on changed files)
image

## Tests & documentation — definition of done

- [ ] No tests or docs needed for this change — why: fixes a translation key and backfills data via a Django data migration; existing model tests (`TestEvidence`) already cover `attachment_hash` computation and still pass. No migration-test convention exists elsewhere in the repo to follow for the backfill itself.

## Summary by CodeRabbit

- **Localization**
  - Added French and English labels for attachment fingerprints.
  - Added English messages for date-format settings, confirmations, errors, and unexpected server responses.

- **User Experience**
  - Updated the X-rays empty state.
  - Added an error message for failed X-rays loading.

- **Data Improvements**
  - Existing attachment records can now display computed fingerprint information when available.

### DIFF
--- a/backend/core/migrations/0187_backfill_evidencerevision_attachment_hash.py
+++ b/backend/core/migrations/0187_backfill_evidencerevision_attachment_hash.py
@@ -10,9 +10,11 @@ logger = logging.getLogger(__name__)
 def backfill_attachment_hash(apps, schema_editor):
     EvidenceRevision = apps.get_model("core", "EvidenceRevision")
 
-    revisions = EvidenceRevision.objects.exclude(attachment="").exclude(
-        attachment__isnull=True
-    ).filter(attachment_hash__isnull=True)
+    revisions = (
+        EvidenceRevision.objects.exclude(attachment="")
+        .exclude(attachment__isnull=True)
+        .filter(attachment_hash__isnull=True)
+    )
 
     for revision in revisions:
         if not default_storage.exists(revision.attachment.name):
--- a/backend/core/migrations/0187_backfill_evidencerevision_attachment_hash.py
+++ b/backend/core/migrations/0187_backfill_evidencerevision_attachment_hash.py
@@ -1,0 +1,42 @@
+import hashlib
+import logging
+
+from django.core.files.storage import default_storage
+from django.db import migrations
+
+logger = logging.getLogger(__name__)
+
+
+def backfill_attachment_hash(apps, schema_editor):
+    EvidenceRevision = apps.get_model("core", "EvidenceRevision")
+
+    revisions = EvidenceRevision.objects.exclude(attachment="").exclude(
+        attachment__isnull=True
+    ).filter(attachment_hash__isnull=True)
+
+    for revision in revisions:
+        if not default_storage.exists(revision.attachment.name):
+            continue
+        try:
+            hash_obj = hashlib.sha256()
+            with default_storage.open(revision.attachment.name, "rb") as f:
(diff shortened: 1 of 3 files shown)


backend/core/migrations/0187_backfill_evidencerevision_attachment_hash.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+
+    for revision in revisions:
+        if not default_storage.exists(revision.attachment.name):
+            continue
+        try:
+            hash_obj = hashlib.sha256()
+            with default_storage.open(revision.attachment.name, "rb") as f:
+                for chunk in iter(lambda: f.read(1024 * 1024), b""):
+                    hash_obj.update(chunk)
```
monsieurswag, line 26: # FYW:

You can do this instead of using `iter`:

```py
while (chunk := f.read(1024 * 1024) != b"":
    hash_obj.update(chunk)
```

monsieurswag (dismissed):

backend/core/migrations/0187_backfill_evidencerevision_attachment_hash.py
Spaghetto784: Okay I see thank you for the review I'm gonna stay with 'iter' object for now cause we used it in `EvidenceRevision.save()` already but I keep that in mind for the next times 😄 

btw it should be : 

```python
while (chunk := f.read(1024 * 1024)) != b"":
    hash_obj.update(chunk)
```
    
 Cause the snippet you gave me will fail with a 'TypeError' if I'm not wrong 🤔

backend/core/migrations/0189_backfill_evidencerevision_attachment_hash.py
```diff
@@ -0,0 +1,44 @@
...
+            hash_obj = hashlib.sha256()
+            with default_storage.open(revision.attachment.name, "rb") as f:
+                for chunk in iter(lambda: f.read(1024 * 1024), b""):
+                    hash_obj.update(chunk)
+            revision.attachment_hash = hash_obj.hexdigest()
+            revision.save(update_fields=["attachment_hash"])
+        except Exception as e:
+            logger.warning(
+                "Failed to backfill attachment hash",
+                revision_id=revision.pk,
```
nas-tabchiche, line 32: `structlog` args but the logger is `logging.get_logger`. Import structlog instead of logging

nas-tabchiche (changes requested):

PR #4795, opened 2026-09-11
fix: improve documents validation flows

### LINKED ISSUE #4777: Frontend logs two lines per 404
**Describe the bug**

The frontend writes two lines to stdout for every HTTP 404: one blank, one `[404] GET /path`. During a weekly vulnerability scan against our instance this produced 42,483 frontend log lines in 22 minutes, roughly half of them empty.

The emitter is SvelteKit's own runtime (`.svelte-kit/output/server/chunks/utils.js`)

```js
function format_server_error(status, error, event) {
const formatted_text = `
\x1B[1;31m[${status}] ${event.request.method} ${event.url.pathname}\x1B[0m`;
if (status === 404) {
    return formatted_text;
}
return `${formatted_text}
${error.stack}`;
}
```

The template literal begins with a newline, which is where the blank line comes from.

It fires because SvelteKit routes unmatched paths through the `handleError` hook and its default implementation logs them. `frontend/src/hooks.server.ts` exports `handle` and `handleFetch` but no `handleError`, so the default applies. This is still the case on `main`.

**To Reproduce**

1. Start the standard Docker Compose stack
2. Request any path that is not a route: `curl -k https://<host>/index.php`
3. Run `docker logs ciso-frontend`
4. See two lines per request, the first one empty:

```
Sep 08 04:32:45 isms ciso-frontend[1026]:
Sep 08 04:32:45 isms ciso-frontend[1026]: [404] GET /dana-na/auth/
```

**Expected behavior**

A 404 on a path that was never a route is not an application error and should not be logged by default and certainly not as two lines. Exporting a `handleError` that skips 404s would remove it, e.g.

```ts
import type { HandleServerError } from '@sveltejs/kit';

export const handleError: HandleServerError = ({ status, error }) => {
if (status !== 404) console.error(error);
return { message: 'Internal Server Error' };
};
```

If you would rather keep the logs for people who want them, gating this behind an env var would work equally well for us.

**Screenshots**

Not applicable

- Device: server (self-hosted Docker Compose, nginx in front, container output to journald)
- OS: Debian 13, Docker
- Browser: not applicable
- Version: CISO Assistant Community v3.21.2

**Additional context**

Our stack ships journald into a SIEM, so this volume has a real cost downstream. The reverse proxy in front already logs the same requests, and with the actual client IP, which the frontend line cannot show.

For completeness: the same scan also produced around 90,000 backend lines from `django_structlog`'s `request_started` / `request_finished` pair, which we resolved on our side with `LOG_LEVEL=WARNING`. Only 124 of the ~40,000 scanned requests touched `/api/` directly — the rest of that backend volume was SSR fan-out from rendering error pages. So the frontend `handleError` is the remaining piece.

I noticed #3028 and #2618 took a similar direction on the backend side, so this may align with what you already want.

### LINKED ISSUE #4779: Uploaded documents reviews can't be approved
**Describe the bug**
It's impossible to review a document that has been imported in ciso.

**To Reproduce**
Go to Documents and create a new one (for example a procedure).
Submit a file, ask for review, approve it and publish it.
Upload a new file for the same document (v2), ask for review.
You can't approve it anywhere, so you can't update the file.

**Expected behavior**
Being able to approve and publish a v2

**Screenshots**

Image

- Last version of Ciso
- Last version of Chrome

ab-smith:
Fix validation cases with documents:
- approved by is shown
- be able to approve non-authored documents
- disable approval if not authorised to do so (avoid a silent reject)
- fix duplicate errors on frontend

## Summary by CodeRabbit

- **New Features**
  - Approved revisions now display the reviewer’s name in document history and PDF metadata.
  - Lifecycle and deletion controls respect user permissions, with disabled states and explanatory tooltips.
  - Document views select the appropriate active revision for lifecycle actions.

- **Bug Fixes**
  - Delete failures now show useful backend-provided details.
  - Server errors are logged consistently, while 404 requests no longer produce unnecessary error logs.

- **Localization**
  - Added English and French translations for “Approved by.”

### DIFF
--- a/backend/doc_management/models.py
+++ b/backend/doc_management/models.py
@@ -153,6 +153,12 @@ class DocumentRevision(AbstractBaseModel, FolderMixin):
         PUBLISHED = "published", _("Published")
         DEPRECATED = "deprecated", _("Deprecated")
 
+    # `reviewer` is also written by mark_change_requested(), so it only means
+    # "approver" once the revision reached one of these.
+    APPROVED_STATUSES = frozenset(
+        {Status.VALIDATED, Status.PUBLISHED, Status.DEPRECATED}
+    )
+
     document = models.ForeignKey(
         ManagedDocument, on_delete=models.CASCADE, related_name="revisions"
     )
--- a/backend/doc_management/templates/doc_management/policy_document_pdf.html
+++ b/backend/doc_management/templates/doc_management/policy_document_pdf.html
@@ -240,6 +240,12 @@ body {
                 <td>{{ author_name }}</td>
             </tr>
             {% endif %}
+            {% if reviewer_name %}
+            <tr>
+                <td>Approved by:</td>
+                <td>{{ reviewer_name }}</td>
+            </tr>
+            {% endif %}
             {% if published_at %}
             <tr>
                 <td>Published:</td>
--- a/backend/doc_management/tests.py
+++ b/backend/doc_management/tests.py
@@ -4,10 +4,10 @@ import pytest
 
 from core.models import Policy
 from core.net_safety import BlockedRequestError
-from doc_management.models import DocumentContainer
+from doc_management.models import DocumentContainer, DocumentRevision
 from doc_management.serializers import ManagedDocumentWriteSerializer
-from doc_management.views import _safe_url_fetcher
-from iam.models import Folder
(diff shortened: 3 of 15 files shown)


Mohamed-Hacene (approved):
0k

PR #4817, opened 2026-09-14
feat(documents): surface pending revisions alongside the one in force

Claquetteuuuh:
## What & why

Both the container list and the document header reported a single revision: the one in force, which stays on the published one until a publish repoints `current_revision`. A v2 still in the approval loop was therefore invisible from the list, and the container showed 100% progress while a review was pending.

Expose `pending_revision` on the container serializer and render it as a secondary badge, and show the in-force revision next to the working one in the uploaded and linked document headers. `status` and `progress` keep describing the revision in force: on a compliance document, which version applies must not be overwritten by work in progress.

Statuses in those headers now go through `safeTranslate` — they were printed raw while the version history right below them was already translated. Also fills in the missing French `validated` status label.

Closes # 

## Test plan

Scenario reproduced locally: a container with a published v1 and a v2 left in the approval loop.

| Screen | Before | After |
|---|---|---|
| `/document-containers` | `Published · 100%`, no trace of the v2 | `Published` + `v2 · In review` badge, progress unchanged |
| Document header (uploaded / linked) | single raw pill `in_review · v2` | `Published · v1` + `In review · v2`, both translated |

Also verified the badge and the second pill **disappear** once the v2 is published (the two revisions converge), and that the full lifecycle still runs end to end: In review → Approve → Validated → Publish → Published.

Query cost measured before/after: the new field was one extra query per container row, now constant and covered by a regression test.

- `uv run pytest` — **4872 passed**, 20 skipped, 1 xfailed
- `pnpm run test` — **210 passed**
- Clicked through on the dev server (list + both header variants, EN)

Screenshots: list badge and the two header pills attached below.

## Checklist

- [x] One focused change, branched off `main` — single commit, 0 behind `origin/main`

- [x] Migrations generated with `makemigrations` and committed (or none needed) — `makemigrations --check` reports no changes; `ACTIVE_STATUSES` is a class constant, not a field

- [x] Svelte 5 syntax; `pnpm run lint` and `pnpm run format` run — both run; `lint` fails repo-wide on 3588 pre-existing problems, **0 on the lines this PR adds**
- [x] New/changed UI strings added to `frontend/messages/en.json` — no new key needed, `validated` already existed in `en.json`; added the missing French one

- [x] Tests added or updated and passing locally — backend `uv run pytest`, frontend `pnpm run test`. e2e **not** run: there is no functional test suite for documents, and building one means driving the whole container + revision flow through the UI
- [x] Regression test added for bug fixes — 5 backend tests (including one locking the `status`/`progress` decision and one guarding the query count) and 4 frontend unit tests
- [x] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added — added a paragraph to `concepts/documents.md` § Lifecycle and versioning; no new page, so `SUMMARY.md` is unchanged
- [ ] No tests or docs needed for this change — tests and docs were needed and added

## Summary by CodeRabbit

- **New Features**
  - Document lists now show pending revision versions and statuses.
  - Document views identify the currently published (“in force”) revision when a newer revision is under review.
  - Document headers show published and in-progress revisions when they differ.
  - Published revision badges include the relevant version number.
  - Pending revision indicators preserve existing document status and progress information.

- **Documentation**
  - Clarified document versioning and publication behavior, including how pending revisions are represented.


frontend/messages/fr.json
```diff
@@ -51,6 +51,7 @@
 	"authorHint": "Partez d’une page vierge ou d’un modèle dans l’éditeur.",
 	"newDocumentSubtitle": "Rédigez-le, téléversez un fichier ou liez une source externe.",
 	"deleteConfirm": "Voulez-vous vraiment supprimer ceci ?",
+	"validated": "Validée",
```
monsieurswag, line 54: # TO FIX:

You MUST always add translations for AT LEAST english and french.
So you MUST at least add an english translation for the `"validated"` translation key.

At frontend/src/lib/components/DocumentEditor/UploadedDocumentView.svelte:245

frontend/src/lib/components/DocumentEditor/LinkedDocumentView.svelte
```diff
@@ -223,9 +228,16 @@
 					{parent?.name || m.untitled()}
 				</h1>
 				{#if currentRevision}
-					<span class="badge {statusStyles[status ?? 'draft']} mt-1 text-xs">
-						{status} · v{currentRevision.version_number}
-					</span>
+					<div class="mt-1 flex flex-wrap items-center gap-1.5">
+						{#if showsInForceRevision}
+							<span class="badge {statusStyles.published} text-xs">
+								{safeTranslate('published')} · v{inForceRevision.version_number}
```
monsieurswag, line 234: # NIT:

Replace `safeTranslate('published')` by `m.published()`.
The `safeTranslate` function MUST only be used for dynamic translations (a non-static(non-constant) string).

frontend/src/lib/components/ModelTable/ModelTable.svelte
```diff
@@ -1188,6 +1188,18 @@
 														{:else}
 															--
 														{/if}
+													{:else if URLModel == 'document-containers' && key === 'status' && row.meta?.pending_revision}
+														<div class="flex items-center space-x-2">
+															<span>{safeTranslate(value ?? '-')}</span>
```
monsieurswag, line 1193: # NIT:

Remove `?? '-'`, it's useless as the `'-'` string can't be translated, and even so, if `value` is `undefined` it will be transalted to `"--"` (as you can see by looking for the `"--"` translation key in `frontend/messages/en.jon`).

backend/doc_management/models.py
```diff
@@ -155,6 +155,11 @@ class Status(models.TextChoices):
         {Status.VALIDATED, Status.PUBLISHED, Status.DEPRECATED}
     )
 
+    # Still moving through the approval loop: neither in force nor retired.
```
monsieurswag, line 158: # NIT:

Remove bad claude comment.

frontend/src/lib/utils/documentRevisions.ts
```diff
@@ -16,3 +16,10 @@ export function pickWorkingRevision<T extends RevisionLike>(revisions: T[]): T |
 		null
 	);
 }
+
+// The revision that currently applies. It stays behind the working revision for
+// the whole approval loop, so a header showing only one of the two hides either
+// what is in force or what is coming.
```
monsieurswag, line 22: # NIT:

We need avoid letting our codebase be filled with useless cryptic comments.
Either delete these comments, or write a real JSDoc docstring.

For example:

```ts
/** Return the first `"published"` revision, or return `null` if none were found.
 * 
 * The "revision in force" is the most recent revision (highest `revision.version_number`) whose `status` is `"published"`.
 */
```

frontend/src/lib/utils/documentRevisions.ts
```diff
@@ -16,3 +16,10 @@ export function pickWorkingRevision<T extends RevisionLike>(revisions: T[]): T |
...
+export function pickInForceRevision<T extends RevisionLike>(revisions: T[]): T | null {
+	return revisions.find((r) => r.status === 'published') ?? null;
+}
```
monsieurswag, line 25: # FIW:

Note that this test case (which make sense) would fail:

```ts
it('returns the LAST published revision (as only the LAST one is the one "in force")', () => {
    expect(pickInForceRevision([rev(2, 'in_review'), rev(2, 'published'), rev(3, 'published')])?.id).toBe('v3');
});
```

I don't consider it a bug though as there should apparently be only ONE `"published"` revision at a time for a document.
But maybe you should make `pickInForceRevision` more defensive to accomodate to this test case.

monsieurswag, line 25: # FIW:

The typing would be way cleaner with something like this:

```ts
interface Revision {
	status: "draft" | "change_requested" | "in_review" | "validated" | "published" | "deprecated"
	[key: string]: unknown
}

export const ACTIVE_REVISION_STATUSES: Revision["status"][] = ['draft', 'change_requested', 'in_review', 'validated'];
export const APPROVED_REVISION_STATUSES: Revision["status"][] = ['validated', 'published', 'deprecated'];

export function pickInForceRevision(revisions: Revision[]): Revision | null {
	return revisions.find((r) => r.status === 'published') ?? null;
}
```

Obviously your current code only adapts to what was already in the codebase.
But don't forget you can also use your PR to make the codebase better.
So when you some code burn your eyes, you have the right to improve it, even if it wasn't the original intent of your PR.

monsieurswag (changes requested):
The PR work as expected, it can be merged once these little things are fixed.

frontend/messages/fr.json
Claquetteuuuh: The English key already exists, en.json:137 "validated": "Validated". French was the only one of the six workflow statuses missing it, which is why the diff only adds that one.
Checked the other five too: draft, inReview, changeRequested, published and deprecated are present in both locales.

monsieurswag (dismissed):

PR #4821, opened 2026-09-15
feat(tasks): add a findings tab to the task detail view

Claquetteuuuh:
## What & why

  A task already carries a `findings` many-to-many, but the detail view had no tab for it: the findings raised from a periodic task were only reachable through the edit form. This adds **Findings** to the tab rail, next to Documents and Incidents, so a completed occurrence's findings are one click away.

  The tab is read-only, mirroring the findings tab on applied controls - a finding belongs to its assessment, so it is still linked from the task form rather than created from here.

  One line of it is not cosmetic. The tab lists through `/findings/?task_templates=<id>`, and that filter was missing from `FindingFilterSet`. An unknown filter is **dropped silently** rather than rejected, so the tab would have rendered fine while listing every finding in the instance instead of the task's own. Both tests cover that failure mode.

  Closes # 

  ## Test plan

  Both tests were checked against a reverted fix: with the filter line removed they
  fail for two distinct reasons, which confirms they actually pin the behaviour
  rather than passing by accident. With the fix in place the whole suite is green.

  - `test_findings_filtered_by_task_template` — without the filter, the task's tab
    answers with every finding in the database instead of its own
  - `test_task_without_findings_lists_none` — without the filter, a task with no
    findings answers with all of them instead of none

  Clicked through on the dev server, on a task linked to one finding out of two in
  the database:

  | Checked | Result |
  |---|---|
  | Tab rail | **Findings** appears with its count badge (`1`) |
  | Tab contents | only the linked finding — the unrelated one is absent |
  | Same finding linked to two tasks | listed once, no duplicate rows from the m2m join |
  | `follow_up` feature flag turned off | the tab disappears, like Incidents under its own flag |
  | Global `/findings` page | unchanged — its filter panel is declared separately in `table.ts` |

  - `uv run pytest` — **4874 passed**, 20 skipped, 1 xfailed (baseline 4872 + the 2 added here)
  - `pnpm run test` — **210 passed**

  Screenshot of the tab attached below.
  task-findings-tab

  ## Checklist

- [ ] CLA accepted (first contribution only) - not a first contribution

- [x] Migrations generated with `makemigrations` and committed (or none needed) — `makemigrations --check` reports no changes; this touches a filterset, not a model

- [x] Svelte 5 syntax; `pnpm run lint` and `pnpm run format` run — the frontend change is declarative config in `crud.ts`, no component; eslint reports 8 pre-existing errors in that file, **0 on the lines added here**
- [x] New/changed UI strings added to `frontend/messages/en.json` — none needed, the tab label reuses the existing `findings` model name

- [x] Tests added or updated and passing locally — backend `uv run pytest`, frontend `pnpm run test`. e2e not run: no functional suite covers the task detail tabs
- [x] Regression test added for bug fixes — two tests pin the filter, including the empty case a dropped filter turns into "every finding in the instance"
- [x] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added — added a sentence to `concepts/tasks.md` documenting the tab and the fact that findings attach to the definition, not the occurrence, so a recurring task pools the findings from all its runs; no new page, so `SUMMARY.md` is unchanged
- [ ] No tests or docs needed for this change — tests and docs were needed and added

## Summary by CodeRabbit

- **New Features**
  - Added exact-match filtering for task templates.
  - Task templates now display their related findings for easier navigation and review. Creation and deletion of findings from the task-template view remain unavailable.


product-docs/concepts/tasks.md
```diff
@@ -17,7 +17,7 @@ graph LR
...
+The task template is the definition — assignee, recurrence rule, expected evidence — and the task occurrence is the actual unit of work scheduled from it. The schedule is a JSON field describing the cadence (DAILY / WEEKLY / MONTHLY / YEARLY with the usual iCalendar refinements), rendered in words wherever it's displayed — *Monthly*, *Every 3 months* — rather than as the raw rule. Templates can be wired to many other objects — applied controls being the canonical one (the "maintains" semantics: this work keeps that control healthy) — and when an occurrence is completed, the evidence revision it produces is back-linked through `task_node` so the audit trail closes the loop. Observations that come out of a run are recorded as [findings](findings-assessments.md) and listed on the definition's **Findings** tab; they attach to the definition rather than to the occurrence, so a recurring task pools the findings raised across all of its runs.
```
monsieurswag, line 20: # TO FIX:

Remove all the changes you did to `product-docs/concepts/tasks.md`.

The documentation under `/product-docs` is our actual user-facing documentation.
You can find this documentation here: https://intuitem.gitbook.io/ciso-assistant/concepts/operations/tasks#mental-model

The text is already a big blob of AI slop, so please don't make it even worse.
You don't have to add documentation for such a simple change don't worry about it.

backend/core/tests/test_finding_task_template_filter.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.mark.django_db
+def test_task_without_findings_lists_none(admin_client, findings):  # noqa: F811
+    """The empty case is the one a dropped filter hides: it answers with every
+    finding in the instance instead of none."""
+    response = admin_client.get(
+        reverse("findings-list"), {"task_templates": str(findings["other_task"].id)}
+    )
+    assert response.json()["results"] == []
```
monsieurswag, line 60: # TO FIX:

Delete the whole test, you don't have to write a test for such a simple change.

Imagine if we had to add a single dedicated test for each field in a `SomeFilterSet.Meta.fields` list.
If so we would have around 150 different test files just to test a SINGLE behavior (the "exact" filter type filtering).

monsieurswag (changes requested):
The PR works as expected.
The filter added to `FindingFilterSet` doesn't have any practical use yet but that's okay.

Just remove the doc change you made + the test.

monsieurswag (approved):

PR #4824, opened 2026-09-15
fix: drop on dispatch from releases (protected tags only)

ab-smith:
## Summary by CodeRabbit

- **Chores**
  - Container image workflows now run only when version tags are pushed.
  - Manual workflow triggering is no longer available.
  - The `latest` image tag is now applied on every container image workflow run, including non-release builds.


nas-tabchiche (approved):
LGTM

eric-intuitem (approved):
LGTM

PR #4826, opened 2026-09-15
feat(applied-controls): add create and select-existing buttons on the linked-object tabs

### LINKED ISSUE #4696: Missing buttons
**Describe the bug**
I'm not able to link an asset to an Applied control.

**To Reproduce**
1. Go to 'Mesures appliquées' and select one
3. Scroll down to linked objects
4. See:
4.1. We are able to add a document by clicking on 'New document'
4.2. We are able to select or create a new evidence using the dedicated buttons
4.3. We arn't able to create or select a Finding
4.4. We arn't able to create or select an Asset
4.5. We arn't able to create or select an Incident

**Expected behavior**
Have a button to select the objects (like the Documents part).

Claquetteuuuh:
## What & why

The linked-object tabs of an applied control were inconsistent. Evidences and Tasks offered both a *create* and a *select existing* button; Documents could only create; Findings, Assets and Incidents offered nothing at all. A user looking at those three tabs concludes the link is impossible, which is exactly what the reporter did.

The link was never impossible, only hard to reach: assets from the applied control's own edit form inside a collapsed "Relationships" section, findings and incidents only from the finding or the incident itself inside a collapsed "More" section. The row checkboxes do not help — they drive bulk edits on objects already listed and cannot bring in one that is not.

This makes all four tabs consistent:

* **Findings, Assets, Incidents** — both buttons, configuration only. `AppliedControlWriteSerializer` already accepted `findings` and `incidents`, and `assets` is a many-to-many owned by the applied control. Creating from a tab pre-links the parent automatically.
* **Documents** — the *select existing* button, which needed a backend change. The relation is declared on the document side (`DocumentContainer.applied_controls`, `related_name="control_documents"`), and a `ModelSerializer` never exposes a reverse accessor on its own. Without an explicit field the tab failed **silently**: DRF drops an unknown key, so the PATCH answered `200` while creating no link at all. `control_documents` is now declared on the serializer, mirroring `findings`, `incidents` and `task_templates`.

This is also the first coverage of the `addExisting` mechanism. It renders through the create button's slot, so a tab that disables creation silently drops it — two tabs on `posture-assessments` are configured that way today and show no button.

Closes #4696

## Test plan

Both test suites were red-checked against a reverted fix, so they pin the behaviour rather than passing by accident.

**Backend** — without the serializer field, the two tests fail for distinct reasons:

assert [] == [Backup policy]   nothing is written
assert 1 == 2                  a pre-existing link survives, the new one is lost

The second case is the one a naive test misses: seed a link beforehand and a silently dropped field still looks like it worked.

**Frontend** — without the configuration:

Error: element(s) not found
- waiting for getByRole('tabpanel', { name: /^Findings/ })
    .getByTestId('select-existing-button')

Walked through each tab manually and checked the result in the database, not just on screen:

| Tab | Action | Result |
|---|---|---|
| Findings | attach an existing finding | 5 → 6 linked |
| Assets | attach an existing asset | appears in the table immediately |
| Incidents | create from the tab | 3 → 4, created object arrives already linked |
| Documents | attach an existing document | 0 → 1 linked |

- `uv run pytest` — **4874 passed**, 20 skipped, 1 xfailed (baseline 4872 + the 2 added here)
- `pnpm run test` — **206 passed**
- `./tests/e2e-tests.sh detailed/applied-control-linked-objects.test.ts` — **1 passed**, covering all four tabs
- `python manage.py check` — no circular import from the new cross-app reference

Screenshot of the Findings tab with both buttons attached below.

## Checklist

- [x] Migrations generated with `makemigrations` and committed (or none needed) — `makemigrations --check` reports no changes; this declares a serializer field, not a model field

- [x] Svelte 5 syntax; `pnpm run lint` and `pnpm run format` run — the frontend change is declarative config, no component. eslint reports 8 pre-existing errors in `crud.ts`, **0 on the lines added here**; the new test file carries the one `logedPage is defined but never used` that every test in the suite has (it is a fixture used for its login side effect)
- [x] New/changed UI strings added to `frontend/messages/en.json` — none needed, the buttons reuse the labels already used by the Evidences and Tasks tabs

- [x] Tests added or updated and passing locally — backend `uv run pytest`, frontend `pnpm run test`, and the new e2e test
- [x] Regression test added for bug fixes — both suites were red-checked; the backend pair specifically covers the silent-drop failure mode
- [ ] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added — checked `concepts/applied-controls.md` and the guides: neither describes the detail-view tabs nor the linking workflow, so nothing there goes stale
- [ ] No tests or docs needed for this change

issue-4696-3-findings-remplis
issue-4696-after
issue-4696-documents-after

## New Features
- Applied controls can link existing documents, findings, assets, and incidents.
- Related-object tabs now offer options to select existing items or add new ones.
- Assets can be linked to documents.

## Bug Fixes
- Evidence review status now reflects revisions filed on the relevant occurrence.
- Folder parent updates now enforce supported hierarchy and permission rules.

## Tests
- Added coverage for linking documents and assets to applied controls.
- Added checks for related-object actions and successful attachments.


nas-tabchiche (changes requested):
The Documents `select-existing` modal unlinks every document already attached to the applied control when you use it to add another one. That is silent data loss on a feature the PR is explicitly shipping, so we can't merge as-is.

The fix is a one-line addition to `AppliedControlSchema` in `schemas.ts`, next to findings and incidents:

```typescript
control_documents: z.string().uuid().optional().array().optional(),
```

PR #4834, opened 2026-09-16
docs(adr): route outbound mail through one service over MAILERS

nas-tabchiche:
## What & why

ADR for making the outbound mail transport pluggable. Status: Proposed.

Microsoft 365 and Google Workspace are retiring password auth for SMTP, and Django 6.1 deprecated the `EMAIL_*` settings in favour of `MAILERS`. The ADR settles how we handle both: one mail service in `core` that tries the mailers in `MAILERS` in order, SMTP stays the default with existing env vars untouched, the `_RESCUE` variables become a second SMTP mailer instead of a special case. New transports: OAuth over SMTP (XOAUTH2) for Microsoft 365 and Google Workspace in the first build, then a Graph backend we write ourselves. Provider HTTP APIs (anymail) are rejected: SMTP with an API key already reaches every provider. Webhooks, inbound and the queue are out of scope.

Includes a transition section for running on-prem instances: no config change on upgrade, no migrations, and what Django 7.0 removes.

Side finding: Django 6.1's SMTP backend no longer accepts a custom `ssl_context`, so `EMAIL_FORCE_TLS_1_2` is a silent no-op on main today. Fix tracked separately.

No code changes. Implementation follows once accepted.

## Checklist

- [x] No tests or docs needed for this change — why: the change is the document

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record documenting the centralized mail service and SMTP-based transport strategy.
  * Documented primary and optional backup mailer configuration, failover behavior, delivery outcomes, and disabled-mail handling.
  * Recorded operational considerations, security requirements, error reporting, and potential future transport improvements.


ab-smith:
We can probably retire the rescue mode since we’ll have MAILERS list to enumerate when needed

eric-intuitem:
It's perhaps the good moment to drop the rescue mailer, which is legacy and mostly useless.

PR #4844, opened 2026-09-17
Source ip and rfc 7239

melinoix:
## What & why

Failed-login entries in the Enterprise audit log ("journal")
only show a timestamp and the user's email, with no source IP , needed for incident
investigation. Since the customer sits behind a WAF, the raw connection IP isn't the
real client IP; it needs to be read from `X-Forwarded-For` (and ideally the
standardized RFC 7239 `Forwarded` header).

This PR:
- Adds `get_client_ip()` in `enterprise_core/signals.py`, used by the
  `user_login_failed` handler. Prefers the RFC 7239 `Forwarded: for=...` header,
  falls back to `X-Forwarded-For` (first hop), then to `REMOTE_ADDR`.
- Surfaces the `LogEntry.remote_addr` field (already captured, already exposed by
  the API, but never displayed) in the audit-log list and detail screens.
- Adds the `remoteAddr` translation key (EN: "Source IP", FR: "IP source").

## Test plan

- Verified `get_client_ip()` against a real Django `HttpRequest` (via
  `RequestFactory`, loaded through the actual `enterprise_core.settings` module):
  `X-Forwarded-For` with multiple hops, RFC 7239 `Forwarded` with a quoted
  IPv6+port value, no proxy headers (falls back to `REMOTE_ADDR`), and
  `request=None`.
- `ruff format --check` / `ruff check` pass on `signals.py`.
- `pnpm run format` / `pnpm run lint` clean on the changed frontend files (repo has
  pre-existing, unrelated lint debt elsewhere).
- visual check on instance

## Checklist

- [x] CLA accepted — n/a (internal contributor)

### Backend — `backend/` / `enterprise/backend/`

- [x] No migration needed (`remote_addr` already existed on `LogEntry`)
- [x] No new dependencies

### Frontend — `frontend/` / `enterprise/frontend/`

- [x] New UI strings added to `frontend/messages/en.json` (and `fr.json`)
- [x] Clicked through in dev server — pending, will attach screenshot
image

## Summary by CodeRabbit

- **New Features**
  - Audit log entries now display the source IP address in list and detail views.
  - Source IP detection supports forwarded proxy headers, including IPv4 and IPv6 addresses, with port information handled correctly and invalid values skipped.

- **Bug Fixes**
  - Failed-login records now capture the client IP when requests pass through proxies.

- **Localization**
  - Added English and French labels for the source IP field.


melinoix:
@coderabbitai review

enterprise/backend/enterprise_core/signals.py
```diff
@@ -11,12 +14,53 @@
...
+
+def _as_valid_ip(address):
+    try:
+        ipaddress.ip_address(address)
+    except ValueError:
+        return None
+    return address
+
+
+def get_client_ip(request):
```
tchoumi313, line 33: Could we reuse allauth’s configured client-IP resolution through `get_adapter().get_client_ip(request)` instead of implementing another parser here?
Allauth already uses this adapter method for its rate limiting and user sessions, and it supports trusted proxy counts/client-IP headers while validating. If we needs custom BFF behavior, we can override AccountAdapter.get_client_ip() in `iam/adapter.py`, keeping audit logging and future allauth throttling on the same resolution policy.

https://github.com/pennersr/django-allauth/blob/b929d5c8fb9abfc9e410546713fde16f1cec851e/allauth/core/internal/httpkit.py#L198

enterprise/backend/enterprise_core/signals.py
```diff
@@ -11,12 +14,53 @@
...
+    if not request:
+        return None
+
+    xff = request.headers.get("X-Forwarded-For")
```
tchoumi313, line 37: The normal login flow goes through the SvelteKit server action, which creates a new backend request without forwarding any client-address information. Therefore, this header will normally be absent and the fallback will record the frontend container/pod address.
Additionally, selecting the first X-Forwarded-For element trusts a potentially client-controlled value. Please establish explicit application-level forwarding from the Svelte login request to Django, similar to the webhook path (line 61 frontend/src/routes/api/workflows/hooks/[workflow_id=uuid]/[node_ref]/[secret]/+server.ts), and ensure the frontend and backend agree on which value is trusted. The raw incoming chain should not simply be copied.
Note that the webhook appends an address and its backend throttle selects the trailing element, while this implementation selects the leading element, so the webhook logic cannot be copied without reconciling that difference.

enterprise/backend/enterprise_core/signals.py
```diff
@@ -11,12 +14,53 @@
...
+                _strip_port((match.group(1) or match.group(2)).strip('"'))
+            )
+            if candidate:
+                return candidate
+
+    return request.META.get("REMOTE_ADDR")
+
+
 @receiver(user_login_failed)
 def log_login_failed(sender, credentials, request, **kwargs):
```
tchoumi313, line 59: the IP never reaches this signal. On the allauth login path, authenticate() is called with request=None: `iam/adapter.py:44` builds AuthTokenSerializer(data={...}) without context={"request": request}, so DRF passes `request=None` into django.contrib.auth.authenticate(), and user_login_failed fires with request=None. get_client_ip(None) returns None, so remote_addr is always NULL here regardless of the header handling below. Fix the adapter to forward the request context, otherwise the whole feature is a no-op on this path. A test should assert the signal receives a non-None request.

in iam/adapter.py, the authenticatte may looks like 
```suggestion
serializer = AuthTokenSerializer(
      data={
          "username": credentials.get("email"),
          "password": credentials.get("password"),
      },
      context={"request": request},   # so authenticate() -> user_login_failed carries the request
  )
```

tchoumi313 (changes requested):
Please add coverage proving that the client address received by the Svelte login action is propagated through the allauth request and stored in LogEntry.remote_addr. The test (if possible) should also verify that a forged leading X-Forwarded-For value cannot control the stored address.

monsieurswag:
Wouldn't this give the ability for any attacker spoof their IP address for instances where there's no WAF ?

For example doing this:

```py
curl -X POST 'http://127.0.0.1:8000/api/_allauth/app/v1/auth/login' -d '{"email":"aaa@gmail.com","password":"badpwd"}' -H 'Content-Type: application/json' -H 'X-Forwarded-For: 1.1.1.1'
{"status": 400, "errors": [{"message": "The email address and/or password you specified are not correct.", "code": "email_password_mismatch", "param": "password"}]}
```

Generate this `LogEntry`:

```py
>>> L.objects.order_by("-timestamp").first().remote_addr
'1.1.1.1'
```

We can't let instances be vulnerable to IP spoofing by default.

A solution could be to create a new global setting (which would default to `False`), and if this setting is set to `True` then we use the RFC 7239 headers for IP resolution.
So it would be the responsability of the instance admin to set this setting to `True` if he considers it appropriate for his instrastructure.

ab-smith:
will be covered on the rate limiting PR, cc @tchoumi313

PR #4874, opened 2026-09-21
fix(mappings): build the engine per request instead of caching it per worker

### LINKED ISSUE #4791: KeyError on 'requirement_assessments' when applying mapping with custom-packaged source framework
**Describe the bug**
Creating a compliance assessment via "Apply a mapping" fails with a 500 Internal Server Error / KeyError, even though the mapping engine correctly computes the projected results.

**To Reproduce**
1. Load a custom framework with packager "local" (imported via direct Excel upload)
2. Load the official ISO/IEC 27001:2022 framework (packager "intuitem")
3. Generate a custom RequirementMappingSet between the two frameworks using tools/excel/cis/prep_mapping_cis_controls_iso_27.py, passing the "local" packager
4. Load this mapping library successfully
5. On an existing compliance assessment using the "local" framework, click "Apply a mapping" and select the ISO 27001 target framework
6. Click Save

**Expected behavior**
A new compliance assessment should be created on the target framework, pre-populated with results/applied controls based on the mapping.

**Actual behavior**
Request POST /api/compliance-assessments/ returns HTTP 500.

Traceback:
File "/code/core/views.py", line 11953, in perform_create
    requirement__urn__in=best_results["requirement_assessments"],
KeyError: 'requirement_assessments'

Local variables at the point of failure show that `audit_from_results` was correctly computed (containing `min_score`, `max_score`, and a fully populated `requirement_assessments` dict with 56 entries, e.g. `urn:local:risk:req_node:cis-controls-v8:1.1`), but `best_results` is an empty dict `{}` when accessed a few lines later.

- source_urn = 'urn:local:risk:framework:cis-controls-v8'
- dest_urn = 'urn:intuitem:risk:framework:iso27001-2022'

Additionally, when a compliance assessment is created manually (not via "Apply a mapping") on the same perimeter/target framework, no inline "mapping suggestion" indicator appears on any requirement, even though the "Mapping suggestion is available for this requirement" banner is shown at the top of the requirements list.

**Environment**
- CISO Assistant Version: v3.21.2
- Build: 97e68c700
- Infrastructure: S-FS
- Deployment: Docker Compose, self-hosted
- Browser: Chrome
- Source framework packager: local (direct Excel import)
- Target framework packager: intuitem (official library)
- Mapping generated via: tools/excel/cis/prep_mapping_cis_controls_iso_27.py, packager arg = "local"

Claquetteuuuh:
## What & why

The mapping engine was a module-level instance, so each gunicorn worker held its own copy of the graph, refreshed by signal only in the worker that handled the write. The other workers kept answering from a graph that predated a library loaded elsewhere. When a request for "Apply a mapping" landed on one of them, `best_mapping_inferences` found no path and returned `{}`, which `perform_create` indexed without a guard.

Reproduced with 2 workers: after loading a mapping library through the UI, the projection endpoint returned the new target **4 times out of 12**. After the change, **12 out of 12**.

Every caller now builds its own engine. `__init__` reads the database, and `_ensure_loaded`, `reload_cache` and `core/mappings/signals.py` all go away since nothing survives the request. The change removes more code than it adds.

The missing guard on `best_results` is added as well: an empty result is a legitimate outcome when the two frameworks are unrelated, and the audit should simply be created without being pre-filled.

Measured cost: around 200 ms per engine-backed action. All five call sites sit behind a deliberate user action (advanced analytics, creating an audit from a baseline, the mapping preview, the framework insights, the `has_mapping_path_to` filter) and none is on a hot path.

Closes #4791

## Test plan

Backend, on this branch rebased on current `main`:

- `uv run pytest` — 5227 passed, 20 skipped, 1 xfailed, 0 failed
- `./tests/e2e-tests.sh --no-snapshot detailed/mappings.test.ts` — 2 passed, including `user can map iso27001-2022 audit to a new csf-1.1 audit`
- `pre-commit run --files` — passed, and no new ruff finding on any touched file

New tests in `backend/app_tests/api/test_mapping_engine_no_shared_cache.py`, each checked red before the fix:

- `test_the_module_exposes_no_shared_engine` — guards against reintroducing the module-level instance
- `test_a_new_engine_sees_a_library_stored_without_its_signal` — a write with no `post_save`, which is what another worker's write looks like from here
- `test_creating_an_audit_from_a_baseline_with_no_mapping_path_does_not_fail` — without the guard it reproduces the exact traceback of #4791
- `test_creating_an_audit_from_a_baseline_applies_the_mapping` — the happy path; forcing the inferences to be empty makes it fail on `assert 'not_assessed' == COMPLIANT`

Manual, with gunicorn at 2 workers behind the dev frontend: loaded a mapping library from the Libraries screen, then reloaded the advanced analytics page of an audit. Before the change the new target appeared on one refresh and vanished on the next; after the change it is there every time.

Three `engine.reload_cache()` calls were removed from `test_api_compliance_assessments.py`: they existed only to stop the shared cache leaking between tests, and there is no shared cache any more.

## Checklist

- [x] Migrations generated with `makemigrations` and committed (or none needed) — none needed, no model change
- [x] New dependencies checked for known vulnerabilities and called out above — no new dependency

- [x] Tests added or updated and passing locally — backend `uv run pytest`, frontend `pnpm run test` / `./tests/e2e-tests.sh` _(if relevant)_ — `pnpm run test` not run, no frontend file is touched
- [ ] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added _(if relevant)_ — internal mechanics only, nothing changes in how the product is understood or used
- [ ] No tests or docs needed for this change — why: tests were added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compliance assessment and audit creation when no mapping exists between frameworks; audits can now be created without a mapping path.
  * Ensured newly created mapping engines detect current libraries and apply valid requirement mappings consistently.
  * Prevented stale mapping data from being reused across requests or command executions.
* **Tests**
  * Added regression coverage for mapping isolation, library detection, unmapped frameworks, and successful mapped assessments.


monsieurswag (approved):
This PR works as expected.

There a very obvious performance loss (creating a `MappingEngine` instance takes 320ms).
Fixing them would require to slightly redesign the current `MappingEngine` so it's outside the scope of this PR.

If this PR is merged please create a Jira issue for this and assign it to me.

PR #4877, opened 2026-09-21
feat(security): add login brute-force protection: rate limiting, ip forwarding, audit hardening

tchoumi313:
## What & why

Adds brute-force / credential-attack protection for local login.

  - Rate-limit local logins per email+IP: 5 failed attempts allowed, 6th blocked, 5-minute window (ACCOUNT_RATE_LIMITS, keyed via _get_login_attempts_cache_key).
  - Forward and pin the real client IP end to end (X-Real-IP from SvelteKit hooks.server.ts through to Django's resolve_client_ip), so rate limiting and audit logs use the actual client, not the proxy.
  - Audit failed login attempts as LogEntry (enterprise LOGIN_FAILED action) plus a structlog login_throttled signal for SIEM.
  - Harden password reset: revoke outstanding tokens on use, send email async via huey, alert on suspicious reset activity.
  - Partition prune_auditlog into independent count quotas for regular vs. security LogEntry actions, so a failed-login flood can't evict business audit history.
  - Fix non-field login error handling in the login form (setError called without a field param for account-level errors like "too many attempts").

Closes #

## Checklist

- [ ] Regression test added for bug fixes _(if relevant)_
- [ ] `product-docs/` updated, new pages listed in `SUMMARY.md`, vocabulary terms added _(if relevant)_

## Summary by CodeRabbit

- **Security**
  - Failed login attempts are rate-limited per client IP, with clearer feedback when limits are reached.
  - Failed login activity is recorded for security monitoring while preserving business audit history.
  - Password resets now revoke the user’s active sessions and tokens.

- **Bug Fixes**
  - Password reset responses no longer reveal whether an email address exists.
  - Password reset emails are processed asynchronously for improved responsiveness.
  - Login errors now display correctly at the field or form level.


backend/iam/adapter.py
```diff
@@ -25,6 +25,12 @@
 DEFAULT_ATTRIBUTE_MAPPING_GROUPS = ["groups"]
 
 
+def resolve_client_ip(request):
+    if request is None:
+        return None
+    return request.headers.get("X-Real-IP") or request.META.get("REMOTE_ADDR")
```
monsieurswag: I am kind uneasy about the idea of making our API IP-spoofable.

With this an attacker can spoof its IP and bypass bruteforce protection for any API which can be requested without passing by a reverse proxy overwriting the `X-Real-IP` header.

This is tricky as we have community-edition users + on-premise clients, and all of these people can have different infrastructures around ciso-assistant.

The following function would be better for example:
```py
import ipaddress

def resolve_client_ip(request):
    if request is None:
        return None

    request_ip = request.META.get("REMOTE_ADDR")
    header_ip = request.headers.get("X-Real-IP")

    if not request_ip:
        return header_ip

    if not header_ip:
        return request_ip

    request_ip_info = ipaddress.ip_address(request_ip)

    if request_ip_info.is_private or request_ip_info.is_loopback:
        return header_ip
    else:
        return request_ip
```

With this solution still have 2 problems though:

1. An attacker could still spoof the API if he is inside the network (has a private IP address).
2. If the frontend/the reverse proxy (in front of the API) doesn't have a private/loopback IP address from the backend's point of view, the resolved IP would always be the same.

The ideal solution would be to only trust `X-Client-IP` from a list of a trusted IPs.
We already have the exact the same kind of logic in our infra configuration settings (see `InfraConfigSerializer.allowed_ips`).

So we should do the same thing for this.

monsieurswag (changes requested):

backend/iam/adapter.py
tchoumi313: Thanks

backend/iam/adapter.py
```diff
@@ -41,22 +47,28 @@ def is_open_for_signup(self, request):
...
+    def get_client_ip(self, request):
+        return resolve_client_ip(request) or super().get_client_ip(request)
+
+    def _get_login_attempts_cache_key(self, request, **credentials):
+        base = super()._get_login_attempts_cache_key(request, **credentials)
+        if request is None:
+            return base
+        from allauth.core.internal.ratelimit import get_ip
+
+        return f"{base}:{get_ip(request)}"
```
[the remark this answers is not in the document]
tchoumi313: no thanks

tchoumi313:
- Added a management command `setup_cache_table` to create necessary tables for database-backed caching.
- Updated `startup.sh` and various deployment scripts to call `setup_cache_table` before migrations.
- Modified `ChangePasswordView` to revoke tokens while excluding the current token.
- Introduced `PasswordResetRateThrottle` to limit password reset requests per IP.
- Enhanced tests for token revocation and login throttling to ensure correct behavior.
- Updated documentation to reflect changes in cache setup and usage.

